PR #3, opened 2020-08-12
Comply with JWT

shizhMSFT:
In this PR, the signature is a JWT variant. Therefore, the signature file is no longer a JSON file.

Additionally, media type is required for the manifest and is verified.

Related: https://github.com/shizhMSFT/nv2/pull/16


trishankatdatadog:
@shizhMSFT Sorry, I don't understand: how is JWT not JSON? Also, the PR seems to ignore canonicalization altogether for now.

Cc @SantiagoTorress @JustinCappos @mnm678

shizhMSFT:
> @shizhMSFT Sorry, I don't understand: how is JWT not JSON? Also, the PR seems to ignore canonicalization altogether for now.
> 
> Cc @SantiagoTorress @JustinCappos @mnm678

A JWT is a concatenation of base64 URL encoded JSON objects. The statement `JWT is not JSON` is elaborated as `A JWT object is not a JSON object but a string of encoded JSON objects`.

JSON cannoicalization is not considerred in this PR, and will be included in the next iteration.

PR #8, opened 2020-08-26
Specification and Prototype - JWT based signature

shizhMSFT:
We propose specification for Notary v2, including the client and server side.

We also provide a client prototype as a refernece implementation of the specification.

Basically, the client `nv2` is able to sign and verify a manifest-based artifact offline. Later, the result signature can be shared to others via OCI-compliant registries as OCI artifacts or as a normal file on the file system.

Related: #2


SteveLasker:
Based on our [Notary v2 call this week](https://hackmd.io/_vrqBGAOSUC_VWvFzWruZw?view), we agreed to merge this into a `prototype-1` branch.
This will give us a baseline to which we can open issues and submit new PRs.

PR #10, opened 2020-08-26
Distribution options for persistence & discovery

SteveLasker:
Includes:

Persistence options
Reference linking options
Signature discovery options
Paging options
Signed-off-by: Steve Lasker stevenlasker@hotmail.com

This is the continuation of #5


docs/distribution/README.md
```diff
@@ -0,0 +1,923 @@
+# OCI Distribution - Notary v2 Signature Support
+
+To support [Notary v2 goals][notaryv2-goals], upload, persistance and discovery of signatures must be supported.
+
+To minimize the complexity of registry operators and projects to adopt Notary v2, a balance between leveraging what already exists and new patterns to support secure discovery are explored.
+
+## Table of Contents
+
+* [Signature Persistance](#signature-persistance)
```
rchincha, line 9: typo - Persistence

docs/distribution/README.md
```diff
@@ -0,0 +1,923 @@
...
+* [Signature Discovery](#signature-discovery)
+* [Signature Pull](#signature-pull)
+* [Example Artifacts](#example-artifacts)
+
+## Signature Persistance
+
+Several options for how to persist a signature were explored. We measure these options against the [goals of Notary v2][notaryv2-goals], specifically:
+
+* Maintain the original artifact digest and collection of associated tags, supporting existing dev through deployment workflows
+* Multiple signatures per artifact, enabling the originating vendor signature, public registry certification and user/environment signatures
```
rchincha, line 20: So potentially allowing for recursive signatures = Sign(Payload, Annotations, Sign(Payload, Annotationss, Sign(...))) as the artifact gets moved around?

docs/distribution/README.md
```diff
@@ -0,0 +1,923 @@
...
+* Native persistance within an OCI Artifact enabled, distribution*spec based registry
+* Artifact and signature copying within and across OCI Artifact enabled, distribution spec based registries
```
rchincha, line 22: Now that dist-spec is brought into this discussion, is dist-spec a NECESSARY condition? or for example, can docker registry be notary-v2 compliant without being dist-spec compliant?

docs/distribution/README.md
```diff
@@ -0,0 +1,923 @@
...
+* Support private registries, where public content may be copied to, and new content originated within
+* Air-gapped environments, where the originating registry of content is not accessible
+
+To support the above requirements, signatures are stored as separate [OCI Artifacts][oci-artifacts]. They are maintained as any other artifact in a registry, supporting standard operations such as listing, deleting, garbage collection and any other content addressable operations within a registry.
+
+Following the [OCI Artifacts][oci-artifacts] design, signatures are identified with: `config.mediaType: "application/vnd.cncf.notary.config.v2+jwt"`.
+The config object contains the signature and signed content. See [nv2-signature-spec][nv2-signature-spec] for details.
+
+Storing a signature as a separate artifact enables the above goals, most importantly the ability to maintain the existing tag and and digest for a given artifact.
+
```
rchincha, line 33: In terms of actual deployment, do we want to explicitly call out a key mgmt or vault interactions? Definition of air-gapped may also mean even a vault may not be trusted, etc. This part may or may not be outside scope of this document.

docs/distribution/README.md
SteveLasker: There was a recent conversation about signing signatures in the 3T SBOM working group as well. It was thought to be more complicated and not clear if it's needed. The current model we're focusing on is independent, parallel signatures. In our core example, acme-rockets may decide to only add their signature if the wabbit-networks signature first exists, and is valid. But the acme-rockets signature would be persisted independently of the wabbit-networks signature. This would mean the production environment only needs to validate the existence of the acme-rockets signature. 
They could also say they must have both the acme-rockets signature and wabbit-networks signature as part of their policy management. But, they're not coupled, today.

Can you think of a scenario where we need chained signatures?

docs/distribution/README.md
SteveLasker: We don't believe we can achieve the e2e goals of Notary v2 without them. If you consider Notary v1/Docker Content Trust required a significant additional service, which had a host of additional issues, we believe this is a reasonable bar.
For instance, our design goals require linking a signature with an artifact without changing the original artifacts tag or digest. This requires another artifact to reference it. We looked into one manifest referring to another with content in the config object. But, that would require registries to implement ref counting, by parsing the `manifest.config` to support garbage collection.
We're now focusing on using Index as it has a linking mechanism. In this case, we need a way to differentiate multi-arch container inedexes, from Notary v2 and CNAB based indexes.
Lastly, we need an api to find the signatures, based on a target artifact.
All of these require changes. The index changes focus on leveraging the hardest parts of registry implementations (ref counting & garbage collection) while adding capabilities that have broader needs.
We'll also need changes to existing clients to pull the workflows together.
We hope the value will be enough, that the relatively small community of people that own the impacted products would be motivated to add the relatively minimal amount of work to enable the e2e workflows. But, that will ultimately be the test of the effort - adoption.

Was there a specific reason you were concerned about changes?

docs/distribution/README.md
SteveLasker: Yes, definitely, but we're looking to the key management working group to make the first round of recommendations for what and how.

docs/distribution/README.md
```diff
@@ -0,0 +1,923 @@
...
+    }
+    ```
+
+**Pros with this approach:**
+
+* Utilize the existing `index.manifests` collection for linking artifacts
+* Registries that support oci index already have infrastructure for tracking `index.manifests`, including delete operations and garbage collection
+* Existing distribution-spec upload APIs are utilized
+* Unlike the manifest proposal, no additional artifact handler would be required to parse the config object for linking artifacts
+* Based on the artifact type:  `manifest.config.mediaType: "application/vnd.cncf.notary.config.v2+jwt"`, role check may be done to confirm the identity has a signer role
```
reasonerjt, line 314: Steve, let me clarify:
Is this signer role optionally required based on the implementation of the registry service, or part of any spec?

docs/distribution/README.md
```diff
@@ -0,0 +1,923 @@
...
+    ```
+
+**Pros with this approach:**
+
+* OCI Artifacts already supports manifest based artifacts, through the `manifest.config.mediaType`
+
+**Cons with this approach:**
+
+* Manifests have no means to reference other artifacts.
+* An alternative is required to link a target artifact with it's signature. Either through parsing the signature `manifest.config` object, or [a separate API for linking objects](#linking-signatures-to-artifacts).
```
reasonerjt, line 170: If a annotation is added to the manifest of signature, it's not required to parse the `manifest.config` object.  But there's a possibility of inconsistency.  i.e. the annotation says it's a signature for "artifact A" but the "reference" in the signature object says it's a signature for "artifact B"

docs/distribution/README.md
```diff
@@ -0,0 +1,923 @@
...
+* Existing distribution-spec upload APIs are utilized
+* Unlike the manifest proposal, no additional artifact handler would be required to parse the config object for linking artifacts
+* Based on the artifact type:  `manifest.config.mediaType: "application/vnd.cncf.notary.config.v2+jwt"`, role check may be done to confirm the identity has a signer role
+* As registry operators may offer role checking for different artifact types, Notary v2 Signatures are just one of many types they may want to authorize
+
+**Cons with this approach:**
+
+* OCI index does not yet support the [OCI config descriptor][oci-descriptor]. This would require a schema change to oci-index, with a version bump.
+  * This has been a [desired item for OCI Artifacts][oci-artifacts-index] to support other artifact types which would base on Index.
+* An additional role check is performed, based on the artifact type. Also noted as a pro as registry operators may want to utilize this for other artifact types, making it a consistent model.
```
reasonerjt, line 321: Is this "role check" required?

docs/distribution/README.md
```diff
@@ -0,0 +1,923 @@
...
+**Pros with this approach:**
+
+* Utilize the existing `index.manifests` collection for linking artifacts
+* Registries that support oci index already have infrastructure for tracking `index.manifests`, including delete operations and garbage collection
+* Existing distribution-spec upload APIs are utilized
+* Unlike the manifest proposal, no additional artifact handler would be required to parse the config object for linking artifacts
+* Based on the artifact type:  `manifest.config.mediaType: "application/vnd.cncf.notary.config.v2+jwt"`, role check may be done to confirm the identity has a signer role
+* As registry operators may offer role checking for different artifact types, Notary v2 Signatures are just one of many types they may want to authorize
+
+**Cons with this approach:**
```
reasonerjt, line 317: Another con is inconsistency, similar to the case mentioned comment:
https://github.com/notaryproject/nv2/pull/10/files#r481223090

docs/distribution/README.md
```diff
@@ -0,0 +1,923 @@
...
+
+* An additional indirection between the original artifact being signed and the individual signatures.
+* An `index.config.mediaType` is still required to identify the type of index being something other than a multi-arch index.
+
+The implied benefit is the signature is moved from the `index.config` to a layer within a manifest. However, config objects are [oci descriptors][oci-descriptor] pointing to blobs. Whether the signature is stored within a config or a layer is little difference. Since we'll need an `index.config.mediaType` to differentiate a signature index from a multi-arch index, are we justifying additional round trips to get a list of signatures?
+
+## Linking Signatures to Artifacts
+
+A signature is only interesting if it's linked to the object it's signing. The question is how.
```
reasonerjt, line 524: I personally prefer not to add new APIs to distribution-spec for this purpose.

docs/distribution/README.md
```diff
@@ -0,0 +1,923 @@
...
+* Existing Index parsing logic is used to track despondencies
+* No additional distribution-spec APIs are required
+
+**Cons with this approach:**
+
+* OCI Index would need to support a config object:  [Add Index Support for Artifact Type #806](https://github.com/opencontainers/image-spec/issues/)
+
+## Signature Discovery
+
+Once a signature artifact is in a registry and linked to its target artifact, how is it retrieved?
```
reasonerjt, line 618: With the current proposed structure of the signature object, it seems Hard to track the digest of signature object in distribution via the digest of the manifest of the artifact that is signed.

docs/distribution/README.md
```diff
@@ -0,0 +1,923 @@
...
+
+The response will be in the following format:
+
+```HTTP
+200 OK
+Content-Type: application/json
+{
+  "digest": "sha256:90659bf80b44ce6be8234e6ff90a1ac34acbeb826903b02cfa0da11c82cbc042",
+  "references": [
+    {
```
reasonerjt, line 676: If purpose of the reference API is generic, I don't see why `config-mediaType` is needed in the response.

If `config-mediaType` is removed this generic API is not very helpful for signature discovery.

docs/distribution/README.md
```diff
@@ -0,0 +1,923 @@
...
+GET /v2/<name>/manifests/sha256:90659bf80b44ce6be8234e6ff90a1ac34acbeb826903b02cfa0da11c82cbc042/signatures/
+```
+
+The response will be in the following format:
+
+```HTTP
+200 OK
+Content-Type: application/json
+{
+  "digest": "sha256:90659bf80b44ce6be8234e6ff90a1ac34acbeb826903b02cfa0da11c82cbc042",
```
reasonerjt, line 640: the field `digest` is redundant?

docs/distribution/README.md
```diff
@@ -0,0 +1,923 @@
...
+  "signatures": [
+    {
+      "digest": "sha256:90659bf80b44ce6be8234e6ff90a1ac34acbeb826903b02cfa0da11c82cbc042",
+      "mediaType": "application/vnd.oci.image.index.v1+json",
+      "size": "1024",
+      "signature-typ": "x509"
```
reasonerjt, line 646: Is it necessary to include `signature-typ` in the response?

If it's removed, the item in `signatures` list is more consistent with definition in:

https://github.com/opencontainers/image-spec/blob/master/descriptor.md

docs/distribution/README.md
SteveLasker: That's a great question. The distribution spec doesn't call out roles at all.  Yet, all registries support a set of roles. 
What roles should a signature capability call out as a requirement in the spec, vs. a capability we know registry operators will implement and how do we acknowledge that need without putting it in the spec?

With the recent feedback that registries may want to assign roles to different artifacts, I'm inclined to think of this as nothing specific to signatures and not actually call out a signing role as a spec requirement.

docs/distribution/README.md
SteveLasker: We've generally avoided using annotations for anything other than interesting metadata you can get when looking at an object. 
But, to your point, we could use annotations, but similar to parsing the config object, it's still new logic registry operators would have to implement. The focus on Index allows registries to leverage the same linking infrastructure they use for multi-arch manifests, and would use for other artifact types like CNAB.

docs/distribution/README.md
SteveLasker: This follows the same logic as: https://github.com/notaryproject/nv2/pull/10#discussion_r481463463
We know registries will implement role-checks, so we just need a design that supports it.

docs/distribution/README.md
SteveLasker: To support Notary v1 server, we need to stand up an entire new service. With all the downsides of inconsistent data across the two services (like a tag is different with/without docker_content_trust enabled)
So, while we'd all like to get Notary v2 for free, we don't see a means without:
- hacking something into the existing APIs
- sacrificing the requirements

The latest approaches we're pursuing is adding new APIs that would work across any artifact type, including signatures. 
There is a get-signatures API in the current thinking, but I'm hoping this will become some generic "find-my-dependencies" api in a subsequent iteration.

Can you think of a proposal that meets the requirements without adding new APIs?

docs/distribution/README.md
SteveLasker: There's an ordering of artifacts issue we're trying to address:
1. wabbit-networks can push their image to docker hub
1. after it arrives, docker hub might certify it with an a docker.io signature
1. after the image is copied to acme-rockets, and tested, a new signature is added

So, the registry can and should store this linking information. Just as we do for manifests-->layers and index-->manifests

What we're doing here is making sure we can also track the reverse. If a docker image is persisted as a manifest, and the signature is persisted as an index, we need a reverse lookup:
manifest-->index

I think the question is somewhat rooted in the lack of an indexing solution in docker/distribution or the spec. What we're saying here is this will be needed. If registries choose to persist the lookup information as files, or an instance of a db is a registry operator question.
We will need to decide how we implement this in the reference implementation of docker/distribution.

docs/distribution/README.md
SteveLasker: Yup, agreed. This is definitely a work in progress. 
If folks continue to agree with storing signatures as index.config objects, and add config to index, then this is the next place we'll be focusing: what should this api look like. And, can it be generic? If it's generic, and only provides the artifact type, how do we know which of the 4 signatures we actually want? Or, do we need to pull all 4 (n) and decide after they're pulled. 
me: 👎

docs/distribution/README.md
SteveLasker: The example shows the same digest is a copy/paste mistake. The idea is the response shows a confirmation of the digest for what was requested, with a collection of things that were found.
I've fixed the example, and added the full descriptor info.
Or, are you saying we shouldn't return a confirmation of the digest?

docs/distribution/README.md
SteveLasker: Yeah, this is the problem noted above at: https://github.com/notaryproject/nv2/pull/10#discussion_r481469759
we need more focus here on whether we'll have a generic api, or a signature specific api.

docs/distribution/README.md
```diff
@@ -0,0 +1,927 @@
...
+* [Signature Pull](#signature-pull)
+* [Example Artifacts](#example-artifacts)
+
+## Signature Persistence
+
+Several options for how to persist a signature were explored. We measure these options against the [goals of Notary v2][notaryv2-goals], specifically:
+
+* Maintain the original artifact digest and collection of associated tags, supporting existing dev through deployment workflows
+* Multiple signatures per artifact, enabling the originating vendor signature, public registry certification and user/environment signatures
+* Native Persistence within an OCI Artifact enabled, distribution*spec based registry
```
samuelkarp, line 21: You have a `*` again.

docs/distribution/README.md
```diff
@@ -0,0 +1,927 @@
...
+          "platform": {
+            "architecture": "ppc64le",
+            "os": "linux"
+          }
+        }
+      ]
+    }
+    ```
+
+    The `manifest.config` contains the following signature information:
```
samuelkarp, line 231: nit: I think you mean `index.config` here.

docs/distribution/README.md
```diff
@@ -0,0 +1,927 @@
...
+
+    ```json
+    {
+      "schemaVersion": 2,
+      "mediaType": "application/vnd.oci.image.index.v2+json",
+      "config": {
+        "mediaType": "application/vnd.cncf.notary.index.config.v2+jwt",
+        "digest": "sha256:222cb130c152895905abe66279dd9feaa68091ba55619f5b900f2ebed38b222c",
+        "size": 1906
+      },
```
samuelkarp, line 377: What's this object in the example here?

docs/distribution/README.md
```diff
@@ -0,0 +1,927 @@
...
+        {
+          "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
+          "digest": "sha256:ec4b8955958665577945c89419d1af06b5f7636b4ac3da7f12184802ad867736",
+          "size": 73109
+        }
+      ]
+    }
+    ```
+
+2. **index digest for the wabbit-networks signature** `sha256:222ibbf80b44ce6be8234e6ff90a1ac34acbeb826903b02cfa0da11c82cb222i`
```
samuelkarp, line 206: Another approach, very similar to this but without taking a dependency on adding a new index `config` descriptor, would be to add the signature descriptor in the `manifests` array instead.  You could have an index that looked like this:

```json
{
    "schemaVersion": 2,
    "mediaType": "application/vnd.oci.image.index.v2+json",
    "manifests": [
        {
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "digest": "sha256:111ma2d22ae5ef400769fa51c84717264cd1520ac8d93dc071374c1be49a111m",
            "size": 7023,
            "platform": {
                "architecture": "ppc64le",
                "os": "linux"
            }
        },
        {
            "mediaType": "application/vnd.cncf.notary.config.v2+jwt",
            "digest": "sha256:222cb130c152895905abe66279dd9feaa68091ba55619f5b900f2ebed38b222c",
            "size": 1906
        }
    ]
}
```

The client can still determine that it's looking for a signature by way of iterating through the `manifests` and finding one with the correct `mediaType`.  (This is a bit of an abuse of the name "manifests", but so is the way you're using "config" so it's not terribly different.)  This approach also gives you the opportunity to add multiple signatures in the same index.

docs/distribution/README.md
```diff
@@ -0,0 +1,927 @@
...
+* define an `int32` field `page_size` in the `List` method's request message. Clients use this field to specify the maximum number of results to be returned by the server. The server **may** further constrain the maximum number of results returned in a single page. If the `page_size` is `0`, the server will decide the number of results to be returned.
+* define a `string` field `next_page_token` in the `List` method's response message. This field represents the pagination token to retrieve the next page of results. If the value is `""`, it means no further results for the request.
+To retrieve the next page of results, client **shall** pass the value of response's `next_page_token` in the subsequent `List` method call (in the request message's `page_token` field):
+
+```HTTP
+GET /v2/<name>/manifests/sha256:90659bf80b44ce6be8234e6ff90a1ac34acbeb826903b02cfa0da11c82cbc042/references/list?page_token=1&page_size=10&next_page_token=<token>
+```
+
+The above specifies that a tags response SHOULD be returned, from the start of the result set, ordered lexically, limiting the number of results to `n`. The response to such a request would look as follows:
```
samuelkarp, line 792: I don't believe there's any requirement for a particular ordering.  (I'd prefer that we leave an ordering requirement out so that implementing services can choose to return ordered results or not.)

samuelkarp (commented):
@SteveLasker Thanks for putting this together and taking the feedback from https://github.com/notaryproject/nv2/pull/5; the examples here are now much more clear!

docs/distribution/README.md
SteveLasker: Yes, now that we've changed from manifest to index...

docs/distribution/README.md
SteveLasker: Updated to say: the `net-monitor:v1` image

docs/distribution/README.md
SteveLasker: This is the curx of the design question:
The challenges are:
* The manifest collection is a collection of `application/vnd.oci.image.manifest.v1+json` or `application/vnd.oci.image.index.v1+json` objects. For the artifacts, we use the `config.mediaType` to identify the type. Here' we're mixing descriptor mediaTypes vs. schema mediaTypes. I kinda like your suggestion as all descriptors are currently limited to the two and this makes it a challenge to return a reverse lookup result and not know what config.mediaType (aka artifact type) is represented.  But, it's a bigger change to registry operators. However, it also means signature is just another artifact type, and we use a generic model that could support other artifact types.
* While it could support multiple signatures, that assumes the multiple signatures are uploaded at once. We've said signatures are additive, so this means adding another signature would either be an update to an index, or another index. Seems intermixing these would add complexity. It's also unclear if I can add multiple images and multiple signatures, what does that mean? Are there multiple signatures to all other artifact types in the manifest collection? 
* The main reason to add config to Index is to enable different types of Index based artifact types. For instance CNAB is a multi-artifact collection. Today, the CNAB is an index that just kinda floats out there. They use annotations, which is really hacky, and they store the artifact content in one of the images, another limitation that gets unblocked if we can say an index is a CNAB. `"index.config.mediaType=application/vnd.org.cnab.bundle.v1.config`
* The config is storing data thing is a common question. A config object is the same as a layer object. But, it's called "config" so it's weird. Should an index support a layers collection as well? How impactful would that be to registry operators?
* The purest way is the [Signature Persistence - Option 3: oci-manifest linked through oci-index](https://github.com/notaryproject/nv2/blob/276abe450feec8f3b54f0774b7dfb47f36670cc5/docs/distribution/README.md#signature-persistence---option-3-oci-manifest-linked-through-oci-index)

Since `index.config.mediaType` is consistent with `manifest.config.mediaType` to identify the artifact type, it just seems less "dirty" to store the signature as a config "layer". 
It's a toss up for how much purity, refactoring, or layering we want to implement.
Since all of this is details "within the sausage" and only a few would have to implement this, vs. how it's exposed to end users. I was leaning towards [Signature Persistence - Option 2: oci-index](https://github.com/notaryproject/nv2/blob/276abe450feec8f3b54f0774b7dfb47f36670cc5/docs/distribution/README.md#signature-persistence---option-2-oci-index). 

I do imagine as we continue sketching this out, we'll see a few more opportunities. 
For instance, you'll see we're actually linking from the artifact to the config of the signature index. Instead of the artifact linking to the index digest, which you then have to query again to get the config object.

docs/distribution/README.md
SteveLasker: Would this be a problem for a standard notary client to get different ordered results from different registries?

docs/distribution/README.md
SteveLasker: I've added additional context to the doc to role checking being optional.

docs/distribution/README.md
SteveLasker: The whole point of this prototype is to explore options. I'm resolving this one for clarity of status, but we're very interested in other options. 

We are hopeful these additions will serve other artifact scenarios, and not be limited to signatures.

docs/distribution/README.md
SteveLasker: I've done another scrub on the digests to hopefully provide better clarity. See the table at the bottom of the document.

docs/distribution/README.md
SteveLasker: I've captured it here:: #15

docs/distribution/README.md
SteveLasker: Can we say that without an `orderBy` parameter, the results MUST return by __ order?
Or, should we have a set of required and optional `orderBy` values? I know search is the Achilles heel to these discussions, but I don't know how we get around it. So, what can we do to have something minimal?

docs/distribution/README.md
SteveLasker: I've added sweet #16 to capture the question of roles.

docs/distribution/README.md
SteveLasker: re-reading the question, I think we're talking more about compliance. We've been capturing registry compliance [here](https://github.com/opencontainers/oci-conformance/tree/master/distribution-spec)

docs/distribution/README.md
SteveLasker: I've lost track of which con I'd add to the list. @reasonerjt, can you comment the specific con so I can add? was it recursive signatures?

PR #12, opened 2020-09-04
OCI distribution - referrer metadata proposal

avtakkar:
Add a proposal for tracking referrer metadata for each manifest and using this ability to store and retrieve signatures. The idea is illustrated using a prototype implementation.


aviral26:
Please see https://github.com/notaryproject/nv2/pull/14 instead.

PR #14, opened 2020-09-09
Proposal for generic reverse lookup

aviral26:
Add a proposal for tracking referrer metadata for each manifest and using this ability to store and retrieve signatures. The idea is illustrated using a prototype implementation.


docs/distribution/README.md
```diff
@@ -0,0 +1,566 @@
...
+- [Persisting Referrer Metadata (Signatures)](#persisting-referrer-metadata-signatures)
+- [Implementation](#implementation)
+- [Push, Discover, Pull Prototype](#push-discover-pull-prototype)
+
+## Signature Persistence
+
+Several [options for how to persist a signature were explored][signature-persistance-options] . We measure these options against the [goals of Notary v2][notaryv2-goals], specifically:
+
+- Maintain the original artifact digest and collection of associated tags, supporting existing dev through deployment workflows
+- Multiple signatures per artifact, enabling the originating vendor signature, public registry certification and user/environment signatures
```
reasonerjt, line 24: I don't think how to store the multiple signatures is discussed in this doc:
https://github.com/notaryproject/nv2/blob/prototype-1/docs/distribution/persistance-discovery-options.md

I assume for multiple signatures admin has to push multiple indexes?

docs/distribution/README.md
```diff
@@ -0,0 +1,566 @@
...
+      "size": "1025",
+      "config-mediaType": "application/vnd.oci.image.index.v1+json"
+    }
+  ]
+}
+```
+
+## Persisting Referrer Metadata (Signatures)
+
+The proposal implements a referrer metadata store for manifests that is essentially a reverse-lookup, by `mediaType`, to referrer config objects. For example, when an OCI index is pushed, if it references a config object of media type `application/vnd.cncf.notary.config.v2+jwt`, a link to the config object is recorded in the referrer metadata store of each referenced manifest.
```
reasonerjt, line 247: How do you generate the response in section ` referrer-metadata API` with such data structure?  IMO the linkage is between the referrer and the manifest, rather than between the config object of the referrer and the manifest.

IMO instead of only storing the referrer metadata, you should first store the referrer?

docs/distribution/README.md
```diff
@@ -0,0 +1,566 @@
...
+
+# Retrieve referrer metadata
+curl -s "http://$registry/v2/$repo/manifests/$manifestDigest/referrer-metadata?media-type=$metadataMediaType" | jq
+```
+
+### Verify signature
+
+```shell
+# Retrieve first signature and store it locally
+metadataDigest=`curl -s "http://$registry/v2/$repo/manifests/$manifestDigest/referrer-metadata?media-type=$metadataMediaType" | jq -r '.referrerMetadata[0]'` && \
```
reasonerjt, line 540: It seems the response is not consistent with the one in the section `referrer-metadata API`

docs/distribution/README.md
```diff
@@ -0,0 +1,566 @@
...
+To generalize discovery, a `referrer-metadata` API is proposed to enable discovery of an referenced objects. To support this reverse lookup prototype, additions are proposed to the Notary v2 fork of the reference implementation [docker/distribution][notaryv2-distribution] through [notaryv2-referrer-api].
+
+A referrer is any registry artifact that has an immutable reference to a manifest. An OCI index is a referrer to each manifest it references. The [OCI image spec][oci-image] does not include a config property for an OCI index and there is no reverse lookup of referrers in docker distribution.
+
+A modified OCI index with an `index.config` property that references a collection of manifests allows us to associate a "type" to the referrer-referenced relationship, where the `index.config.mediaType` = `application/vnd.cncf.notary.config.v2+jwt`.
+
+### referrer-metadata API
+
+```HTTP
+`GET http://localhost:5000/v2/net-monitor/manifests/11wma2d22ae5ef400769fa51c84717264cd1520ac8d93dc071374c1be49a11wm/referrer-metadata`
```
reasonerjt, line 210: From API's perspective, IMHO, it would be a better design if we first introduce the `referrer` API, which returns all referrer, including the optionally selected metadata, such as the digest of config blob.
The caller may filter the result by adding `config-mediaType` to the query string.

docs/distribution/README.md
aviral26: /cc @SteveLasker to comment on the requirement for including complete descriptors in the response

@reasonerjt , you're right. The linkage is between the referrer and the manifest. However, with this approach, we get some optimization for referrer metadata lookups since clients would ask for referrer config objects when requesting signatures. What scenarios would require a client to query the list of referrers for a manifest?

SteveLasker:
@reasonerjt, we've moved onto prototype-2, which implements the current [oci.artifact.manifest or #29](https://github.com/opencontainers/artifacts/pull/29)

I'm going to merge this one, just as we're no longer making changes to prototype-1. Please review https://github.com/notaryproject/nv2/tree/prototype-2 and open any issues you'd like to see discussed.

PR #41, opened 2021-03-11
Relate TUF design to Notary v2 goals

mnm678:
Based on discussion in the last meeting, I added a section to clarify why the TUF design is being proposed that puts on paper some of what has previously been discussed in meetings about the use of TUF in Notary v2.

### DIFF
--- a/tuf-design.md
+++ b/tuf-design.md
@@ -9,10 +9,10 @@ This design will be used in conjunction with the OCI artifact work done by other
 The use of TUF in Notary v2 is part of an effort to address the [goals](https://github.com/notaryproject/requirements#goals) of the Notary v2 project.
 Specifically, TUF addresses the following goals:
 * "Signatures attesting to authenticity and/or certification" TUF targets
-   metadata provides signed information about images so that they can be
-   authenticated by users. TUF metadata shows not only that the image
+   metadata provides signed information about artifacts so that they can be
+   authenticated by users. TUF metadata shows not only that the artifact
    and tag have been signed, but also provides clear information about who
-   certified the image and whether they were trusted to do so.
+   certified the artifact and whether they were trusted to do so.
 * "Key hierarchies and delegation". TUF delegations provide a clear key
    hierarchy and allow roles to delegate part of their responsibility to
    another trusted role.
@@ -26,7 +26,7 @@ Specifically, TUF addresses the following goals:
    repository or organization level so that larger organizations may control
    their own root keys, while smaller projects can take advantage of centralized
    root key management on public registries. From these root keys, delegations
-   allow for acquisition of keys for particular images.
+   allow for acquisition of keys for particular artifact.
 
 In addition to providing solutions to the above goals, this document describes
 some variations that have been made to TUF to ensure that it will work with the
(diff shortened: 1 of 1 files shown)


SteveLasker (commented):
Thanks @mnm678 for the updates. I've added comments that reflect the original notes, as the addition of the requirements & goals are helpful.

tuf-design.md
```diff
@@ -4,6 +4,36 @@ Based on discussions with the Notary v2 community over the past few months, we d
 
 This design will be used in conjunction with the OCI artifact work done by others in the Notary v2 community to ensure compliance with the OCI specification and integration with existing registries.
 
+## Alignment with Notary v2 project goals
+
+The use of TUF in Notary v2 is part of an effort to address the [goals](https://github.com/notaryproject/requirements#goals) of the Notary v2 project.
```
NiazFK, line 9: Could we also address the requirements that have been agreed upon in key management?

mnm678:
@SteveLasker Thanks for the review! I added some clarifications inline to address many of your comments.

> > Updating fully qualified references to artifacts
> 
> The initial signature should be fully intact. Including what registry it originated from. No change to the content, or the wraper. This is the integrity check.

In this case, what would a registry do if they want to maintain this original signature as well as re-sign it? Should they re-sign using the original fully qualified reference? I can see two possible options here, in one the signature is copied and the original signer is the primary party trusted to sign the image. In this case any additional signatures would have to match the original name so they could have an exact match. In the other, the artifact is being copied into a private registry and the copier is the party trusted to sign it. In this case, it may be desirable to sign the image using the fully qualified reference in the new registry.

There may also be an in-between option that allows only the last part of the name to match, but this raises questions about the scope of namespaces.

> 
> > Hiding images from other users
> 
> This section implies Coke & Pepsi content can be aggregated in some way, even at some anonymized leve is not permitted.
> 
> It's great to see the requirements get added and covered. The concerns over immutability, and how much updating of content must be maintained is of great concern.
> 
> Consider we might [etched all the immutable content pushed to a registry in glass (Project Silica)](https://news.microsoft.com/innovation-stories/ignite-project-silica-superman/). The basic premise of TUF suggests a continual service runs to continually update snapshot and timestamps of all data. The vast majority of registry content is immutable. Following [best tagging practices](https://stevelasker.blog/2018/03/01/docker-tagging-best-practices-for-tagging-and-versioning-docker-images/), even the tags should be unique, with the important exception of stable tags. And, we do need a solution for tag signing/locking. However, I'd suggest we should be focusing on a solution that doesn't require all data must be continually updated to assure it's still valid. It's a non-scalable solution. If the computers stop processing the data, because the data was etched in glass, put in a mountain, revived years later, is that data considered invalid?

Users who need to download an image today have different security requirements than users who want to download an image from a mountain years later. Users today need a way to ensure that the signature they are verifying was signed by a key that was valid today, and not compromised or expired. This is vital to any mechanism that prioritizes key revocation. Users years from now have the benefit of hindsight to avoid compromised packages, and so can ignore timestamps when verifying metadata. I would argue that it is more important to ensure that users today are downloading secure software than it is to ensure that future users don’t have to override a timestamp check when using legacy software.

tuf-design.md
mnm678: Added!

mnm678:
I'm going to close this in favor of a discussion of goals on the [tuf-notary](https://github.com/notaryproject/tuf) repository.

PR #48, opened 2021-04-05
Add description of key revocation and rescinding signatures

mnm678:
Add these scenarios to the TUF design document.


mnm678:
Moved to notaryproject/tuf#1

PR #49, opened 2021-04-06
Ci 2

marcofranssen:


SteveLasker:
Thanks @marcofranssen, there's a bunch of great cleanup items here, such as the individual vscode settings, .gitignore entries and extensions.

@aviral26, @shizhMSFT can we merge this in, to cleanup things?

shizhMSFT:
@marcofranssen Thanks for the PR.

Can you remove the settings in `.vscode` since developers may have their own configurations?

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/opencontainers/image-spec v1.0.1
 	github.com/urfave/cli/v2 v2.3.0
 	golang.org/x/sys v0.0.0-20210330210617-4fbd30eecc44 // indirect
+	golang.org/x/tools v0.1.0
```
shizhMSFT, line 17: What is this used for?

Makefile
```diff
@@ -1,36 +1,53 @@
...
+vendor: ## vendores the go modules
 	GO111MODULE=on go mod vendor
 
 .PHONY: install-docker-plugins
-install-docker-plugins: $(addprefix bin/,$(DOCKER_PLUGINS))
+install-docker-plugins: $(addprefix bin/,$(DOCKER_PLUGINS)) ## installs the docker plugins
 	cp $(addprefix bin/,$(DOCKER_PLUGINS)) ~/.docker/cli-plugins/
+
+lint: ## use goimports to lint
+	$(GOPATH)/bin/goimports -d -e -local github.com/notaryproject/nv2 $$(go list -f {{.Dir}} ./...)
```
shizhMSFT, line 49: Let's try not embed any go tools in our repo.

go.mod
marcofranssen: These are the tools mentioned in `tools.go`.

Makefile
marcofranssen: This tool is very commonly used in vscode to format the file and organize imports.

This tool enables to verify this using the same settings during CI/CD to maintain a consistent code style and usage of imports.

.github/workflows/golang.yml
```diff
@@ -0,0 +1,49 @@
+name: Go CI
```
shizhMSFT, line 1: nit: Do we have a badge?

.editorconfig
```diff
@@ -0,0 +1,22 @@
+root = true
+
+[*]
```
shizhMSFT, line 3: How are `*.md` files managed?

.gitignore
```diff
@@ -1,6 +1,4 @@
-# VS Code
-.vscode
```
shizhMSFT, line 2: Please keep `.vscode` ignored since not all users are using `vscode` although I am a fan of vscode.

.gitignore
```diff
@@ -1,6 +1,4 @@
-# VS Code
-.vscode
-
-# Custom
-bin/
-vendor/
+# Custom
+bin/
+vendor/
+nv2
```
shizhMSFT, line 4: Why ignore `nv2`? This `nv2` binary is now in the `bin` folder.

.vscode/extensions.json
```diff
@@ -0,0 +1,9 @@
+{
```
shizhMSFT, line 1: Please remove this file since it is under `.vscode`.

.vscode/extensions.json
```diff
@@ -0,0 +1,9 @@
...
+  // See http://go.microsoft.com/fwlink/?LinkId=827846
+  // for the documentation about the extensions.json format
+  "recommendations": [
+    // Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+    "editorconfig.editorconfig",
```
shizhMSFT, line 6: Personally, I don't use `editorconfig` and use `goreturns` for formating go files.

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
```
shizhMSFT, line 1: Please remove this file since it is under `.vscode`.

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
...
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+      {
+          "name": "Launch",
```
shizhMSFT, line 8: This `launch.json` is not really useful since we now have 3 binaries. Therefore, we should have
- Launch nv2
- Launch docker-generate
- Launch docker-nv2

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
```
shizhMSFT, line 1: Please remove this file since it is under `.vscode`.

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
...
+	"go.formatTool": "goimports",
```
shizhMSFT, line 2: This file itself is not well formated.

Makefile
```diff
@@ -1,11 +1,17 @@
-GO_BUILD_FLAGS = 
+GO_BUILD_FLAGS =
 DOCKER_PLUGINS = docker-generate docker-nv2
 COMMANDS       = nv2 $(DOCKER_PLUGINS)
+GOPATH := $(shell go env GOPATH)
```
shizhMSFT, line 4: space ?

Makefile
shizhMSFT: It is common in vscode to use `goimports` to  organize imports but it also gives many options like `goimports`, `goreturns`, or `gofumports`. Furthermore, those tools are usually installed outside of this repo.

For CI, we should use `go fmt` since `goimports` is a superset of `go fmt` with the ability to manage imports. If there is a missing package in the source code, we must treat it as an error and not auto-resolved by `goimports`.

Makefile
```diff
@@ -46,4 +56,12 @@ install-docker-%: bin/docker-%
...
 
 .PHONY: install-docker-plugins
-install-docker-plugins: $(addprefix install-,$(DOCKER_PLUGINS))
+install-docker-plugins: $(addprefix install-,$(DOCKER_PLUGINS)) ## installs the docker plugins
+	cp $(addprefix bin/,$(DOCKER_PLUGINS)) ~/.docker/cli-plugins/
+
+lint: ## use goimports to lint
+	$(GOPATH)/bin/goimports -d -e -local github.com/notaryproject/nv2 $$(go list -f {{.Dir}} ./...)
+
+install-tools: download ## install tools from tools.go
```
shizhMSFT, line 65: based on the above comment. We don't need to install tools if we use `go fmt`.

tools.go
```diff
@@ -0,0 +1,7 @@
+// +build tools
```
shizhMSFT, line 1: As mentioned in above comment, we should use `go fmt` instead of `goimports`. Thus we don't need this.

go.mod
shizhMSFT, line 17: Even if we need go tools, should we always use the latest version of goimports?

tools.go
shizhMSFT, line 1: In this repo, we follow https://github.com/golang-standards/project-layout.

Additional tools should be in `tools` folder instead of `tools.go`.

shizhMSFT (changes requested):

.gitignore
marcofranssen: I would still include it, as it will allow to share those settings for people that are using vscode. It doesn't hurt if people are not using .vscode imho.

.gitignore
marcofranssen: if someone still decides to run `go build` it will still end up as nv2 in the root.

.vscode/extensions.json
marcofranssen: I still believe the .vscode should be included so at least people using vscode can benefit from it. It doesn't hurt if people are not using vscode right?

Makefile
marcofranssen: goimports also takes care of ordering the imports. So they are nicely ordered. That gives some consistency as well on the imports reducing diffs in future code changes.

.editorconfig
marcofranssen: md files would comply with these general settings. Would you prefer an override for some of these settings for md files?

.vscode/extensions.json
marcofranssen: Maybe something we should do, as it also eliminates the need for the fix-line-endings make targets.

.vscode/extensions.json
SteveLasker: Hey @marcofranssen, as much as we love VSCode, it seems common practice to not include tooling specific entries in oss efforts.

.gitignore
shizhMSFT: Then how about other binaries like `docker-nv2` and `docker-generate`? What if we have more later?

.editorconfig
shizhMSFT: We probably should do. /cc @SteveLasker 

Here's an example: https://github.com/containerd/containerd/blob/master/docs/.editorconfig

go.mod
shizhMSFT: Can you add a comment for this import as it is not actually used in the code.

Makefile
marcofranssen: The build will fail anyhow. gomimports will not install it in CI.

tools.go
marcofranssen: I disaggree as the imports will not follow a standardized sorting check.
I started very simple with this CI branch, but eventually I would go with something like `golangci`, which is used by many OSS projects.

here an example from the helm repo.

https://github.com/helm/helm/blob/main/.golangci.yml

.vscode/extensions.json
marcofranssen: @SteveLasker so you are expecting every potential contributor to accordingly setup his own environment. E.g. configuring the goimports to accordingly order the imports statements etc.

If people or not using vscode it doesn't hurt to have this, If they do, they have the settings in place to get them up and running very quickly and have settings that are aligned with the CI checks.

.github/workflows/golang.yml
marcofranssen: Added

Makefile
marcofranssen: I removed the linting from this PR as this is a point of discussion. can create a new PR once we come to common understanding.

.gitignore
marcofranssen: Removed it from this PR. But still think it should be included to get people quickly onboarded.

.gitignore
marcofranssen: Removed it.

.vscode/extensions.json
marcofranssen: Removed it from this PR so we can move forward. Still think we should use this to get new contributors more quickly onboarded.

shizhMSFT (approved):
LGTM

PR #52, opened 2021-04-14
Multiple signatures support

### LINKED ISSUE #51: Support persistence of multiple keys
The current `nv2` client persists the keys at:
` /home/[user]/.docker/nv2/sha256/[image-digest].jwt`

This means if we have 2 or more keys, they'll override them.

What we'd like to support is:
```bash
docker pull registry.wabbit-networks.io/net-monitor:v1
docker tag registry.wabbit-networks.io/net-monitor:v1 registry.acme-rockets.io/net-monitor:v1
# some validation of the image
docker notary sign \
    --key ./wabbit-networks.key \
    --cert ./wabbit-networks.crt \
    registry.acme-rockets.io/net-monitor:v1
docker push registry.acme-rockets.io/net-monitor:v1
```

In the above case, there are 2 signatures on `registry.wabbit-networks.io/net-monitor:v1`, as it was re-signed after some validation, and pushed to `registry.acme-rockets.io/net-monitor:v1`, Both signatures should be pushed.

shizhMSFT:
Resolves #51

### DIFF
--- a/cmd/docker-nv2/push.go
+++ b/cmd/docker-nv2/push.go
@@ -37,29 +37,42 @@ func pushImage(ctx *cli.Context) error {
 	}
 
 	fmt.Println("Pushing signature")
-	sigPath := config.SignaturePath(desc.Digest)
-	sig, err := ioutil.ReadFile(sigPath)
+	sigDigests, err := config.SignatureDigests(desc.Digest)
 	if err != nil {
-		if os.IsNotExist(err) {
-			return errors.New("signature not found")
-		}
 		return err
+	}
+	if len(sigDigests) == 0 {
+		return errors.New("no signatures found")
 	}
 
 	client, err := docker.GetSignatureRepository(ctx.Context, ctx.Args().First())
 	if err != nil {
 		return err
 	}
-	sigDesc, err := client.Put(ctx.Context, sig)
-	if err != nil {
-		return err
-	}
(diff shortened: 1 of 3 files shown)


cmd/docker-nv2/push.go
```diff
@@ -37,29 +37,42 @@ func pushImage(ctx *cli.Context) error {
...
 	if err != nil {
 		return err
 	}
-	sigDesc, err := client.Put(ctx.Context, sig)
-	if err != nil {
-		return err
-	}
+	pushSignature := func(sigDigest digest.Digest) error {
+		sigPath := config.SignaturePath(desc.Digest, sigDigest)
+		sig, err := ioutil.ReadFile(sigPath)
```
marcofranssen, line 54: ioutil should not be used anymore as it is deprecated since go 1.16. 

Recommend is to use the replacements from the `os` package.

https://golang.org/doc/go1.16#ioutil

SteveLasker (approved):
LGTM for the dual signature support
Defer to @shizhMSFT  on the question @marcofranssen raised

cmd/docker-nv2/push.go
shizhMSFT: Good catch as this code is just migrated from `1.14 to 1.16`. I've resolved it and may scan other code later.

PR #54, opened 2021-04-16
Persist signatures when pulling

### LINKED ISSUE #51: Support persistence of multiple keys

shizhMSFT:
Resolves #51

### DIFF
--- a/cmd/docker-nv2/pull.go
+++ b/cmd/docker-nv2/pull.go
@@ -4,10 +4,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 
 	"github.com/docker/distribution/reference"
 	"github.com/notaryproject/notary/v2"
+	"github.com/notaryproject/nv2/cmd/docker-nv2/config"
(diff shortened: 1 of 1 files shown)


SteveLasker (approved):
LGTM

PR #62, opened 2021-05-15
add addon opa

honeybadger10:


SteveLasker:
Pinging @ritazh for :eyes:

addons/opa/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+# create ns
+kubectl create namespace opa
+
+
+# change context to ns opa
+
+
+# generate certs
+openssl genrsa -out ca.key 2048
```
marcofranssen, line 19: can we move this into a code block like

\`\`\`bash
openssl …………
\`\`\`

addons/opa/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+# create pod and/or deployment
+# for the deployment to succeed, OPA needs to be able to reach the registry
+# change the registry in pod.yaml and deployment.yaml to your registrys IP/FQDN
+kubectl apply -f app/pod.yaml
+kubectl apply -f app/deployment.yaml
+```
+
+
+## denied deployment
```
marcofranssen, line 88:
```suggestion
## denied deployment

```

addons/opa/README.md
```diff
@@ -0,0 +1,103 @@
...
+        name: opa
+EOF
+
+# create webhook
+kubectl apply -f webhook-configuration.yaml
+
+```
+
+
+## allowed deployment
```
marcofranssen, line 67:
```suggestion
## allowed deployment

```

addons/opa/README.md
```diff
@@ -0,0 +1,103 @@
+# nv2-opa
+
+## install OPA
```
marcofranssen, line 3:
```suggestion
## install OPA

```

marcofranssen (changes requested):
Some minor things captured by markdown linting.

Furthermore I think it would be great to template the k8s resources in a helm chart to have less manual setup steps and replacements in the k8s objects.

It is a good practice to put helm charts in a `charts` folder in the root of the repository.

See here some examples.

https://github.com/philips-labs/helm-charts

See here an example outline to have a quickstart.

```bash
tree charts
charts
└── opa
    ├── Chart.yaml
    ├── LICENSE
    ├── templates
    │   ├── NOTES.txt
    │   ├── _helpers.tpl
    │   ├── admission-controller-configmap.yaml
    │   ├── admission-controller-deployment.yaml
    │   └── admission-controller-service.yaml
    └── values.yaml

2 directories, 8 files
```

You can create this using `helm create` 

```bash
helm create --help

This command creates a chart directory along with the common files and
directories used in a chart.

For example, 'helm create foo' will create a directory structure that looks
something like this:

    foo/
    ├── .helmignore   # Contains patterns to ignore when packaging Helm charts.
    ├── Chart.yaml    # Information about your chart
    ├── values.yaml   # The default values for your templates
    ├── charts/       # Charts that this chart depends on
    └── templates/    # The template files
        └── tests/    # The test files
```

addons/opa/rego/nv2.rego
```diff
@@ -0,0 +1,195 @@
...
+
+
+operations := {"CREATE", "UPDATE"}
+kind := {"Deployment", "Pod"}
+
+artifactType := "application/vnd.cncf.notary.v2"
+mediaType := "application/vnd.cncf.notary.signature.v2+jwt"
+
+
+## tbd: how to load certs
```
ritazh, line 12: Is there a missing step here?

addons/opa/rego/nv2.rego
```diff
@@ -0,0 +1,195 @@
...
+}
+
+## get nv2 signature
+get_nv2_jwt(image_obj, digest) = nv2_jwt {
+  
+	url_digest_arr := [ "https:/", image_obj.registry, "v2", image_obj.image, "blobs", digest ]
+  url_digest := concat("/", url_digest_arr)
+  
+	headers_json := {"Content-Type": "application/json"}
+  response_signature := http.send({"method": "get", "url": url_digest, "headers": headers_json, "tls_insecure_skip_verify": true})
```
ritazh, line 165: can you give an example of an url that returns a signature?

sozercan:
does this work with private registries?

ritazh:
In general, it would be helpful to include instructions for:
- prerequisites for which registry supports this feature and how to enable the feature on the registry
- setting up nv2

SteveLasker:
> does this work with private registries? or public ones that requires tokens?

Oh, that's a good point. CNCF distribution doesn't have auth, but any cloud deployment would need to validate with an authenticated connection to the registry. Most cloud providers would have a Service Identity like configuration, so I'm not sure how that associates with OPA/Gatekeeper as customers are really trying to move away from any secrets model as it complicates key rotation.

SteveLasker:
> In general, it would be helpful to include instructions for:
> * prerequisites for which registry supports this feature and how to enable the feature on the registry

At this point, nobody supports it as we're still incubating the idea, so only a local instance of CNCF distribution will work. 
Fall 2021, that's another story :)

SteveLasker:
Adding @mnltejaswini ^

SteveLasker:
@honeybadger10, just a ping. We haven't forgotten about this. There's some work happening to enable data providers to enable the passthrough of data within OPA to enable the flows.  
We should be able to pop this up again soon.

dtzar:
@honeybadger10 - thank you for your PR.  It's been inactive for a long time and we want to avoid specific implementations which utilize notation in this repository, so I'm closing it out.  Perhaps you could put something into your own repository?

For instance, we have [ratify](https://github.com/deislabs/ratify) which utilizes notation v2 with gatekeeper and there are others like [connaisseur](https://github.com/sse-secure-systems/connaisseur) - currently only supports notary v1.

PR #68, opened 2021-08-13
Added certificates generate command

### LINKED ISSUE #66: Add key generation support to nv2
The nv2 binary doesn't currently have a way to generate a key pair for signing which makes it harder than it needs to be for new users to use nv2. Please add a subcommand that generates a keypair.

I suggest `nv2 keygen` as a subcommand. At a minimum, it needs to perform the functional equivalent of the following as mentioned [here](https://github.com/notaryproject/nv2/blob/prototype-2/docs/nv2/demo-script.md).

```
openssl req \
  -x509 \
  -sha256 \
  -nodes \
  -newkey rsa:2048 \
  -days 365 \
  -subj "/CN=${REGISTRY}/O=wabbit-networks inc/C=US/ST=Washington/L=Seattle" \
  -addext "subjectAltName=DNS:${REGISTRY}" \
  -keyout ./wabbit-networks.key \
  -out ./wabbit-networks.crt
```

sajayantony:
The command will generate an 2084 bits rsa based crt and key file.
The files will be placed in $HOME/.notary/keys

Signed-off-by: Sajay Antony <sajaya@microsoft.com>

Fixes #66

```bash
❯ ./bin/nv2 certificates generate wabbit-networks
Generating RSA Key with 3072 bits
Generated certificates expiring on 2022-08-25T12:22:00-07:00
Wrote self-signed certificate file: /home/sajay/.notation/keys/wabbit-networks.crt
Wrote private key file: /home/sajay/.notation/keys/wabbit-networks.key
```

### DIFF
--- a/cmd/nv2/gencert.go
+++ b/cmd/nv2/gencert.go
@@ -21,6 +21,9 @@ import (
 )
 
 const (
+	// This needs to be configurable. Once the location of the
+	// configuration is finalized this parameter and file locations
+	// should be exposed as options on the cli with mock tests.
 	KEYS_BASE_DIR = ".notary"
 )
 
@@ -31,11 +34,17 @@ var generateCertCommand = &cli.Command{
 	Flags: []cli.Flag{
 		&cli.IntFlag{
 			Name:     "rsaBits",
-			Usage:    "rsaBits 2048",
+			Usage:    "--rsaBits 3072",
 			Required: false,
-			Value:    2048,
+			Value:    3072,
+		},
+		&cli.StringFlag{
+			Name:     "not-after",
+			Usage:    "--not-after 2006-01-02T15:04:05-07:00 (default is 1 year)",
+			Required: false,
 		},
 	},
+
 	Action: runGenerateCert,
 }
 
@@ -74,9 +83,26 @@ func runGenerateCert(ctx *cli.Context) error {
 		return errors.New("Missing required [host] parameter")
 	}
 
+	// Set certificate validity
+	notBefore := time.Now()
+	notAfter := time.Now().Add(time.Duration(365 * 24 * time.Hour))
+
+	expiry := ctx.String("not-after")
+	var err error
+	if len(expiry) != 0 {
+		notAfter, err = time.Parse(time.RFC3339, expiry)
+		if err != nil {
+			return fmt.Errorf("Invalid --not-after %s value specified %v", expiry, err)
+		}
+	}
+
+	if notAfter.Before(notBefore) {
+		return fmt.Errorf("Invalid --not-after that is earlier than not-before [%s] specified", notBefore.Format(time.RFC3339))
+	}
+
+	// Generate RSA Bits
 	rsaBits := ctx.Int("rsaBits")
 	var priv interface{}
-	var err error
 
 	fmt.Printf("Generating RSA Key with %d bits\n", rsaBits)
 	priv, err = rsa.GenerateKey(rand.Reader, rsaBits)
@@ -88,16 +114,6 @@ func runGenerateCert(ctx *cli.Context) error {
 	// ECDSA, ED25519 and RSA subject keys should have the DigitalSignature
 	// KeyUsage bits set in the x509.Certificate template
 	keyUsage := x509.KeyUsageDigitalSignature
-	// Only RSA subject keys should have the KeyEncipherment KeyUsage bits set. In
-	// the context of TLS this KeyUsage is particular to RSA key exchange andgit st
-	// authentication.
-	if _, isRSA := priv.(*rsa.PrivateKey); isRSA {
-		keyUsage |= x509.KeyUsageKeyEncipherment
-	}
-
-	// Set certificate validity to one year from now.
-	notBefore := time.Now()
-	notAfter := notBefore.Add(time.Duration(365 * 24 * time.Hour))
 
 	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
 	serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
@@ -106,15 +122,10 @@ func runGenerateCert(ctx *cli.Context) error {
 	}
 
 	template := x509.Certificate{
-		SerialNumber: serialNumber,
-		Subject: pkix.Name{
-			Organization: []string{"Acme Co"},
-		},
-		NotBefore: notBefore,
-		NotAfter:  notAfter,
-
+		SerialNumber:          serialNumber,
+		NotBefore:             notBefore,
+		NotAfter:              notAfter,
 		KeyUsage:              keyUsage,
-		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
 		BasicConstraintsValid: true,
 	}
 
@@ -125,6 +136,10 @@ func runGenerateCert(ctx *cli.Context) error {
 		} else {
 			template.DNSNames = append(template.DNSNames, h)
 		}
+	}
+
+	template.Subject = pkix.Name{
+		Organization: []string{hosts[0]},
 	}
 
 	derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, publicKey(priv), priv)
@@ -155,6 +170,8 @@ func runGenerateCert(ctx *cli.Context) error {
 	if err := certOut.Close(); err != nil {
 		return fmt.Errorf("Error closing cert.pem: %v", err)
 	}
+
+	fmt.Printf("Generating certificates expiring on %s\n", notAfter.Format(time.RFC3339))
 	fmt.Printf("Writing public key file: %s\n", crtFilePath)
 
 	// Write the private key file
--- a/cmd/nv2/gencert.go
+++ b/cmd/nv2/gencert.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"crypto"
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
@@ -24,11 +25,19 @@ const (
 	// This needs to be configurable. Once the location of the
 	// configuration is finalized this parameter and file locations
 	// should be exposed as options on the cli with mock tests.
-	KEYS_BASE_DIR = ".notary"
+	KEYS_BASE_DIR = ".notation"
 )
 
+var certificatesCommand = &cli.Command{
+	Name:  "certificates",
+	Usage: "Commands to manage certificates",
+	Subcommands: []*cli.Command{
+		generateCertCommand,
+	},
+}
+
 var generateCertCommand = &cli.Command{
-	Name:      "generate-certificates",
+	Name:      "generate",
 	Usage:     "Generates a test crt and key file.",
 	ArgsUsage: "[host]",
 	Flags: []cli.Flag{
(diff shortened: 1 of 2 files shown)


sajayantony:
@lachie83 - This composes with  my other PR that will make adding and removing the certs much cleaner - #70

cmd/nv2/gencert.go
```diff
@@ -0,0 +1,184 @@
...
+	template := x509.Certificate{
+		SerialNumber: serialNumber,
+		Subject: pkix.Name{
+			Organization: []string{"Acme Co"},
+		},
+		NotBefore: notBefore,
+		NotAfter:  notAfter,
+
+		KeyUsage:              keyUsage,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
```
dlorenc, line 117: Why is ServerAuth set here?

cmd/nv2/gencert.go
```diff
@@ -0,0 +1,184 @@
...
+
+	if err != nil {
+		return fmt.Errorf("Failed to generate private key: %v", err)
+	}
+
+	// ECDSA, ED25519 and RSA subject keys should have the DigitalSignature
+	// KeyUsage bits set in the x509.Certificate template
+	keyUsage := x509.KeyUsageDigitalSignature
+	// Only RSA subject keys should have the KeyEncipherment KeyUsage bits set. In
+	// the context of TLS this KeyUsage is particular to RSA key exchange andgit st
```
dlorenc, line 92: I'm not sure why keyencipherment is set. Are these certs used for TLS somehow?

cmd/nv2/gencert.go
```diff
@@ -0,0 +1,184 @@
...
+	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
+	serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
+	if err != nil {
+		return fmt.Errorf("Failed to generate serial number: %v", err)
+	}
+
+	template := x509.Certificate{
+		SerialNumber: serialNumber,
+		Subject: pkix.Name{
+			Organization: []string{"Acme Co"},
```
trishankatdatadog, line 111: Why hardcode to `Acme Co`?

cmd/nv2/gencert.go
```diff
@@ -0,0 +1,184 @@
...
+	// Only RSA subject keys should have the KeyEncipherment KeyUsage bits set. In
+	// the context of TLS this KeyUsage is particular to RSA key exchange andgit st
+	// authentication.
+	if _, isRSA := priv.(*rsa.PrivateKey); isRSA {
+		keyUsage |= x509.KeyUsageKeyEncipherment
+	}
+
+	// Set certificate validity to one year from now.
+	notBefore := time.Now()
+	notAfter := notBefore.Add(time.Duration(365 * 24 * time.Hour))
```
trishankatdatadog, line 100: Should ideally be user-configurable.

cmd/nv2/gencert.go
```diff
@@ -0,0 +1,184 @@
...
+var generateCertCommand = &cli.Command{
+	Name:      "generate-certificates",
+	Usage:     "Generates a test crt and key file.",
+	ArgsUsage: "[host]",
+	Flags: []cli.Flag{
+		&cli.IntFlag{
+			Name:     "rsaBits",
+			Usage:    "rsaBits 2048",
+			Required: false,
+			Value:    2048,
```
trishankatdatadog, line 36: Might be a good idea to default to at least 3072 to try to buy some futureproofing.

lachie83: Also user-configurable with a sane set of defaults?

cmd/nv2/gencert.go
sajayantony: I'll remove all TLS related usage here since these are only tests certs for signing.

cmd/nv2/gencert.go
```diff
@@ -0,0 +1,201 @@
...
+			return "", err
+		}
+	}
+
+	return keysDir, nil
+}
+
+func runGenerateCert(ctx *cli.Context) error {
+
+	host := ctx.Args().First()
```
shizhMSFT, line 81: Why not accept an argument list instead of a signle argument, making `host` a `[]string` slice?

cmd/nv2/gencert.go
```diff
@@ -0,0 +1,201 @@
...
+		if ip := net.ParseIP(h); ip != nil {
+			template.IPAddresses = append(template.IPAddresses, ip)
+		} else {
+			template.DNSNames = append(template.DNSNames, h)
+		}
+	}
+
+	template.Subject = pkix.Name{
+		Organization: []string{hosts[0]},
+	}
```
shizhMSFT, line 143: Missing `CommonName` in the template.

cmd/nv2/gencert.go
```diff
@@ -0,0 +1,201 @@
...
+
+	keysDir := filepath.Join(dirname, KEYS_BASE_DIR, "keys")
+	_, err = os.Stat(keysDir)
+	if os.IsNotExist(err) {
+		err = os.MkdirAll(keysDir, 0700) // Only user has permissions on this directory
+		if err != nil {
+			return "", err
+		}
+	}
```
shizhMSFT, line 74: if `os.IsNotExists(err)` returns `false`, we should return `err`.

cmd/nv2/gencert.go
```diff
@@ -0,0 +1,201 @@
...
+
+	return keysDir, nil
```
shizhMSFT, line 76: We need to ensure that `keysDir` is really a directory. It could be a regular file or symbolic link.

cmd/nv2/gencert.go
```diff
@@ -0,0 +1,201 @@
...
+	certOut, err := os.OpenFile(crtFileName, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0600)
+	if err != nil {
+		return fmt.Errorf("Failed to open %s for writing: %v", crtFilePath, err)
+	}
+
+	if err := pem.Encode(certOut, &pem.Block{Type: "CERTIFICATE", Bytes: derBytes}); err != nil {
+		return fmt.Errorf("Failed to write data to %s: %v", crtFilePath, err)
+	}
+	if err := certOut.Close(); err != nil {
+		return fmt.Errorf("Error closing cert.pem: %v", err)
```
shizhMSFT, line 171: Should be error closing `crtFilePath`.

cmd/nv2/gencert.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	fmt.Printf("Generating certificates expiring on %s\n", notAfter.Format(time.RFC3339))
+	fmt.Printf("Writing public key file: %s\n", crtFilePath)
```
shizhMSFT, line 175: Should be `Generated` and `Wrote`.

cmd/nv2/gencert.go
```diff
@@ -0,0 +1,201 @@
...
+	if err != nil {
+		return fmt.Errorf("Unable to marshal private key: %v", err)
+	}
+	if err := pem.Encode(keyOut, &pem.Block{Type: "PRIVATE KEY", Bytes: privBytes}); err != nil {
+		return fmt.Errorf("Failed to write data to key.pem: %v", err)
+	}
+	if err := keyOut.Close(); err != nil {
+		return fmt.Errorf("Error closing %s: %v", keyFilePath, err)
+	}
+	fmt.Printf("Writing private key file: %s\n", keyFilePath)
```
shizhMSFT, line 199: Should be `Wrote`.

cmd/nv2/gencert.go
shizhMSFT: BTW, it is not just a public key, but a self-signed cert.

cmd/nv2/gencert.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	hosts := strings.Split(host, ",")
+	for _, h := range hosts {
+		if ip := net.ParseIP(h); ip != nil {
+			template.IPAddresses = append(template.IPAddresses, ip)
+		} else {
+			template.DNSNames = append(template.DNSNames, h)
+		}
+	}
```
priteshbandi, line 146: as this is codesigning not ssl certificate, we don't need to support DNSName or ip-address.

cmd/nv2/gencert.go
```diff
@@ -0,0 +1,209 @@
...
+
+	template.Subject = pkix.Name{
+		Organization: []string{hosts[0]},
+		CommonName:   hosts[0],
```
priteshbandi, line 150: Can we take `Organization Name` from user instead of host? host is confusing as certificate is not used for tls.

cmd/nv2/gencert.go
```diff
@@ -0,0 +1,209 @@
...
+	derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, priv.Public(), priv)
+	if err != nil {
+		return fmt.Errorf("Failed to create certificate: %v", err)
+	}
+
+	// Write the crt public key file
+	keysDir, err := ensureKeysDir()
+	if err != nil {
+		return fmt.Errorf("Could not access keys directory: %v", err)
+	}
```
priteshbandi, line 162: Should be done as first thing after validating arguments.

cmd/nv2/gencert.go
```diff
@@ -0,0 +1,209 @@
...
+		},
+	},
+
+	Action: runGenerateCert,
+}
+
+func ensureKeysDir() (string, error) {
+
+	// Expected to ensure ~/.notation/keys
+	dirname, err := os.UserHomeDir()
```
priteshbandi, line 63: As discussed in #76 we should support XDG_Base_Directory.

cmd/nv2/gencert.go
```diff
@@ -0,0 +1,209 @@
...
+
+	crtFileName := path.Join(keysDir, hosts[0]+".crt")
+	crtFilePath, err := filepath.Abs(crtFileName)
+	if err != nil {
+		return fmt.Errorf("Unable to get full path of the file: %v", err)
+	}
+
+	certOut, err := os.OpenFile(crtFileName, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0600)
+	if err != nil {
+		return fmt.Errorf("Failed to open %s for writing: %v", crtFilePath, err)
```
priteshbandi, line 172: s/open/create file?

cmd/nv2/gencert.go
```diff
@@ -0,0 +1,209 @@
...
+	ArgsUsage: "[host]",
+	Flags: []cli.Flag{
+		&cli.IntFlag{
+			Name:     "rsaBits",
+			Usage:    "--rsaBits 3072",
+			Required: false,
+			Value:    3072,
+		},
+		&cli.StringFlag{
+			Name:     "not-after",
```
priteshbandi, line 51: Will it be a better customer experience to take validity period `validForDays` instead of actual date?
Pros: 
* easy to associate default value of 1 year
* No need to worry about not-after being in past.

```
certificates generate --not-after 2006-01-02T15:04:05-07:00
certificates generate --valid-for-days 365
```

sajayantony:
@priteshbandi #83 PR adds all this capability into notation alpha. 
I will close this draft PR and we should be now trying to get this all into one alpha release which we can iterate on is my thinking.

@shizhMSFT - can see if there are inputs here you might not have already considered in your PR?

cmd/nv2/gencert.go
shizhMSFT: We still need them as the `x509` package checks `SAN` instead of `CN` starting from golang `1.15`. See https://golang.org/doc/go1.15#commonname

cmd/nv2/gencert.go
shizhMSFT: `Organization` is removed in the Notation CLI Alpha PR.

cmd/nv2/gencert.go
shizhMSFT: Notation CLI Alpha honors XDG directories.

cmd/nv2/gencert.go
shizhMSFT: `validForDays` might not be a good one. For testing purposes, people may want short-lived certs like serveral hours or minutes.

PR #76, opened 2021-08-27
Configuration file spec

SteveLasker:
Capturing the configuration file, at its current state


specs/notation-config.md
```diff
@@ -0,0 +1,59 @@
...
+
+To enable persisted configuration, simplifying the execution of the `notation` cli, the following configuration file will be available
+
+> Note: there will be a policy based configuration that will come at a later point.
+
+## Location
+
+The default location and file will be stored at: `~/.notary/notation-config.json`. The `notation` cli and libraries will support alternate locations through a `config-location` parameters.
+
+## Properties
```
juliusl, line 11: I would write this as a table,

Property | Type |  Value
------ | ------ | ---
`enabled`|_bool_|on/off switch for default notation behavior. --enabled false would disable all automatic validations
`verificationCerts.certs`|_array_|collection of name/value pairs for a collection of public certs that are used for verification. These may be replaced with a future policy configuration.
`cert.name`|_string_|a named reference to the certificate
`cert.path`|_string_|a location by which the certificate can be found by the notation cli or notation libraries
`signing-keys.keys`|_array_|a collection of name/value pairs of signing keys.
`key.name`|_string_|a named reference to the key
`key.path`|_string_|a location by which the key can be found by the notation cli or notation libraries
`signing-keys.default`|_string_|the signing key to be used when `notation sign` is called without `--name`
`insecureRegistries`|_array_|a list of registries that may be used without https

juliusl (approved):

specs/notation-config.md
```diff
@@ -0,0 +1,59 @@
+# Notation Configuration
+
+To enable persisted configuration, simplifying the execution of the `notation` cli, the following configuration file will be available
+
+> Note: there will be a policy based configuration that will come at a later point.
+
+## Location
+
+The default location and file will be stored at: `~/.notary/notation-config.json`. The `notation` cli and libraries will support alternate locations through a `config-location` parameters.
```
shizhMSFT, line 9: Will `~/.notation/config.json` be better?

shizhMSFT, line 9: It seems for Unix / Linux only. How about the locations on Windows or macOS?

specs/notation-config.md
```diff
@@ -0,0 +1,59 @@
...
+
+- `enabled` - bool, which acts as an on/off switch for default notation behavior. --enabled false would disable all automatic validations
+- `verificationCerts` - collection of name/value pairs for a collection of public certs that are used for verification. These may be replaced with a future policy configuration.
+  - `name` - a named reference to the certificate
+  - `path` - a location by which the certificate can be found by the notation cli or notation libraries
+- `signing-keys` - a collection of name/value pairs of signing keys.
+  - `name` - a named reference to the key
+  - `path` - a location by which the key can be found by the notation cli or notation libraries
+  - `default` - the signing key to be used when `notation sign` is called without `--name`
+- `insecureRegistries` - a list of registries that may be used without https
```
shizhMSFT, line 21: Should `insecureRegistries` be called `plainHTTP`?

specs/notation-config.md
```diff
@@ -0,0 +1,59 @@
...
+      {
+        "name": "wabbit-networks.io",
+        "path": "~/./notary/keys/wabbit-networks.crt"
+      },
+      {
+        "name": "import.acme-rockets.io",
+        "path": "~/./notary/keys/import-acme-rockets.crt"
+      }
+    ]
+  },
```
shizhMSFT, line 41: Why do we need `certs` inside of `verificationCerts`? It is redundant.

specs/notation-config.md
```diff
@@ -0,0 +1,59 @@
...
+## Location
+
+The default location and file will be stored at: `~/.notary/notation-config.json`. The `notation` cli and libraries will support alternate locations through a `config-location` parameters.
+
+## Properties
+
+- `enabled` - bool, which acts as an on/off switch for default notation behavior. --enabled false would disable all automatic validations
+- `verificationCerts` - collection of name/value pairs for a collection of public certs that are used for verification. These may be replaced with a future policy configuration.
+  - `name` - a named reference to the certificate
+  - `path` - a location by which the certificate can be found by the notation cli or notation libraries
```
shizhMSFT, line 16: We need to define whether it is an absolute path or it can be a relative path or both.

specs/notation-config.md
```diff
@@ -0,0 +1,59 @@
...
+- `signing-keys` - a collection of name/value pairs of signing keys.
+  - `name` - a named reference to the key
+  - `path` - a location by which the key can be found by the notation cli or notation libraries
```
shizhMSFT, line 19: Same here for absolute or / and relative path.

shizhMSFT (changes requested):

specs/notation-config.md
```diff
@@ -0,0 +1,59 @@
...
+
+The default location and file will be stored at: `~/.notary/notation-config.json`. The `notation` cli and libraries will support alternate locations through a `config-location` parameters.
+
+## Properties
+
+- `enabled` - bool, which acts as an on/off switch for default notation behavior. --enabled false would disable all automatic validations
+- `verificationCerts` - collection of name/value pairs for a collection of public certs that are used for verification. These may be replaced with a future policy configuration.
+  - `name` - a named reference to the certificate
+  - `path` - a location by which the certificate can be found by the notation cli or notation libraries
+- `signing-keys` - a collection of name/value pairs of signing keys.
```
shizhMSFT, line 17: Can we make naming convension consistant? like `signingKeys`?

specs/notation-config.md
```diff
@@ -0,0 +1,59 @@
+# Notation Configuration
+
+To enable persisted configuration, simplifying the execution of the `notation` cli, the following configuration file will be available
+
+> Note: there will be a policy based configuration that will come at a later point.
+
+## Location
```
jonjohnsonjr, line 7: Consider following the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) instead of just [polluting my home directory](https://0x46.net/thoughts/2019/02/01/dotfile-madness/).

specs/notation-config.md
SteveLasker: I was trying to align with `keys` which has `default` nested, so it needed `keys`. 
Suggestions?

specs/notation-config.md
SteveLasker: Sure, what should we articulate those?

SteveLasker: `notation` is one of the cli's, but the configuration would be for notary. I'm trying to avoid having too much folder sprawl.

specs/notation-config.md
SteveLasker: Good question. What do you think from a security perspective?

specs/notation-config.md
SteveLasker: I've added a TODO section so we can merge and come back to address

specs/notation-config.md
SteveLasker: For better or worse, I was copying the docker config. which is what we have in prototype-2. 
I'm going to leave for now. If folks want to change, can we open an issue to discuss?
```json
{
  "insecure-registries": [
    "registry.wabbit-networks.io",
    "localhost"
  ]
}
```

specs/notation-config.md
SteveLasker: Leaving for this PR, and lets address how we want to handle `verificationCerts` and `signingKeys`
Do we want to pull default out to a root node?

specs/notation-config.md
SteveLasker: Good question. What do you think from a security perspective?

specs/notation-config.md
SteveLasker: noting for evaluation. 
@jonjonsonjr, can you point to some other implementations that have used XDG Base Directories?
I found:
- https://github.com/kubernetes/minikube/issues/4109 which also deferred if someone wants to support.
- https://golangrepo.com/repo/adrg-xdg-go-miscellaneous

jonjohnsonjr: https://wiki.archlinux.org/title/XDG_Base_Directory#Support

specs/notation-config.md
shizhMSFT: It is better if it is an absolute path.

specs/notation-config.md
```diff
@@ -0,0 +1,63 @@
...
+
+The default location and file will be stored at: `~/.notary/notation-config.json`. The `notation` cli and libraries will support alternate locations through a `config-location` parameters.
+
+> TODO: Add Windows and Mac locations
+
+## Properties
+
+Property | Type |  Value
+------ | ------ | ---
+`enabled`|_bool_|on/off switch for default notation behavior. --enabled false would disable all automatic validations
```
shizhMSFT, line 17: `enabled` is used for `docker-notation`. This field is not meaningful for the `notation` CLI.

specs/notation-config.md
shizhMSFT: I am following
- https://pkg.go.dev/os#UserCacheDir
- https://pkg.go.dev/os#UserConfigDir

specs/notation-config.md
sajayantony: @SteveLasker - Moving this into the notary location exposes the notation CLI to need ACLs and policise that are present with notaryV1. My recommendatiaon is to have this under .notation for now and open an issue to unify the config location with notaryv1 if that is the requirement. 
/cc @gokarnm

specs/notation-config.md
jonjohnsonjr: > On Unix systems, it returns $XDG_CONFIG_HOME as specified by https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html if non-empty, else $HOME/.config. On Darwin, it returns $HOME/Library/Application Support. On Windows, it returns %AppData%. On Plan 9, it returns $home/lib.

This is perfect, thanks!

specs/notation-config.md
sajayantony: For now I feel we can go with  @shizhMSFT's recommendation and if we want to move the default location or file name we can track that as a separate item. The change currently will only affect directories that are typical locations for this specific CLI.

specs/notation-config.md
```diff
@@ -0,0 +1,63 @@
...
+      {
+        "name": "wabbit-networks.io",
+        "path": "~/./notary/keys/wabbit-networks.key"
+      },
+      {
+        "name": "import.acme-rockets.io",
+        "path": "~/./notary/keys/import-acme-rockets.key"
+      }
+    ]
+  },
```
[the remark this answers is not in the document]
gokarnm: +1 on requiring signing certificate in addition to private key.

specs/notation-config.md
SteveLasker: > is a requirement, and the motivation for unifying config with notary V1.

It was more about reducing folder sprawl. I'm trying to avoid end users having to "feel the pain" of multiple folders being created as a result of using v1 and v2. 
We don't have much usage of notary v1, so it's a minimal issue. 
It's not the most pressing issue, so you could also argue, if most users don't have notary v1, is it really folder sprawl as they'll only have a single folder, "./notation".
I'll update to use notation and not be something to take more time for us to think about

SteveLasker: > are we thinking of supporting truststore and trustpolicy evaluation

Yes, we should incorporate truststore policy. However, suggesting we add that as we iterate the details of the trust store policy. This will document the current state, so we can release a notation cli that we can iterate with. 
As we complete the truststore policy design and implementation, we can update this config spec.

specs/notation-config.md
SteveLasker: @shizhMSFT, this is in the current config file. Are we having separate configs for docker-notation and notation?

specs/notation-config.md
```diff
@@ -0,0 +1,63 @@
...
+------ | ------ | ---
+`enabled`|_bool_|on/off switch for default notation behavior. --enabled false would disable all automatic validations
+`verificationCerts.certs`|_array_|collection of name/value pairs for a collection of public certs that are used for verification. These may be replaced with a future policy configuration.
+`cert.name`|_string_|a named reference to the certificate
+`cert.path`|_string_|a location by which the certificate can be found by the notation cli or notation libraries
+`signing-keys.keys`|_array_|a collection of name/value pairs of signing keys.
+`key.name`|_string_|a named reference to the key
+`key.path`|_string_|a location by which the key can be found by the notation cli or notation libraries
+`signing-keys.default`|_string_|the signing key to be used when `notation sign` is called without `--name`
+`insecureRegistries`|_array_|a list of registries that may be used without https
```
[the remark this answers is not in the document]
SteveLasker: notation pushes and pulls content to a registry using the oras-go libraries. insecureRegistires is a list of registries enables the flow, as an exception to the normal HTTPS required flag. 
It's not just for local instances of distribution. Many IoT scenarios run non https scenarios.

SteveLasker:
This reflects the current implementation of #83 

Suggestion to merge this and #83, and open Issues and PRs for the proposed name changes. 
This provides a baseline to test and iterate with issues on functioning bits from main.

specs/notation-config.md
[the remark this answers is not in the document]
priteshbandi: opened issue: https://github.com/notaryproject/notation/issues/89

NiazFK (approved):
LGTM with https://github.com/notaryproject/notation/issues/89

PR #78, opened 2021-08-27
hello signing samples

SteveLasker:
These are the target commands, lifting the [prototype-2](https://github.com/notaryproject/notation/tree/prototype-2) efforts to the nearly RC state of the [artifacts-spec](https://github.com/oras-project/artifacts-spec/)

Note: these commands don't work, **_yet_**. This is the target for what will work.
Please iterate any feedback here and we can incorporate into the work to support these commands.

PR #87, opened 2021-09-07
Add Notation Quick Start

SteveLasker:
Adds a basic scenario to the root readme.


README.md
```diff
@@ -4,12 +4,31 @@ Notation is a project to add signatures as standard items in the registry ecosys
...
+  ```bash
+  curl -Lo notation.tar.gz https://github.com/shizhMSFT/notation/releases/download/v0.5.2/notation_0.5.2_linux_amd64.tar.gz
+  tar xvzf notation.tar.gz -C ~/bin notation
+  ```
+- Build, Push, Sign, Verify the `net-monitor` software
+  ```bash
+  export IMAGE=localhost:5000/net-monitor:v1
+  docker build -t $IMAGE https://github.com/wabbit-networks/net-monitor.git#main
+  docker push $IMAGE
+  notation cert generate-test --default --trust "wabbit-networks-dev"
```
gokarnm, line 26: It's not clear what --default and --trust do, and what is wabbit-networks-dev. We can add comments where we want to explain what a particular notation command does.

README.md
```diff
@@ -4,12 +4,31 @@ Notation is a project to add signatures as standard items in the registry ecosys
...
+  notation sign $IMAGE
```
gokarnm, line 27: Do we want to demonstrate push/pull signatures from repository?

README.md
```diff
@@ -4,12 +4,31 @@ Notation is a project to add signatures as standard items in the registry ecosys
...
+  notation list $IMAGE
```
gokarnm, line 28: A comments explaining what this command does will help.

README.md
SteveLasker: This was the quick-start, which is an abbreviated version of: #78 
This is the challenge with having required parameters, for good security reasons, while enabling quick-starts.
What would you suggest to keep that balance?

README.md
SteveLasker: `notation sign` defaults to include push. There are separate apis for generating an offline signature, which we'd cover in a different walkthrough for offline/ephemeral client scenarios.

README.md
SteveLasker: Similar to above and #78 
How would you suggest the quick-start balance, compared to longer explanations?

Once we get automated builds, users could test this themselves, and call `--help` to get more details or see #78

README.md
gokarnm: The other docker commands are straightforward and probably well known to the reader. Where we introduce the notation commands, it would be nice to have a one line comments explaining what the specific command does.
```suggestion
  # Generate a test certificate, make it the default for signing, and add it as a trusted identity for verification
  notation cert generate-test --default --trust "wabbit-networks-dev"
```

README.md
gokarnm:
```suggestion
  # List the image, and any associated signatures
  notation list $IMAGE 
```

NiazFK (approved):
LGTM

PR #103, opened 2021-10-15
Remove prototype-2 links from readme

sajayantony:
Signed-off-by: Sajay Antony <sajaya@microsoft.com>


SteveLasker (approved):
With the notation cli merged to main, this is good hygiene to remove the older prototype 
LGTM

PR #122, opened 2021-11-01
No default push on local sign

### LINKED ISSUE #121: [Bug] Default push on local signing
The `--push` option is default to `true` when the `--local` option is toggled in the `notation sign` command.
However, the `--push` option should not be ignored if `--push-reference` is present.

Note: The `--pull` option is ignored when the `--local` option is toggled in the `notation verify` command.

shizhMSFT:
Fixes #121

### DIFF
--- a/cmd/notation/sign.go
+++ b/cmd/notation/sign.go
@@ -67,8 +67,7 @@ func runSign(ctx *cli.Context) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


SteveLasker (approved):
LGTM

SteveLasker (approved):
LGTM

PR #126, opened 2021-11-03
Support anonymous access

### LINKED ISSUE #124: [Bug] 401 Unauthorized on Public Registries
`notation sign` and `notation verify` does not work on public registries with `401 Unauthorized` response.

Example:
```
$ notation sign docker.io/library/hello-world@sha256:f54a58bc1aac5ea1a25d796ae155dc228b3f0e11d046ae276b39c4bf2f13d8c4
2021/11/03 06:58:13 https://registry-1.docker.io/v2/library/hello-world/manifests/sha256:f54a58bc1aac5ea1a25d796ae155dc228b3f0e11d046ae276b39c4bf2f13d8c4: 401 Unauthorized
```

### LINKED ISSUE #125: [Feature] Support identity token when fetching registry access token
When fetching the bearer token (registry access token), `notation` assumes the auth to the `realm` is basic auth.
However, it could also be a bearer auth.

To present the identity token with bearer auth instead of the basic auth, the users should leave `--username` blank and put the identity token in the `--password`.
```bash
unset NOTATION_USERNAME
export NOTATION_PASSWORD=$token
```

shizhMSFT:
Fixes #124

### DIFF
--- a/pkg/registry/auth.go
+++ b/pkg/registry/auth.go
@@ -35,13 +35,16 @@ func (t *authTransport) RoundTrip(originalReq *http.Request) (*http.Response, er
 	if err != nil {
 		return nil, err
 	}
-	if resp.StatusCode != http.StatusUnauthorized || t.username == "" {
+	if resp.StatusCode != http.StatusUnauthorized {
 		return resp, nil
 	}
 
 	scheme, params := parseAuthHeader(resp.Header.Get("Www-Authenticate"))
 	switch scheme {
 	case "basic":
+		if t.username == "" {
+			return resp, nil
+		}
 		resp.Body.Close()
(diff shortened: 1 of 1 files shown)


sajayantony:
Should we remove Basic Auth from notation?

shizhMSFT:
> Should we remove Basic Auth from notation?

As a security product, yes, we should remove basic auth. Could you create an issue for this?

shizhMSFT:
Howver, it might be OK to use basic auth since it is about availablity. If we reject basic auth, then users cannot even pull the image from lots of registries.

Besides, notation signatures are designed to be transferring in an insecure channel.

SteveLasker (approved):
LGTM

PR #141, opened 2021-11-24
ci/cd support: export some go routines for use in ci/cd

rchincha:
Useful for integration testing from within golang test files.

Signed-off-by: Ramkumar Chinchani <rchincha@cisco.com>


SteveLasker:
@gokarnm, @iamsamirzon, @shizhMSFT, what do you think? Would love to get tests going.

rchincha:
Without this, have to unfortunately either fork (and maintain that fork) or invoke `notation` as a binary like so:
https://github.com/project-zot/zot/blob/main/pkg/api/controller_test.go#L3181

SteveLasker:
@rchincha , that would not be our desire. this still shows as a draft/WIP. Are you ready to submit for merge?

rchincha:
Updated the PR, pls take a look if it is acceptable in current form or needs refactoring.

SteveLasker:
@shizhMSFT, @gokarnm, WDYT?
@rchincha, if you're ready for review, it would help to flip the [draft] bit to [ready for review].

SteveLasker:
Hi @rchincha, can you please resolve the merge conflicts and switch from draft so we can approve and merge?
CC @shizhMSFT

rchincha:
```
$ go test -v ./pkg/signature/...
=== RUN   TestSign
--- PASS: TestSign (0.28s)
PASS
ok      github.com/notaryproject/notation/pkg/signature (cached)
```

rchincha:
> Hi @rchincha, can you please resolve the merge conflicts and switch from draft so we can approve and merge? CC @shizhMSFT

Done. But would consider this an iteration to reorg code. Will add more unit tests to demonstrate ci/cd integration.

shizhMSFT:
In general, the final product of this `notation` repository is the `notation` binary. All users should invoke the `notation` binary for any usage, including testing. The reason behind it is also simple that _passing the tests built from packages_ does not imply _it will work with the final binary_.

Additionally, since [/pkg](https://github.com/golang-standards/project-layout/blob/master/pkg/README.md) is a path for **library code** , we need to think twice before putting anything into `/pkg`. BTW, library users should try `notation-go-lib` and `oras-go` first in order to avoid heavy dependencies introduced by `notation`.

shizhMSFT:
In general, the final product of this `notation` repository is the `notation` binary. All users should invoke the `notation` binary for any usage, including testing. The reason behind it is also simple that _passing the tests built from packages_ does not imply _it will work with the final binary_.

Additionally, since [/pkg](https://github.com/golang-standards/project-layout/blob/master/pkg/README.md) is a path for **library code** , we need to think twice before putting anything into `/pkg`. BTW, library users should try `notation-go-lib` and `oras-go` first in order to avoid heavy dependencies introduced by `notation`.

sajayantony:
Capturing comments from the discussion - the CLI is going to just become a thin shim over `notation-go` that would enable you to take dependency on notation-go directly and not expose pkg's from the CLI.

shizhMSFT:
> @shizhMSFT , what's your suggestion to @rchincha here? Should this PR be updated to reflect the refactoring? Or, should this move to notation-go?

Tests should be done against the binaries or the `notation-go` library.

dtzar:
It seems we should close this PR based on the discussion above...  @rchincha I see you filed an issue in the other repository.  Will this different approach per Shiwei's suggestion work for you?

PR #142, opened 2021-12-09
feat: add initial framework for notation extensibility

aramase:
Adds framework to:

- Add, remove, list kms plugin
- Add kms keys for signing and verification

This is a proof of concept based on the initial hackmd spec. For more details on the implementation refer to https://docs.google.com/document/d/1tDxprA7_H2tVJP0YZFznuTTqDwFP4qks5SaSMaOpS80/edit?usp=sharing.

### DIFF
--- a/cmd/notation/cert.go
+++ b/cmd/notation/cert.go
@@ -137,7 +137,7 @@ func listCerts(ctx *cli.Context) error {
 	}
 
 	// write out
-	printCertificateSet(cfg.VerificationCertificates.Certificates)
+	printCertificateSet(cfg.VerificationCertificates.Certificates, cfg.VerificationCertificates.KMSCerts)
 	return nil
 }
 
@@ -172,7 +172,7 @@ func removeCerts(ctx *cli.Context) error {
 	return nil
 }
 
-func printCertificateSet(s config.CertificateMap) {
+func printCertificateSet(s config.CertificateMap, k config.KMSProfileMap) {
 	maxNameSize := 0
 	for _, ref := range s {
 		if len(ref.Name) > maxNameSize {
@@ -183,6 +183,23 @@ func printCertificateSet(s config.CertificateMap) {
 	fmt.Printf(format, "NAME", "PATH")
 	for _, ref := range s {
 		fmt.Printf(format, ref.Name, ref.Path)
+	}
+
+	var maxKeyIDSize int
+	for _, ref := range k {
+		if len(ref.Name) > maxNameSize {
+			maxNameSize = len(ref.Name)
+		}
+		if len(ref.ID) > maxKeyIDSize {
+			maxKeyIDSize = len(ref.ID)
+		}
+	}
+	fmt.Println()
+	// iterate over KMS profiles
+	format = fmt.Sprintf("%%-%ds\t%%-%ds\t%%s\n", maxNameSize, maxKeyIDSize)
+	fmt.Printf(format, "NAME", "ID", "PLUGIN NAME")
+	for _, ref := range k {
+		fmt.Printf(format, ref.Name, ref.ID, ref.PluginName)
 	}
 }
 
--- a/cmd/notation/cert.go
+++ b/cmd/notation/cert.go
@@ -34,6 +34,20 @@ var (
 				Aliases: []string{"n"},
 				Usage:   "certificate name",
 			},
+			&cli.StringFlag{
+				Name:    "id",
+				Aliases: []string{"i"},
+				Usage:   "key id",
+			},
+			&cli.StringFlag{
+				Name:    "plugin",
+				Aliases: []string{"p"},
+				Usage:   "key plugin",
+			},
+			&cli.BoolFlag{
+				Name:  "kms",
+				Usage: "add key to KMS cert list",
+			},
 		},
 		Action: addCert,
 	}
@@ -86,6 +100,13 @@ var (
 )
 
 func addCert(ctx *cli.Context) error {
+	if ctx.Bool("kms") {
+		return addKMSCertToList(ctx)
+	}
+	return addCertToList(ctx)
+}
+
+func addCertToList(ctx *cli.Context) error {
 	// initialize
 	path := ctx.Args().First()
 	if path == "" {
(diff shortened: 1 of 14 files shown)


SteveLasker:
This addresses https://github.com/notaryproject/roadmap/issues/32
To get feedback, the proposal is to merge this in, get feedback on remote signing, and incorporate any feedback into the https://github.com/notaryproject/roadmap/issues/26 spec. 
Once feedback is captured, and incorporated into the spec, we'll update to match the spec.

SteveLasker (approved):
LGTM

pkg/signature/kms_jws.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	// marshal request
+	reqBytes, err := json.Marshal(req)
+	if err != nil {
+		return nil, err
+	}
+
+	// execute plugin
+	return p.executor.Execute(ctx, string(reqBytes))
```
gokarnm, line 76: The response json is not being parsed, seems the response being returned by plugin is raw signature envelope. 

https://hackmd.io/jhYbA4aNRFW7nmp7gPTIBA#Signature-Envelope-Generator

gokarnm (changes requested):

cmd/notation/cert.go
```diff
@@ -34,6 +34,20 @@ var (
...
+			},
+			&cli.StringFlag{
+				Name:    "plugin",
+				Aliases: []string{"p"},
+				Usage:   "key plugin",
+			},
+			&cli.BoolFlag{
+				Name:  "kms",
+				Usage: "add key to KMS cert list",
+			},
```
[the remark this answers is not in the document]
SteveLasker: The separate list was a topic of discussion. I've got mixed feelings, but we left it like this for now. It would be great for the spec to clarify what we want.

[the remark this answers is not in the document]
aramase: The current keys have `keyPath` and `certPath` properties which don't really apply to kms keys. The KMS keys only need `id` and `pluginName`.  IMO, it would be better to keep them distinct rather than building a common struct to encompass every single property across multiple key types and stash them in a single slice.

pkg/signature/kms_jws.go
SteveLasker: I'd suggest we merge as is, make progress on the updated spec, and do a round of updates as we learn more.

pkg/executor/executor.go
```diff
@@ -0,0 +1,49 @@
...
+	"context"
+	"fmt"
+	"os/exec"
+	"strings"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+type Interface interface {
```
shizhMSFT, line 14: According to [effective go](https://go.dev/doc/effective_go#interface-names), it is better to name it `Executor`.

pkg/executor/executor.go
```diff
@@ -0,0 +1,49 @@
...
+	Execute(ctx context.Context, argsf string, a ...interface{}) ([]byte, error)
+}
+
+type executor struct {
+	cmd     string
+	prepend []string
+}
+
+var _ Interface = &executor{}
```
shizhMSFT, line 23: nit: this line can be removed as it is checked at line 25.

pkg/executor/executor.go
```diff
@@ -0,0 +1,49 @@
...
+	return &executor{
+		cmd:     cmd,
+		prepend: prepend,
+	}
+}
+
+// Execute executes the command with the given arguments.
+func (e *executor) Execute(ctx context.Context, argsf string, a ...interface{}) ([]byte, error) {
+	substituted := fmt.Sprintf(argsf, a...)
+	args := append(e.prepend, strings.Split(substituted, " ")...)
```
shizhMSFT, line 35: I know this is a PoC. Could you create an issue to track this as it may cause issue if `a` contains spaces.

For instance, `"%s %s", "foo", "foo bar"` will become `["foo", "foo", "bar"]`, not `[ "foo", "foo bar"]`.

pkg/signature/kms_jws.go
shizhMSFT, line 76: This is for security purpose as `reqBytes` may contain format symbols.

```suggestion
	return p.executor.Execute(ctx, "%s", string(reqBytes))
```

cmd/notation/main.go
```diff
@@ -1,10 +1,10 @@
 package main
 
 import (
-	"log"
 	"os"
 
 	"github.com/notaryproject/notation/internal/version"
+	log "github.com/sirupsen/logrus"
```
shizhMSFT, line 7: Do we need to update the `go.mod` file?

shizhMSFT (approved):
Approved with suggestions.

pkg/executor/executor.go
aramase: Is there ever an instance where we have a single arg that's separated by space? I'll add a TODO to handle this scenario.

cmd/notation/main.go
aramase: Thanks, ran `go mod tidy`.

aramase:
Thanks for the review. @shizhMSFT! Addressed the comments in a new commit.

SteveLasker:
@gokarnm, @iamsamirzon are you good with merging this, as is?

cmd/notation/verify.go
```diff
@@ -21,11 +23,7 @@ var verifyCommand = &cli.Command{
 	ArgsUsage: "<reference>",
 	Flags: []cli.Flag{
 		flagSignature,
-		&cli.StringSliceFlag{
-			Name:    "cert",
-			Aliases: []string{"c"},
-			Usage:   "certificate names for verification",
-		},
+		flagCert,
```
priteshbandi, line 26: Why did we move this moved to `common.go`?

cmd/notation/key.go
```diff
@@ -38,6 +38,20 @@ var (
...
+			},
+			&cli.StringFlag{
+				Name:    "plugin",
+				Aliases: []string{"p"},
+				Usage:   "key plugin",
+			},
+			&cli.BoolFlag{
+				Name:  "kms",
+				Usage: "add key to KMS key list",
+			},
```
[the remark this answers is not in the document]
priteshbandi: +1 why do we really need this? Can we use presence and absence of `plugin` param to detect whether its kms plugin or not.

cmd/notation/cert.go
```diff
@@ -184,6 +247,23 @@ func printCertificateSet(s config.CertificateMap) {
...
+
+	var maxKeyIDSize int
+	for _, ref := range k {
+		if len(ref.Name) > maxNameSize {
+			maxNameSize = len(ref.Name)
+		}
+		if len(ref.ID) > maxKeyIDSize {
+			maxKeyIDSize = len(ref.ID)
+		}
+	}
```
priteshbandi, line 259: Why are we doing this? something peculiar to golang's `sprintf`(line 262)? Can we remove this piece of code?

cmd/notation/plugin.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+func addPlugin(ctx *cli.Context) error {
+	// initialize
+	args := ctx.Args()
+	switch args.Len() {
+	case 0:
+		return errors.New("missing plugin name and path")
+	case 1:
+		return errors.New("missing plugin path for the correspoding plugin name")
```
priteshbandi, line 62: What if user provides the name and not the path?

SteveLasker:
@aramase, per [notaryproject/issues/120](https://github.com/notaryproject/notaryproject/issues/120), can you change the PR to the [feat-kv-extensibility](https://github.com/notaryproject/notation/tree/feat-kv-extensibility) branch?

NiazFK:
LGTM

PR #175, opened 2022-05-12
Directory Structure Spec

### LINKED ISSUE #167: Spec : Notation directory structure
Spec for config location and dir structure. Should specify the directory structure for notation, paths for trust store, signing keys config, trust policy config, and plugins.

shizhMSFT:
Resolves #167

Signed-off-by: Shiwei Zhang <shizh@microsoft.com>

### DIFF
--- a/specs/directory.md
+++ b/specs/directory.md
@@ -49,6 +49,8 @@ Default directory paths for various operating systems at user level are specifie
 | `CONFIG`  | `$XDG_CONFIG_HOME` | `%AppData%`      | `~/Library/Application Support` |
 | `CACHE`   | `$XDG_CACHE_HOME`  | `%LocalAppData%` | `~/Library/Caches`              |
 
+On Unix, `$XDG_CONFIG_HOME` is default to `~/.config` and `$XDG_CACHE_HOME` is default to `~/.cache` if XDG environment variables are empty.
+
 There is no default `BIN` path at user level since the `notation` binary can be put anywhere as long as it in the `PATH` environment variable. Common directories on Unix/Darwin are `~/bin` and `~/.local/bin` where manual `PATH` update by users may be required.
 
 ## Structure
--- a/specs/directory.md
+++ b/specs/directory.md
@@ -1,0 +1,167 @@
+# Notation Directory Structure
+
+The `notation` CLI requires local file systems support for the following components across various platforms.
+
+- `notation` binary
+- Plugins
+- Configurations
+- Trust stores
+- Trust policies
+- Signature caches
+- Local key store
+
+This documentation specifies the recommended directory structure for those components.
+
+## Category
+
+The directories for various components are classified into the following catagories.
+
+| Alias     | Description                                                                         |
+| --------- | ----------------------------------------------------------------------------------- |
+| `BIN`     | Directory for executable binaries                                                   |
+| `LIBEXEC` | Directory for binaries not meant to be executed directly by users' shell or scripts |
+| `CONFIG`  | Directory for configurations                                                        |
+| `CACHE`   | Directory for user-specific cache                                                   |
+
+Although it is recommended to install `notation` with its plugins and default configurations at the system level, it is possible to install at the user level.
+
+On Unix systems, `notation` follows [Filesystem Hierarchy Standard][FHS] for system level directories and [XDG Base Directory Specification][XDG] for user level directories. On Windows, [Known Folders][KF] and [App Settings][AS] are followed equivalently. On Darwin, [macOS File System][macOS_FS] with [System Integrity Protection][SIP] is followed equivalently. If a file with the same name exists at the system level and the user level, the file at the user level takes over the priority.
+
+### System Level
+
+Default directory paths for various operating systems at system level are specified as below.
+
+| Directory | Unix           | Windows                       | Darwin                         |
+| --------- | -------------- | ----------------------------- | ------------------------------ |
+| `BIN`     | `/usr/bin`     | `%ProgramFiles%/notation/bin` | `/usr/local/bin`               |
+| `LIBEXEC` | `/usr/libexec` | `%ProgramFiles%`              | `/usr/local/lib`               |
+| `CONFIG`  | `/etc`         | `%ProgramData%`               | `/Library/Application Support` |
+
+`CACHE` is omitted since it is user specific.
+
+### User Level
+
+Default directory paths for various operating systems at user level are specified as below.
+
+| Directory | Unix               | Windows          | Darwin                          |
+| --------- | ------------------ | ---------------- | ------------------------------- |
+| `LIBEXEC` | `$XDG_CONFIG_HOME` | `%AppData%`      | `~/Library/Application Support` |
+| `CONFIG`  | `$XDG_CONFIG_HOME` | `%AppData%`      | `~/Library/Application Support` |
+| `CACHE`   | `$XDG_CACHE_HOME`  | `%LocalAppData%` | `~/Library/Caches`              |
+
+There is no default `BIN` path at user level since the `notation` binary can be put anywhere as long as it in the `PATH` environment variable. Common directories on Unix/Darwin are `~/bin` and `~/.local/bin` where manual `PATH` update by users may be required.
+
+## Structure
+
+The overall directory structure for `notation` is summarized as follows.
+
+```
+{BIN}
+└── notation
+{CONFIG}
+└── notation
+    ├── config.json
+    ├── private
+    │   ├── {key-name}.crt
+    │   └── {key-name}.key
+    └── trust
+        ├── policy.json
+        └── store
+            └── {trust-store-type}
+                └── {named-store}
+                    └── {cert-file}
+{CACHE}
+└── notation
+    └── signature
+        └── {manifest-digest-algorithm}
+            └── {manifest-digest}
+                └── {signature-digest-algorithm}
+                    └── {signature-digest}.sig
+{LIBEXEC}
+└── notation
+    └── plugins
+        └── {plugin-name}
+            └── notation-{plugin-name}
+```
+
+### Notation Binary
+
+The path for the `notation` binary is
+
+```
+{BIN}/notation
+```
+
+### Plugin
+
+Plugin are binaries not meant to be executed directly by users' shell or scripts. The path of a plugin follows the pattern below.
+
+```
+{LIBEXEC}/notation/plugins/{plugin-name}/notation-{plugin-name}
+```
+
+### General Configuration
+
+The path of the general configuration file of the `notation` CLI is
+
+```
+{CONFIG}/notation/config.json
+```
+
+### Trust Store
+
+The path of a certificate file in a [Trust Store][TS] follows the pattern of
+
+```
+{CONFIG}/notation/trust/store/{trust-store-type}/{named-store}/{cert-file}
+```
+
+### Trust Policy
+
+The path of the [Trust Policy][TP] file is
+
+```
+{CONFIG}/notation/trust/policy.json
+```
+
+### Signature Caches
+
+The signatures are cached to optimize the network traffic. The path of a cached signature for a certain manifest follows the pattern below.
+
+```
+{CACHE}/notation/signature/{manifest-digest-algorithm}/{manifest-digest}/{signature-digest-algorithm}/{signature-digest}.sig
+```
+
+or in a hierarchical view
+
+```
+{CACHE}
+└── notation
+    └── signature
+        └── {manifest-digest-algorithm}
+            └── {manifest-digest}
+                └── {signature-digest-algorithm}
+                    └── {signature-digest}.sig
+```
+
+### Local key store
+
+Developers sign artifacts using local private keys with associated certificate chain. The default directory structure for testing purpose is suggested as follows.
+
+```
+{CONFIG}/notation/private/{key-name}.crt
+{CONFIG}/notation/private/{key-name}.key
+```
+
+Developers SHOULD consider safer places to store the passphrase-protected key and certificate pair, or opt to remote signing.
+
+[References]::
+
+[FHS]: https://refspecs.linuxfoundation.org/fhs.shtml "Filesystem Hierarchy Standard"
+[XDG]: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html "XDG Base Directory Specification"
+[KF]: https://docs.microsoft.com/windows/win32/shell/knownfolderid "Known Folders"
+[AS]: https://docs.microsoft.com/windows/apps/design/app-settings/store-and-retrieve-app-data "App Settings"
+[macOS_FS]: https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW14 "macOS File System"
+[SIP]: https://support.apple.com/HT204899 "System Integrity Protection"
+[TS]: https://github.com/notaryproject/notaryproject/blob/main/trust-store-trust-policy-specification.md#trust-store "Trust Store"
+[TP]: https://github.com/notaryproject/notaryproject/blob/main/trust-store-trust-policy-specification.md#trust-policy "Trust Policy"
--- a/specs/directory.md
+++ b/specs/directory.md
@@ -194,15 +194,16 @@ Examples are shown on various platforms where the user `exampleuser` overrides t
 │       └── trust
 │           ├── policy.json
 │           └── store
-│               ├── ca
-│               │   ├── acme-rockets
-│               │   │   ├── cert1.pem
-│               │   │   └── cert2.pem
-│               │   └── wabbit-networks
-│               │       └── cert3.pem
-│               └── tsa
-│                   └── publicly-trusted-tsa
-│                       └── tsa-cert1.pem
+│               └── x509
+│                   ├── ca
+│                   │   ├── acme-rockets
+│                   │   │   ├── cert1.pem
+│                   │   │   └── cert2.pem
+│                   │   └── wabbit-networks
+│                   │       └── cert3.pem
+│                   └── tsa
+│                       └── publicly-trusted-tsa
+│                           └── tsa-cert1.pem
 ├── home
 │   └── exampleuser
 │       ├── .cache
@@ -231,12 +232,13 @@ Examples are shown on various platforms where the user `exampleuser` overrides t
 │               └── trust
 │                   ├── policy.json
 │                   └── store
-│                       ├── ca
-│                       │   └── acme-rockets
-│                       │       └── cert4.pem
-│                       └── tsa
-│                           └── publicly-trusted-tsa
-│                               └── tsa-cert2.pem
+│                       └── x509
+│                           ├── ca
+│                           │   └── acme-rockets
+│                           │       └── cert4.pem
+│                           └── tsa
+│                               └── publicly-trusted-tsa
+│                                   └── tsa-cert2.pem
 └── usr
     ├── bin
     │   └── notation
@@ -264,15 +266,16 @@ C:.
 │       └── trust
 │           ├── policy.json
 │           └── store
-│               ├── ca
-│               │   ├── acme-rockets
-│               │   │   ├── cert1.pem
-│               │   │   └── cert2.pem
-│               │   └── wabbit-networks
-│               │       └── cert3.pem
-│               └── tsa
-│                   └── publicly-trusted-tsa
-│                       └── tsa-cert1.pem
+│               └── x509
+│                   ├── ca
+│                   │   ├── acme-rockets
+│                   │   │   ├── cert1.pem
+│                   │   │   └── cert2.pem
+│                   │   └── wabbit-networks
+│                   │       └── cert3.pem
+│                   └── tsa
+│                       └── publicly-trusted-tsa
+│                           └── tsa-cert1.pem
 └── Users
     └── exampleuser
         └── AppData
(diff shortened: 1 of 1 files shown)


specs/directory.md
```diff
@@ -0,0 +1,167 @@
...
+The directories for various components are classified into the following catagories.
+
+| Alias     | Description                                                                         |
+| --------- | ----------------------------------------------------------------------------------- |
+| `BIN`     | Directory for executable binaries                                                   |
+| `LIBEXEC` | Directory for binaries not meant to be executed directly by users' shell or scripts |
+| `CONFIG`  | Directory for configurations                                                        |
+| `CACHE`   | Directory for user-specific cache                                                   |
+
+Although it is recommended to install `notation` with its plugins and default configurations at the system level, it is possible to install at the user level.
```
priteshbandi, line 26: Why we want user to install `notation` with its plugins and default configurations at the system level? Isnt user level more secure and isolated?

specs/directory.md
```diff
@@ -0,0 +1,167 @@
...
+└── notation
+    └── signature
+        └── {manifest-digest-algorithm}
+            └── {manifest-digest}
+                └── {signature-digest-algorithm}
+                    └── {signature-digest}.sig
+{LIBEXEC}
+└── notation
+    └── plugins
+        └── {plugin-name}
```
priteshbandi, line 83: Do you mean Fully qualified name? Two plugin can have same name.

specs/directory.md
```diff
@@ -0,0 +1,167 @@
...
+## Structure
+
+The overall directory structure for `notation` is summarized as follows.
+
+```
+{BIN}
+└── notation
+{CONFIG}
+└── notation
+    ├── config.json
```
gokarnm, line 63: We discussed about moving signing keys into `signingKeys.json` for remote keys using plugins, and including only global notation config settings like `unsecureRegistry` into `config.json`.

specs/directory.md
```diff
@@ -0,0 +1,167 @@
...
+On Unix systems, `notation` follows [Filesystem Hierarchy Standard][FHS] for system level directories and [XDG Base Directory Specification][XDG] for user level directories. On Windows, [Known Folders][KF] and [App Settings][AS] are followed equivalently. On Darwin, [macOS File System][macOS_FS] with [System Integrity Protection][SIP] is followed equivalently. If a file with the same name exists at the system level and the user level, the file at the user level takes over the priority.
+
+### System Level
+
+Default directory paths for various operating systems at system level are specified as below.
+
+| Directory | Unix           | Windows                       | Darwin                         |
+| --------- | -------------- | ----------------------------- | ------------------------------ |
+| `BIN`     | `/usr/bin`     | `%ProgramFiles%/notation/bin` | `/usr/local/bin`               |
+| `LIBEXEC` | `/usr/libexec` | `%ProgramFiles%`              | `/usr/local/lib`               |
```
shizhMSFT, line 37: I will add examples to demonstrate how the directories looks like on different platforms.

specs/directory.md
```diff
@@ -0,0 +1,167 @@
...
+
+The overall directory structure for `notation` is summarized as follows.
+
+```
+{BIN}
+└── notation
+{CONFIG}
+└── notation
+    ├── config.json
+    ├── private
```
gokarnm, line 64: Is this for notation generated local keys only? What about local keys referred by `notation sign $IMAGE --keyFile somekey --certFile somecert` which could be in any location, assuming they won't be copied over here.

specs/directory.md
shizhMSFT: `notation` may come with some default plugins or company-specific shared plugins.

For instance, `docker` has `docker-app` and `docker-buildx` bundled when `docker` CLI is installed.

shizhMSFT: Of course, plugins can be installed at the user level.

specs/directory.md
```diff
@@ -0,0 +1,167 @@
...
+        ├── policy.json
+        └── store
+            └── {trust-store-type}
+                └── {named-store}
+                    └── {cert-file}
+{CACHE}
+└── notation
+    └── signature
+        └── {manifest-digest-algorithm}
+            └── {manifest-digest}
```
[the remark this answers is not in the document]
shizhMSFT: No, we cannot because of the cross-platform support. On Windows, colon `:` in the path is not allowed.

specs/directory.md
shizhMSFT: Yes, please refer to the plugin spec for details. I will also have an example for that.

specs/directory.md
```diff
@@ -0,0 +1,167 @@
...
+                    └── {signature-digest}.sig
+```
+
+### Local key store
+
+Developers sign artifacts using local private keys with associated certificate chain. The default directory structure for testing purpose is suggested as follows.
+
+```
+{CONFIG}/notation/private/{key-name}.crt
+{CONFIG}/notation/private/{key-name}.key
```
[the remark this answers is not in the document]
shizhMSFT: or we can do the older way:

```
{CONFIG}/notation/certificates/{key-name}.crt
{CONFIG}/notation/keys/{key-name}.pem
```

specs/directory.md
shizhMSFT: Yes, exactly.

specs/directory.md
shizhMSFT: Added `signing/keys.json`.

specs/directory.md
shizhMSFT: Added examples

specs/directory.md
shizhMSFT: Added a link to the plugin spec.

specs/directory.md
```diff
@@ -0,0 +1,375 @@
...
+```
+{BIN}/notation
+```
+
+### Plugin
+
+[Plugins][Plugin] are binaries not meant to be executed directly by users' shell or scripts. The path of a plugin follows the pattern below.
+
+```
+{LIBEXEC}/notation/plugins/{plugin-name}/notation-{plugin-name}
```
qmuntal, line 103: Should we note that `windows` binaries must end with `.exe`? The extension is not part of `{plugin-name}`, so it can cause confusion for Windows developers reading this spec.

shizhMSFT: Good catch. I'll also update the Windows example.

SteveLasker:
I've added [Balancing Security and Usability](https://github.com/notaryproject/notaryproject/discussions/159) for reference

Based on the PR, it looks like we are securing the configuration and policy files to directories a normal user would be unable to configure. We should also assure the notation client doesn't allow the user to override the configuration files with non-admin directories or files. 
There may be some configurations that are less secure, and more usability, for instance reducing the number of parameters the user must enter every time. However, anything that implies security, MUST be placed in a directory requiring admin rights for write access.

specs/directory.md
```diff
@@ -0,0 +1,387 @@
...
+    │   │   └── {key-name}.crt
+    │   ├── keys
+    │   │   └── {key-name}.pem
+    │   └── keys.json
+    └── trust
+        ├── policy.json
+        └── store
+            └── {trust-store-type}
+                └── {named-store}
+                    └── {cert-file}
```
rgnote, line 82: I feel these paths are bit lengthy to traverse manually. Can we use shorter ones like below?
```
{CONFIG}/notation/signing-keys/{key-name}/cert.pem
{CONFIG}/notation/signing-keys/{key-name}/key.pem
{CONFIG}/notation/trust-store/{trust-store-type}/*
{CONFIG}/notation/trust-policy.json
```

I personally feel shorter paths are easier to use. Would like to hear others opinions. @gokarnm what do you think?

specs/directory.md
```diff
@@ -0,0 +1,387 @@
...
+
+On Unix systems, `notation` follows [Filesystem Hierarchy Standard][FHS] for system level directories and [XDG Base Directory Specification][XDG] for user level directories. On Windows, [Known Folders][KF] and [App Settings][AS] are followed equivalently. On Darwin, [macOS File System][macOS_FS] with [System Integrity Protection][SIP] is followed equivalently. If a file with the same name exists at the system level and the user level, the file at the user level takes over the priority.
+
+### System Level
+
+Default directory paths for various operating systems at system level are specified as below.
+
+| Directory | Unix           | Windows                       | Darwin                         |
+| --------- | -------------- | ----------------------------- | ------------------------------ |
+| `BIN`     | `/usr/bin`     | `%ProgramFiles%/notation/bin` | `/usr/local/bin`               |
```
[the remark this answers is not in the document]
gokarnm: On Windows %ProgramFiles% generally has sub directories for each application, with each dir containing all the files it needs (unlike say MacOS .app file). 
I don't think it's not typical for Windows apps to have `bin` directory though, `%ProgramFiles%/notation/notation` should suffice.

specs/directory.md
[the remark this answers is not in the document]
gokarnm: It's not clear by the directory names the purpose of keys and certificates. If these are local keys generated by notation, it's not clear why we need two directories, can you provide details/example of what the .crt and .pem contain (private keys and certificate chains).

specs/directory.md
```diff
@@ -0,0 +1,387 @@
...
+```
+{BIN}
+└── notation
+{CACHE}
+└── notation
+    └── signature
+        └── {manifest-digest-algorithm}
+            └── {manifest-digest}
+                └── {signature-digest-algorithm}
+                    └── {signature-digest}.sig
```
gokarnm, line 67: Is there a reason we need a sub directory for `{signature-digest-algorithm}`, a given manifest digest will only contain one signature. 
Is `{manifest-digest}` the artifact manifest digest or signature manifest digest, maybe update the dir structure to make that clearer.

specs/directory.md
```diff
@@ -0,0 +1,387 @@
...
+## Structure
+
+The overall directory structure for `notation` is summarized as follows.
+
+```
+{BIN}
+└── notation
+{CACHE}
+└── notation
+    └── signature
```
gokarnm, line 63: Rename to `notation/signatures` ?

specs/directory.md
```diff
@@ -0,0 +1,387 @@
...
+                └── {signature-digest-algorithm}
+                    └── {signature-digest}.sig
+```
+
+### Signing Key Store
+
+Developers sign artifacts using local private keys with associated certificate chain. The signing key information is tracked in a JSON file at
+
+```
+{CONFIG}/notation/signing/keys.json
```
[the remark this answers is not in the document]
gokarnm: [Plugin signing keys](https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md#using-a-plugin-for-signing) will go into keys.json, I'm not sure if we'll support local keys generated by notation too in the same file, @shizhMSFT ?

specs/directory.md
gokarnm: I don't have strong opinions, but current dir structure has some additional levels due to `notation\signing` and `notation\trust`. @shizhMSFT is there a benefit to having these sub directories, anything we loose by flattening the structure?

 Things to consider

1. Users will initially have to be aware of some paths the dir structure when we don't provide them CLI command (mainly trust policy and trust store), so we can keep paths short as long as they are clear and unambiguous.
2. Administrators who will deploy trust stores and policies will not use CLI commands, they will copy files in specific locations. 

I like the suggestion of combining the `signing/certificates` and `signing/keys` to `signing-keys/{key-name}`. My suggestions.

```
{CONFIG}/notation/signingKeys.json
{CONFIG}/notation/trustPolicy.json
{CONFIG}/notation/config.json
{CONFIG}/notation/signing-keys/{key-name}/cert.crt 
{CONFIG}/notation/signing-keys/{key-name}/key.pem 
{CONFIG}/notation/trust-store/{trust-store-type}/* 
```

specs/directory.md
```diff
@@ -0,0 +1,387 @@
...
+
+The path of the [Trust Policy][TP] file is
+
+```
+{CONFIG}/notation/trust/policy.json
+```
+
+### Signature Caches
+
+The signatures are cached to optimize the network traffic. The path of a cached signature for a certain manifest follows the pattern below.
```
gokarnm, line 144: Can you provide some more details here for my benefit, we can probably capture the details in another doc. How does caching work, given that additional signatures will be added over time. Will notation-go always call references API and skip downloading signature blobs it has already downloaded? We still have to download signature manifests every time, correct?

gokarnm (changes requested):

specs/directory.md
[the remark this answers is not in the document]
shizhMSFT: Usually, `.crt` is a certificate file or a certificate chain (a bundle of certificates). `.pem` can be a certificate or a key.

specs/directory.md
```diff
@@ -0,0 +1,387 @@
...
+
+`CACHE` is omitted since it is user specific.
+
+### User Level
+
+Default directory paths for various operating systems at user level are specified as below.
+
+| Directory | Unix               | Windows          | Darwin                          |
+| --------- | ------------------ | ---------------- | ------------------------------- |
+| `LIBEXEC` | `$XDG_CONFIG_HOME` | `%AppData%`      | `~/Library/Application Support` |
```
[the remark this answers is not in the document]
shizhMSFT: I'd like to follow the strategy of docker CLI that user plugins are stored under the config folder for easier management.

Reference: https://github.com/docker/cli/issues/1534

specs/directory.md
```diff
@@ -0,0 +1,387 @@
...
+| `LIBEXEC` | `/usr/libexec` | `%ProgramFiles%`              | `/usr/local/lib`               |
+| `CONFIG`  | `/etc`         | `%ProgramData%`               | `/Library/Application Support` |
+
+`CACHE` is omitted since it is user specific.
+
+### User Level
+
+Default directory paths for various operating systems at user level are specified as below.
+
+| Directory | Unix               | Windows          | Darwin                          |
```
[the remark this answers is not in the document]
shizhMSFT: Added default path for `XDG_*`.

specs/directory.md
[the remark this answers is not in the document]
shizhMSFT: It is actually common to have `bin` directory under the program folder. For instance,

```
C:\Users\<user_name>\AppData\Local\Programs\Microsoft VS Code\bin
C:\Program Files (x86)\Microsoft SDKs\Azure\CLI2\wbin
C:\Program Files\Go\bin
```

The advantage of having a `bin` folder is that it is safter to add the `bin` folder with executables only to the `PATH` environment variable.

specs/directory.md
```diff
@@ -0,0 +1,387 @@
...
+```
+{LIBEXEC}/notation/plugins/{plugin-name}/notation-{plugin-name}.exe
+```
+
+### General Configuration
+
+The path of the general configuration file of the `notation` CLI is
+
+```
+{CONFIG}/notation/config.json
```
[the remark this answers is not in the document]
shizhMSFT: This is out of the scope of this document. We should have a separate documentation for config.

specs/directory.md
```diff
@@ -0,0 +1,387 @@
...
+### Unix
+
+```
+/
+├── etc
+│   └── notation
+│       ├── config.json
+│       └── trust
+│           ├── policy.json
+│           └── store
```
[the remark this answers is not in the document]
shizhMSFT: Good catch. It should be `x509/ca` and `x509/tsa` according to [trust-store spec](https://github.com/notaryproject/notaryproject/blob/9cfd6a91b1dfad48886af17edf7cc8235556f550/trust-store-trust-policy-specification.md#trust-store).

specs/directory.md
```diff
@@ -0,0 +1,387 @@
...
+
+### Signing Key Store
+
+Developers sign artifacts using local private keys with associated certificate chain. The signing key information is tracked in a JSON file at
+
+```
+{CONFIG}/notation/signing/keys.json
+```
+
+Since the signing key store is user-specific, the system level `{CONFIG}` is not recommended. Developers SHOULD consider safe places to store the passphrase-protected key and certificate pairs, or opt to remote signing.
```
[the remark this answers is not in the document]
shizhMSFT: In the most case, it is not recommended. However, what if there are computer-level signing keys?

specs/directory.md
shizhMSFT: `trust/store` and `trust/policy` are better than `trust-store` and `trust-policy` as they are all under `trust` and are well organized. It also avoids naming convention. For example, `signing-keys` and `signingKeys` are not consistency in terms of naming.

shizhMSFT: > Administrators who will deploy trust stores and policies will not use CLI commands, they will copy files in specific locations.

They will deploy those using scripts, not drag-and-drop GUIs.

specs/directory.md
```diff
@@ -0,0 +1,387 @@
+# Notation Directory Structure
+
+The `notation` CLI requires local file systems support for the following components across various platforms.
+
+- `notation` binary
+- Plugins
+- Configurations
```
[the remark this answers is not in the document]
shizhMSFT: `Configurations` means `Global configurations` / `System configurations` and `User configurations`.

specs/directory.md
shizhMSFT: A manifest may have multiple signatures signed by various parties.

The `{manifest-digest}` is the digest of the subject manifest. i.e. The manifest to be signed.

shizhMSFT: The section `Signature Caches` clarifies that the manifest is the manifest to be signed.

specs/directory.md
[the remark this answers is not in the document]
rgnote: Should we call it `local-keys.json`?

specs/directory.md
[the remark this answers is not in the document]
rgnote: These keys are generated by notation, correct? If yes, where does notation generate these keys? user config or system config? If former, then how can keys exist in system config?

specs/directory.md
rgnote: > trust/store and trust/policy are better than trust-store and trust-policy as they are all under trust and are well organized. 

The path `{CONFIG}/notation/trust` is ambiguous. It exists on the file system but does not provide any real value. The flattened structure is better as each and every possible path is self-explanatory.

Now that I understand better about these files, I propose below structure
```
{CONFIG}/notation/signingkeys.json
{CONFIG}/notation/trustpolicy.json
{CONFIG}/notation/config.json
{CONFIG}/notation/local-keys/{key-name}/cert.pem
{CONFIG}/notation/local-keys/{key-name}/key.pem
{CONFIG}/notation/local-keys/{key-name}/chain.pem 
{CONFIG}/notation/trust-store/{trust-store-type}/* 
```

AFAIK camel case for multi-word config file names (trustPolicy.json, signingKeys.json) is not common (for example `.gitconfig` or `.gitignore`). I tried to look for any examples in docker but couldn't find any.

specs/directory.md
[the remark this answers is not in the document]
shizhMSFT: The only way for notation to generate keys is for test purpose, i.e. `notation cert generate-test`. The resulted keys are in the user config. For those keys in the system config, they are usually configured by the system admins.

specs/directory.md
shizhMSFT: The folder name `local-keys` seems making more sense than `signing-keys`.

BTW, should we remove all hyphens? so we have `truststore` and `localkeys`?

specs/directory.md
```diff
@@ -0,0 +1,395 @@
...
+
+or in a hierarchical view
+
+```
+{CACHE}
+└── notation
+    └── signatures
+        └── {manifest-digest-algorithm}
+            └── {manifest-digest}
+                └── {signature-digest-algorithm}
```
gokarnm, line 160: Should we rename `{signature-digest-algorithm}` to `{signature-blob-digest-algorithm}` to make it clearer, I (incorrectly) assumed this was related to signing algorithm hash earlier and that was confusing.

specs/directory.md
```diff
@@ -0,0 +1,395 @@
...
+
+The path of the [Trust Policy][TP] file is
+
+```
+{CONFIG}/notation/trust/policy.json
+```
+
+### Signature Caches
+
+The signatures are cached to optimize the network traffic. The path of a cached signature for a certain manifest follows the pattern below.
```
gokarnm, line 146:
```suggestion
The signatures are cached to optimize the network traffic. The path of a cached signatures for a certain target manifest (e.g. an image manifest) follows the pattern below.

```

specs/directory.md
```diff
@@ -0,0 +1,395 @@
...
+| --------- | ----------------------------------------------------------------------------------- |
+| `BIN`     | Directory for executable binaries                                                   |
+| `LIBEXEC` | Directory for binaries not meant to be executed directly by users' shell or scripts |
+| `CONFIG`  | Directory for configurations                                                        |
+| `CACHE`   | Directory for user-specific cache                                                   |
+
+Although it is recommended to install `notation` with its plugins and default configurations at the system level, it is possible to install at the user level.
+
+On Unix systems, `notation` follows [Filesystem Hierarchy Standard][FHS] for system level directories and [XDG Base Directory Specification][XDG] for user level directories. On Windows, [Known Folders][KF] and [App Settings][AS] are followed equivalently. On Darwin, [macOS File System][macOS_FS] with [System Integrity Protection][SIP] is followed equivalently. If a file with the same name exists at the system level and the user level, the file at the user level takes over the priority.
```
gokarnm, line 28: This makes sense from a usability perspective for certain components like config.json, but seems inverse of what is desired for trust store and trust policy.

gokarnm:
```suggestion
On Unix systems, `notation` follows [Filesystem Hierarchy Standard][FHS] for system level directories and [XDG Base Directory Specification][XDG] for user level directories. On Windows, [Known Folders][KF] and [App Settings][AS] are followed equivalently. On Darwin, [macOS File System][macOS_FS] with [System Integrity Protection][SIP] is followed equivalently.

**TODO** : If a file with the same name exists at the system level and the user level, the mechanism to resolve priority needs to be defined (tracked in #203).

```

gokarnm (approved):
Created an issue (#203) and added a suggestion to track precedence of system vs user files. Rest looks great!

PR #176, opened 2022-05-13
Add plugin sign capabaility

qmuntal:
This PR adds support for using external keys to sign images with the plugins.

### DIFF
--- a/internal/cmd/signer.go
+++ b/internal/cmd/signer.go
@@ -35,7 +35,6 @@ func GetSigner(ctx *cli.Context) (notation.Signer, error) {
 		return &jws.PluginSigner{
 			Runner:       runner,
 			KeyID:        key.ExternalKey.ID,
-			KeyName:      key.Name,
 			PluginConfig: key.PluginConfig,
 		}, nil
 	}
--- a/internal/cmd/signer.go
+++ b/internal/cmd/signer.go
@@ -27,9 +27,13 @@ func GetSigner(ctx *cli.Context) (notation.Signer, error) {
 		return signature.NewSignerFromFiles(key.X509KeyPair.KeyPath, key.X509KeyPair.CertificatePath)
 	}
 	if key.ExternalKey != nil {
+		mgr := manager.New(config.PluginDirPath)
+		runner, err := mgr.Runner(key.PluginName)
+		if err != nil {
+			return nil, err
+		}
 		return &jws.PluginSigner{
-			Runner:       manager.NewManager(),
-			PluginName:   key.PluginName,
+			Runner:       runner,
 			KeyID:        key.ExternalKey.ID,
 			KeyName:      key.Name,
 			PluginConfig: key.PluginConfig,
--- a/internal/cmd/signer.go
+++ b/internal/cmd/signer.go
@@ -5,7 +5,8 @@ import (
 	"time"
 
 	"github.com/notaryproject/notation-go"
-	"github.com/notaryproject/notation-go/crypto/timestamp"
+	"github.com/notaryproject/notation-go/plugin/manager"
+	"github.com/notaryproject/notation-go/signature/jws"
 	"github.com/notaryproject/notation/pkg/config"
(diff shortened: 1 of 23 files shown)


internal/cmd/signer.go
```diff
@@ -14,33 +15,30 @@ import (
 // GetSigner returns a signer according to the CLI context.
 func GetSigner(ctx *cli.Context) (notation.Signer, error) {
 	// read paths of the signing key and its corresponding cert.
```
gokarnm, line 17:
```suggestion
	// Construct a signer from key and cert file if provided as CLI arguments
```

internal/cmd/signer.go
```diff
@@ -14,33 +15,30 @@ import (
...
-		keyPath = key.KeyPath
-		certPath = key.CertificatePath
+	if keyPath := ctx.String(FlagKeyFile.Name); keyPath != "" {
+		certPath := ctx.String(FlagCertFile.Name)
+		return signature.NewSignerFromFiles(keyPath, certPath)
 	}
-
-	// construct signer
-	signer, err := signature.NewSignerFromFiles(keyPath, certPath)
+	key, err := config.ResolveKey(ctx.String(FlagKey.Name))
```
gokarnm, line 22:
```suggestion
    // Construct a signer from preconfigured key pair in config.json
    // if key name is provided as the CLI argument
	key, err := config.ResolveKey(ctx.String(FlagKey.Name))
```

internal/cmd/signer.go
```diff
@@ -14,33 +15,30 @@ import (
...
 	if err != nil {
 		return nil, err
 	}
-	if endpoint := ctx.String(FlagTimestamp.Name); endpoint != "" {
-		if signer.TSA, err = timestamp.NewHTTPTimestamper(nil, endpoint); err != nil {
+	if key.X509KeyPair != nil {
+		return signature.NewSignerFromFiles(key.X509KeyPair.KeyPath, key.X509KeyPair.CertificatePath)
+	}
+	if key.ExternalKey != nil {
```
gokarnm, line 29:
```suggestion
    // Construct a plugin signer if key name provided as the CLI argument
    // corresponds to an external key
	if key.ExternalKey != nil {
```

internal/cmd/signer.go
```diff
@@ -14,33 +15,30 @@ import (
...
 			return nil, err
 		}
+		return &jws.PluginSigner{
+			Runner:       runner,
+			KeyID:        key.ExternalKey.ID,
+			PluginConfig: key.PluginConfig,
+		}, nil
 	}
-	return signer, nil
+	return nil, errors.New("unsupported key")
```
gokarnm, line 41:
```suggestion
	return nil, errors.New("unsupported key, either provide a local key and certificate file paths, or a key name in config.json, check [DOC_PLACEHOLDER] for details")
```

pkg/signature/jws.go
```diff
@@ -24,6 +24,9 @@ func NewSignerFromFiles(keyPath, certPath string) (*jws.Signer, error) {
 	if err != nil {
 		return nil, err
 	}
+	if len(cert.Certificate) == 0 {
+		return nil, errors.New("missing signer certificate chain")
```
gokarnm, line 28:
```suggestion
		return nil, fmt.Errorf("%q does not contain a signer certificate chain", certPath)
```

gokarnm: Ideally we should push line 22 onward in this method into notation-go, any cert parse/validate should be done in lower layers.

gokarnm (changes requested):

pkg/signature/jws.go
qmuntal: Agree, but I would leave this change until https://github.com/notaryproject/notation-go/issues/36 is implemented, else we will be moving functions up and down without a clear target.

gokarnm (approved):

gokarnm (approved):
LGTM!

SteveLasker (approved):
LGTM

PR #181, opened 2022-05-26
Fix pluginConfig parsing

qmuntal:
`--pluginConfig` flag is currently not being split by `,` when there are multiple entries.

This PR fixes the parsing of the `--pluginConfig` flag to correctly account for multiple entries. 

Signed-off-by: qmuntal <qmuntaldiaz@microsoft.com>

### DIFF
--- a/internal/cmd/flags.go
+++ b/internal/cmd/flags.go
@@ -45,17 +45,18 @@ var (
 		Usage:   "original reference",
 	}
 
-	FlagPluginConfig = &cli.StringSliceFlag{
+	FlagPluginConfig = &cli.StringFlag{
 		Name:    "pluginConfig",
 		Aliases: []string{"pc"},
 		Usage:   "list of comma-separated {key}={value} pairs that are passed as is to the plugin, refer plugin documentation to set appropriate values",
 	}
 )
 
-func ParseFlagPluginConfig(pluginConfigSlice []string) (map[string]string, error) {
-	if len(pluginConfigSlice) == 0 {
+func ParseFlagPluginConfig(v string) (map[string]string, error) {
+	if v == "" {
 		return nil, nil
 	}
+	pluginConfigSlice := strings.Split(v, ",")
 	m := make(map[string]string, len(pluginConfigSlice))
 	for _, c := range pluginConfigSlice {
 		if k, v, ok := strings.Cut(c, "="); ok {
--- a/cmd/notation/key.go
+++ b/cmd/notation/key.go
@@ -135,7 +135,7 @@ func addExternalKey(ctx *cli.Context, pluginName, keyName string) (config.KeySui
 	if p.Err != nil {
 		return config.KeySuite{}, fmt.Errorf("invalid plugin: %w", p.Err)
 	}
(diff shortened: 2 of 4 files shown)


dtzar (approved):

internal/cmd/flags.go
```diff
@@ -45,17 +45,18 @@ var (
...
 	}
 )
 
-func ParseFlagPluginConfig(pluginConfigSlice []string) (map[string]string, error) {
-	if len(pluginConfigSlice) == 0 {
+func ParseFlagPluginConfig(v string) (map[string]string, error) {
+	if v == "" {
 		return nil, nil
 	}
+	pluginConfigSlice := strings.Split(v, ",")
```
gokarnm, line 59: If we directly split the string we won't handle cases where the plugin param value itself contains a delimiter, correct?

`--pluginConfig a=b, c="value1,value2"`

qmuntal: Good shoot. Fixed in https://github.com/notaryproject/notation/pull/181/commits/e173eb55240d045caa62acdfa0c57f3befc002c6.

internal/cmd/flags.go
```diff
@@ -45,17 +46,21 @@ var (
...
+	if v == "" {
 		return nil, nil
 	}
+	pluginConfigSlice, err := splitQuoted(v)
+	if err != nil {
+		return nil, err
+	}
 	m := make(map[string]string, len(pluginConfigSlice))
 	for _, c := range pluginConfigSlice {
 		if k, v, ok := strings.Cut(c, "="); ok {
```
rgnote, line 66: Here, we can check for empty key/value and throw an error.

rgnote: Can we also trim the leading and trailing white spaces?

qmuntal: Both done

internal/cmd/flags.go
```diff
@@ -45,27 +46,102 @@ var (
...
-	for _, c := range pluginConfigSlice {
-		if k, v, ok := strings.Cut(c, "="); ok {
+	flags, err := splitQuoted(val)
+	if err != nil {
+		return nil, err
+	}
+	m := make(map[string]string, len(flags))
+	for _, c := range flags {
+		if c == "" {
+			return nil, fmt.Errorf("empty entry: %q", c)
```
rgnote, line 82: This doesn't error out `a=b,=`
Let's check the key and value are not empty after space trimming them.

qmuntal: Done!

internal/cmd/flags.go
```diff
@@ -45,27 +46,106 @@ var (
...
+		return nil, err
+	}
+	m := make(map[string]string, len(flags))
+	for _, c := range flags {
+		c := strings.TrimSpace(c)
+		if c == "" {
+			return nil, fmt.Errorf("empty entry: %q", c)
+		}
 		if k, v, ok := strings.Cut(c, "="); ok {
+			if k == "" || v == "" {
```
rgnote, line 86:
```suggestion
			k := strings.TrimSpace(k)
			v := strings.TrimSpace(v)
			if k == "" || v == "" {
```

rgnote (changes requested):
Rest looks good

rgnote (approved):
LGTM

PR #210, opened 2022-06-28
Add Credential File to Auth Spec

### LINKED ISSUE #206: Define local credential experience for Registry Authentication
Issue to track registry authentication experience in the absence of docker compatible credential provider in #192 . The proposed storage of unencrypted registry credential on local file system allows an attacker to potentially get read/write access to the registry.

Other options to consider, which arguably do not provide better security but gives the responsibility of securing the credentials to the end user, rather than in built support of unencrypted credentials in `notation`.
1. In the absence of a configured credential provider, require the user to provide user name and password as CLI arguments on every command that interacts with the registry (e.g. `notation sign` , `notation verify`)
2. Notation provides a minimal credential provider that takes input from environment variables (one option) as a script, end user can modify the script to pick credentials from file, or any other location they prefer.

I prefer option 1 for RC1, with addition of option 2 post RC1.

@SteveLasker , @sajayantony , @michaelb990 would be great to get your feedback.

Original content from PR #192
---

### Credential File

The credential file is alternative credential store when credential helpers are not available. The default file path is

```
{CONFIG}/notation/auth.json
```

The credential file path can be altered by setting the `credsFile` field of the notation config.

```json
{
    "credsFile": "/absolute/path/to/auth.json"
}
```

Since credentials are stored in plaintext, the permission of the credential file MUST be kept minimum when storing credentials. On Unix / Linux, the permission MUST be either `0600` (default) or `0400` (read-only).

shizhMSFT:
Resolves #206

Signed-off-by: Shiwei Zhang <shizh@microsoft.com>

### DIFF
--- a/specs/registry-auth.md
+++ b/specs/registry-auth.md
@@ -118,7 +118,21 @@ The credential store can be specified globally or per registry by setting the no
 
 ### Credential File
 
-TODO: Define local credential experience without credential provider support (#206).
+The credential file is alternative credential store when credential helpers are not available. The default file path is
+
+```
+{CONFIG}/notation/auth.json
+```
+
+The credential file path can be altered by setting the `credsFile` field of the notation config.
+
+```json
+{
+    "credsFile": "/absolute/path/to/auth.json"
+}
+```
+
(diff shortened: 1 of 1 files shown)


shizhMSFT:
Closing as per https://github.com/notaryproject/notation/issues/206#issuecomment-1168096534

Issue #119, opened 2021-10-27
[Feature] notation login 


SteveLasker:
The current notation cli supports `-u / -p` for each command. However, having to provide the username/password for each command is tedious at best.
Alternatively, cli commands provide a `notation login -u -p` type of experience where the subsequent commands can forgo having to provide the commands each time.

Ideally, this would be provided by the `oras-go` library, and notation just benefits from that common experience.

- [ ] auth should support encrypted storage or usernames and passwords, possibly following the docker encryption story
- [ ] not be dependent on the docker client or libraries being installed
- [ ] support basic auth flows that use cloud provider tokens for native authentication
- [ ] support 2fa cloud provider capabilities

SteveLasker:
@shizhMSFT will you be able to integrate this into RC1?
Instead of having to set `NOTATION_USERNAME` and `NOTATION_PASSWORD`, we can pickup the docker cred store.

shizhMSFT:
This functionality is currently not present in `oras-go`. Therefore, I would deter it to post-RC1.

shizhMSFT:
Login functionalities are currently not present in `oras-go`. Let me estimate if it fits the RC1 plan.

gokarnm:
@shizhMSFT were you able to look into #170, basic auth with `notation sign $IMAGE -u <USERNAME> -p <PASSWORD>` seems to be broken, will doing a new release with ORAS integration changes fix this issue?

shizhMSFT:
Yes, a new release of `notation` should fix it.

gokarnm:
Just clarifying, a new release of `notation` alpha, without any code changes should fix this issue? Some ORAS integration work was done in Jan (#150).

shizhMSFT:
@binbin-li Could you ack this item as you are working on this?

binbin-li:
acked

dtzar:
Other than 2fa - it seems this work is complete and we can close this item out?  @binbin-li @SteveLasker 
IMO I don't think 2fa is important for now and should be tracked in a separate issue.

SteveLasker:
If someone has tried the experience, as opposed to reviewing PRs, I'm happy to close. 
The 2fa was related to cloud provider specific login scenarios. For instance, you can use `az acr login` to set the docker credentials for registry operations.
I believe aws has a similar thing, but implemented through a docker plug-in.

[...]

PR #223, opened 2022-07-06
notation login CLI

SteveLasker:
Updates for #119 
Signed-off-by: Steve Lasker <stevenlasker@hotmail.com>

### DIFF
--- a/specs/notation-cli.md
+++ b/specs/notation-cli.md
@@ -4,16 +4,18 @@ The following spec outlines the notation CLI.
 The CLI commands are what's currently available in [notation v0.7.1-alpha.1](https://github.com/notaryproject/notation/releases/tag/v0.7.1-alpha.1). The CLI experience in alpha.1 does not represent the final user experience, and CLI commands may have breaking changes before RC release as the CLI experience is finalized.
 
 ## Table of Contents
-- [notation](#notation)
-- [sign](#sign): Signs artifacts
-- [verify](#verify): Verifies OCI Artifacts
-- [push](#push): Push signature to remote
-- [pull](#pull): Pull signatures from remote
-- [list](#list): List signatures from remote
+
+- [notation](#notation): command group for signing and verification operations
+- [cache](#cache): Manage signature cache
 - [certificate](#certificate): Manage certificates used for verification
 - [key](#key): Manage keys used for signing
-- [cache](#cache): Manage signature cache
+- [list](#list): List signatures from remote
+- [login](#login): Provide credentials for authenticated registry operations
 - [plugin](#plugin): Manage KMS plugins
+- [pull](#pull): Pull signatures from remote
+- [push](#push): Push signature to remote
+- [sign](#sign): Signs artifacts
+- [verify](#verify): Verifies OCI Artifacts
 
 ## notation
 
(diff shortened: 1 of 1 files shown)


specs/notation-cli.md
```diff
@@ -127,6 +129,32 @@ OPTIONS:
...
+notation login --help
+NAME:
+   notation login - Provides credentials for authenticated registry operations
+
+USAGE:
+   notation login [options] [server]
+
+OPTIONS:
+   --username value, -u value    Username for registry operations
+   --password value, -p value    Password for registry operations
```
shizhMSFT, line 144: How about environment variables associated with `-u` and `-p` like `NOTATION_USERNAME` and `NOTATION_PASSWORD`?

shizhMSFT (approved):
LGTM with suggestions.

specs/notation-cli.md
```diff
@@ -127,6 +129,32 @@ OPTIONS:
    --help, -h                  show help (default: false)
 ```
 
+## login
+
+```console
+notation login --help
+NAME:
+   notation login - Provides credentials for authenticated registry operations
```
gokarnm, line 137:
```suggestion
   notation login - Login to a registry with credentials for authenticated registry operations
```

specs/notation-cli.md
```diff
@@ -36,6 +37,7 @@ COMMANDS:
    certificate, cert  Manage certificates used for verification
    key                Manage keys used for signing
    list, ls           List signatures from remote
+   login              Provide credentials for authenticated registry operations
```
gokarnm, line 40:
```suggestion
   login              Login to a registry with credentials for authenticated registry operations
```

specs/notation-cli.md
gokarnm:
```suggestion
   --username value, -u value    Username for registry authentication
   --password value, -p value    Password for registry authentication
```

gokarnm: Should we also support `--password-stdin` like [docker cli](https://docs.docker.com/engine/reference/commandline/login/#provide-a-password-using-stdin) does?

specs/notation-cli.md
```diff
@@ -127,6 +129,32 @@ OPTIONS:
...
+OPTIONS:
+   --username value, -u value    Username for registry operations
+   --password value, -p value    Password for registry operations
+   --help, -h                    Show help (default: false)
+
+POSITIONAL
+  <server>                       The registry URL for authentication
+
+GLOBAL ARGUMENTS
+   --plain-http                  Registry access via plain HTTP (default: false)
```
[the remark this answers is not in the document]
gokarnm: @shizhMSFT, what is `--insecure` used for, and how common is the usage for `--ca-file` (I assume provides custom TLS trust store).

gokarnm (approved):
LGTM with suggestions.

specs/notation-cli.md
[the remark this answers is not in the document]
shizhMSFT: `--insecure` is to skip the TLS cert check. Since `notation` is security oriented, we should not provide this functionality.

`--ca-file` is to specify the root CA cert to registry access. See https://github.com/oras-project/oras/issues/217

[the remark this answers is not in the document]
SteveLasker: @shizhMSFT, I was thinking this would be done with the `insecure registries` option in config.  
Does that still exist? 
Seems calling `--insecure` is too tedious. For the scenarios/instances where https isn't supported, such as IOT environments, I just assumed the config would solve the problem and not something that's done ad-hoc requiring the cli to support it.

specs/notation-cli.md
SteveLasker: Saying login, where the parameter says login feels redundant. I was just trying to find a synonym verb.

SteveLasker:
I've updated the PR to:
- add $NOTATION_USERNAME, $NOTATION_PASSWORD on `notation login`
- Removed -u -p from other commands, streamlinging so users can benefit from existing cred stores, not having to provide -u -p on every command. This is similar to `docker push|pull` where the user either does `docker login`, or benefits from a previously configured store.

PR #235, opened 2022-07-08
Added conversion of DER to PEM for verify

### LINKED ISSUE #157: add support for binary-encoded x509 certs
`notation verify` only supports validating with ASCII-encoded x509 certs (PEM)

Today, you can convert a binary encoded DER to PEM and verify using PEMs:
`openssl x509 -in mycert.crt -inform DER -out mycert.pem`

However, users shouldn't need to do this conversion manually and notation should add support for binary-encoded x509 certs

patrickzheng200:

Resolves #157

### DIFF
--- a/pkg/signature/jws.go
+++ b/pkg/signature/jws.go
@@ -3,6 +3,7 @@ package signature
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


pkg/signature/jws.go
```diff
@@ -52,7 +53,10 @@ func NewVerifierFromFiles(certPaths []string) (*jws.Verifier, error) {
 			return nil, err
 		}
 		if !verifier.VerifyOptions.Roots.AppendCertsFromPEM(data) {
-			return nil, fmt.Errorf("failed to parse PEM certificate: %q", path)
+			certPEM := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: data})
```
binbin-li, line 56: Probably we can add a comment to explain that we're trying to convert DER to PEM? But it looks good to me overall.

patrickzheng200: > Probably we can add a comment to explain that we're trying to convert DER to PEM? But it looks good to me overall.

Sure. Will add a comment to it.

JeyJeyGao: Did you test it with some corner cases? like empty file, wrong cert file.

patrickzheng200: > Did you test it with some corner cases? like empty file, wrong cert file.

Yes, both scenarios are covered. Also tested end to end with the example net-monitor image.

SteveLasker:
@patrickzheng200, was this accidently closed, vs. merged?

patrickzheng200:
> @patrickzheng200, was this accidentally closed, vs. merged?

I noticed that pkg/signature/jws.go file has been removed from notation. Do we still need this PR?

patrickzheng200:
We are supporting notation verify with binary encoded DER in [notation-core-go](https://github.com/notaryproject/notation-core-go/blob/1d963fd57141fedb192a064867fb3114e43220cb/x509/cert.go#L24).

PR #249, opened 2022-07-15
Use cobra cli

### LINKED ISSUE #183: Command flags after arguments are discarded without warning
Several folks and I have been hit by this issue:

- `notation sign --key=$KEY_NAME $IMAGE ` works
- `notation sign $IMAGE --key=$KEY_NAME` does not works and errors out with `signing key not found`.

The same problem applies to any non-required flag which is used together with an argument, i.e. `-p/-u`. 

Using flags after arguments seems not to be supported by `urfave/cli`, see https://github.com/urfave/cli/issues/976.

We should at least error early with a meaningful message instead of executing the normal workflow and having a random error message.

@dtzar

chloeyin:

### Fix:
Fix #183 by using `github.com/spf13/cobra` to replace `github.com/urfave/cli/v2`, which is a cli package also used by [oras-cli](https://github.com/oras-project/oras).

### Compatibility:
According to the [cli spec](https://github.com/notaryproject/notation/blob/main/specs/notation-cli.md), there are two incompatible changes which are not well-supported by cobra:

1. It doesn't support multiple aliases, which will affect the user experience of `notation push` and `notation pull`.
Before this pr, when running `notation push`, you can use both `-s` or `-f` as an alias for `--signature`, as specified in [spec](https://github.com/notaryproject/notation/blob/main/specs/notation-cli.md#push).
Now, we don't have a `-f` flag as an alias. If running `notation push --help`, it will print the message below, which only supports `-s` as an alias for `--signature`:
```
notation push --help
Push signature to remote

Usage:
  notation push [reference] [flags]

Flags:
  -h, --help                help for push
  -p, --password string     password for generic remote access
      --plain-http          remote access via plain HTTP
  -s, --signature strings   signature files
  -u, --username string     username for generic remote access
```

2. It doesn't support string alias, which will affect the user experience of `notation key add`.
Before this pr, when running `notation key add`, we can use `-pc` as an alias for `--pluginConfig`:
```
notation key add --help
NAME:
   notation key add - Add key to signing key list

USAGE:
   notation key add [command options] [<key_path> <cert_path>]

OPTIONS:
   --default, -d                     mark as default (default: false)
   --id value                        key id (required if --plugin is set)
   --name value, -n value            key name
   --plugin value, -p value          signing plugin name
   --pluginConfig value, --pc value  list of comma-separated {key}={value} pairs that are passed as is to the plugin, refer plugin documentation to set appropriate values
```
Now we use a single char instead of a string as an alias. If running `notation key add --help`, it will print the message below, which use `-c` as an alias for `--pluginConfig` instead of `-pc`:
```
notation key add --help
Add key to signing key list

Usage:
  notation key add [key_path cert_path] [flags]

Flags:
  -d, --default               mark as default
  -h, --help                  help for add
      --id string             key id (required if --plugin is set)
  -n, --name string           key name
  -p, --plugin string         signing plugin name
  -c, --pluginConfig string   list of comma-separated {key}={value} pairs that are passed as is to the plugin, refer plugin documentation to set appropriate values
```


binbin-li:
40 files seem a bit huge to me. I feel you could split it into 3 PRs, one for each cli set: `notation`, `docker-notation` and `docker-generate`.

PR #250, opened 2022-07-15
Use cobra cli for docker-generate command

chloeyin:

### What
Fix https://github.com/notaryproject/notation/issues/183 by using `github.com/spf13/cobra` to replace `github.com/urfave/cli/v2`, which is a cli package also used by [oras-cli](https://github.com/oras-project/oras).

### Notice
The whole PR will update about 50 files, which is too difficult to review. I will split them into 4 PRs.
I expect these 4 PRs will do different jobs:
1. The first PR will update the `docker-generate` command to use `cobra` and add some common flags which`cobra` will use in later PRs.
2. The second PR will update the `docker-notation` command to use `cobra`.
3. The third PR will update the `notation` command to use `cobra`.
4. The forth PR will do some clean up jobs, including remoing unused functions and variables, etc.

This is the first PR of the series.


internal/cmd/flags.go
```diff
@@ -53,6 +56,70 @@ var (
...
+		Usage:     "original reference",
+	}
+	SetPflagReference = func(cmd *cobra.Command) {
+		cmd.Flags().StringP(PflagReference.Name, PflagReference.Shorthand, "", PflagReference.Usage)
+	}
+
+	// TODO: cobra does not support string shorthand
+	PflagPluginConfig = &pflag.Flag{
+		Name:      "pluginConfig",
+		Shorthand: "c",
```
binbin-li, line 115: Can you check with @SteveLasker that "c" is fine in this case?

cmd/docker-generate/manifest_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+func TestGenerateManifestCmd(t *testing.T) {
+	tests := []struct {
+		expectedOutput    string
+		expectedReference string
+		args              []string
+		expectedErr       bool
+	}{
+		{
+			expectedOutput:    "abc",
```
binbin-li, line 15: Probably we can have some more meaningful strings defined instead of "abc", "def".

cmd/docker-generate/metadata.go
```diff
@@ -5,9 +5,21 @@ import (
 	"os"
 
 	"github.com/notaryproject/notation/internal/docker"
-	"github.com/urfave/cli/v2"
+	"github.com/spf13/cobra"
 )
 
+func metadataCommand() *cobra.Command {
+	cmd := &cobra.Command{
```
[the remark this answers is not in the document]
chloeyin: Yes, I will update the code.

internal/cmd/flags.go
chloeyin: Ok, I will check it.

cmd/docker-generate/metadata.go
```diff
@@ -5,9 +5,20 @@ import (
 	"os"
 
 	"github.com/notaryproject/notation/internal/docker"
-	"github.com/urfave/cli/v2"
+	"github.com/spf13/cobra"
 )
 
+func metadataCommand() *cobra.Command {
```
patrickzheng200, line 11: Same here. We should generate a unit test for this function as well.

cmd/docker-generate/generate.go
```diff
@@ -1,12 +1,13 @@
...
-	"github.com/urfave/cli/v2"
+	"github.com/spf13/cobra"
 )
 
-var generateCommand = &cli.Command{
-	Name: "generate",
-	Subcommands: []*cli.Command{
-		manifestCommand,
-	},
+func generateCommand() *cobra.Command {
```
patrickzheng200, line 7: This func does not have a unit test yet. (Under [Github Actions](https://github.com/notaryproject/notation/actions/runs/2677789762), there's the notation-coverage-report showing coverage percentage of the current build.)

patrickzheng200 (changes requested):

cmd/docker-generate/manifest.go
```diff
@@ -6,26 +6,24 @@ import (
...
+	}
+	cmd.Flags().StringP("output", "o", "", "write to a file instead of stdout")
+	return cmd
 }
 
-func generateManifest(ctx *cli.Context) error {
+func generateManifest(cmd *cobra.Command) error {
 	var reader io.Reader
-	if reference := ctx.Args().First(); reference != "" {
+	if reference := cmd.Flags().Arg(0); reference != "" {
```
qweeah, line 26: We should use [cobra arguments](https://cobra.dev/#positional-and-custom-arguments) intead, which is more flexible. You can add a struct type to save all the flags and arguments.

cmd/docker-generate/manifest.go
```diff
@@ -6,26 +6,24 @@ import (
...
-	Usage:     "generates the manifest of a docker image",
-	ArgsUsage: "[<reference>]",
-	Flags: []cli.Flag{
-		&cli.StringFlag{
-			Name:    "output",
-			Aliases: []string{"o"},
-			Usage:   "write to a file instead of stdout",
+func generateManifestCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "manifest [reference]",
```
qweeah, line 14: We can use `Args` field to impose sanctity check to the arguments.

cmd/docker-generate/manifest_test.go
```diff
@@ -0,0 +1,69 @@
+package main
```
qweeah, line 1: This is new to me and sounds like a good way to do unit testing on flag/argument parsing behavior. @shizhMSFT  Do we need to apply this to oras-project/oras?

internal/cmd/flags.go
```diff
@@ -53,6 +56,70 @@ var (
...
+	PflagReference = &pflag.Flag{
+		Name:      "reference",
+		Shorthand: "r",
+		Usage:     "original reference",
+	}
+	SetPflagReference = func(cmd *cobra.Command) {
+		cmd.Flags().StringP(PflagReference.Name, PflagReference.Shorthand, "", PflagReference.Usage)
+	}
+
+	// TODO: cobra does not support string shorthand
```
qweeah, line 112: "[a-zA-Z]" is also a string, maybe change this to the below:
> // TODO: cobra only supports letter shorthand

internal/cmd/flags.go
chloeyin: > Can you check with @SteveLasker that "c" is fine in this case?

I have checked this with @shizhMSFT , we can drop the support.

cmd/docker-generate/manifest.go
```diff
@@ -6,27 +6,42 @@ import (
...
-		&cli.StringFlag{
-			Name:    "output",
-			Aliases: []string{"o"},
-			Usage:   "write to a file instead of stdout",
+type generateManifestOpts struct {
+	output    string
+	reference string
+}
+
+func generateManifestCommandWithOpts() (*cobra.Command, *generateManifestOpts) {
```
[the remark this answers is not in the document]
chloeyin: Now use an argument instead of the return value to get the `opts`.

SteveLasker (approved):
LGTM for the move to cobra.

dtzar:
@priteshbandi - I created a [new discussion](https://github.com/notaryproject/notation/discussions/251) for dropping those commands.  I believe it is a good thing to consider.  I think we should not hold up this PR though on the outcome of the discussion.

However @chloeyin I do think you should skip the `docker-notation` PR for now unless you already have it effectively completed (I know the request was in a bigger PR earlier).

chloeyin:
> @priteshbandi - I created a [new discussion](https://github.com/notaryproject/notation/discussions/251) for dropping those commands. I believe it is a good thing to consider. I think we should not hold up this PR though on the outcome of the discussion.
> 
> However @chloeyin I do think you should skip the `docker-notation` PR for now unless you already have it effectively completed (I know the request was in a bigger PR earlier).

Yes, I can skip the `docker-notation` PR and keep the `docker-notation` pkg using `github.com/urfave/cli/v2` for now.

qweeah (approved):

dtzar:
Since the plan from the discussion yesterday is to remove this.  I'm going to go ahead and merge it.

Issue #211, opened 2022-06-28
notation auth to CNCF ORAS registry


FeynmanZhou:
### Discussed in https://github.com/notaryproject/notation/discussions/170

<div type='discussions-op-text'>

<sup>Originally posted by **jimmyraywv** May  5, 2022</sup>
I am using the CNCF ORAS registry distribution in k8s. I have setup basic auth with `htpasswd`. The oras CLI can authenticate with the registry using `-u ... -p ...` flags. However, the notation `0.7.1-alpha.1` CLI fails auth with the same `-u ... -p ...` flags. Can someone suggest what I am missing? The command I am using is below:

This succeeds:
```
oras discover -o tree $IMAGE  -u <USERNAME> -p <PASSWORD>
```

This does not succeed:
```
notation sign $IMAGE -u <USERNAME> -p <PASSWORD>
2022/05/04 19:03:32 https://<HOST>:443/v2/<REPO>/manifests/<VERSION>: 401 Unauthorized
```</div>

dtzar:
This was fixed in https://github.com/oras-project/oras-go/releases/tag/v2.0.0-alpha

dtzar:
The [package listed we're currently using ](oras.land/oras-go/v2 v2.0.0-20220620164807-8b2a54608a94) is behind where the fix is at.  Recommend someone upgrades to the latest release https://github.com/oras-project/oras-go/releases/tag/v2.0.0-rc.1 and then we can close this issue out.

dtzar:
@binbin-li since you just updated the login info and I see you listed a TODO on this, assigning to you.  Hopefully quick to complete.

binbin-li:
@dtzar looks like this issue is the same as https://github.com/notaryproject/notation/issues/183, and @chloeyin is fixing it with the PR: https://github.com/notaryproject/notation/pull/255

dtzar:
No @shizhMSFT this has not been resolved by 255.  Look at the go.mod file in main now, it still has a todo and has not been updated.

https://github.com/notaryproject/notation/blob/92807080abba772dc63f64b274b442708f5099c4/go.mod#L15

shizhMSFT:
@dtzar This issue is about UX (i.e. the positions of flags). It is not related to `oras-go` but CLI packages resolved by #255

shizhMSFT:
Removing the `dependency-oras-go` label to reduce confusion.

dtzar:
I see the oras-go dependency was updated in #269

PR #255, opened 2022-07-20
use cobra for notation cli

chloeyin:

### What
This PR fixes https://github.com/notaryproject/notation/issues/183 by using `github.com/spf13/cobra` to replace github.com/urfave/cli/v2, which is a cli package also used by [oras-cli](https://github.com/oras-project/oras).
This PR also fixed https://github.com/notaryproject/notation/issues/211, which is caused by `urfave`.

### Notice
According to the statement from previous PR(https://github.com/notaryproject/notation/pull/250), this is the second PR of the whole series, which updates `cmd/notation`.
As we have discussed in the https://github.com/notaryproject/notation/discussions/251, `docker-notation` may be moved to another repo, so I will left most of the code unchanged.  
Though, I changed `signImage` function of the `docker-notation` cmd only to make sure the code can be compiled.
I will add unit test in the next PR.

### DIFF
--- a/cmd/docker-notation/sign.go
+++ b/cmd/docker-notation/sign.go
@@ -34,7 +34,7 @@ var signCommand = &cli.Command{
 func signImage(ctx *cli.Context) error {
 	// TODO: make this change only to make sure the code can be compiled
 	// According to the https://github.com/notaryproject/notation/discussions/251,
-	//  we can update/deprecate it later
+	// we can update/deprecate it later
 	signerOpts := &cmd.SignerFlagOpts{
 		Key:      ctx.String(cmd.FlagKey.Name),
 		KeyFile:  ctx.String(cmd.FlagKeyFile.Name),
--- a/cmd/docker-notation/sign.go
+++ b/cmd/docker-notation/sign.go
@@ -32,7 +32,15 @@ var signCommand = &cli.Command{
 }
 
 func signImage(ctx *cli.Context) error {
-	signer, err := cmd.GetSigner(ctx)
+	// TODO: make this change only to make sure the code can be compiled
+	// According to the https://github.com/notaryproject/notation/discussions/251,
+	//  we can update/deprecate it later
+	signerOpts := &cmd.SignerFlagOpts{
+		Key:      ctx.String(cmd.FlagKey.Name),
+		KeyFile:  ctx.String(cmd.FlagKeyFile.Name),
+		CertFile: ctx.String(cmd.FlagCertFile.Name),
+	}
+	signer, err := cmd.GetSigner(signerOpts)
 	if err != nil {
 		return err
 	}
@@ -55,7 +63,7 @@ func signImage(ctx *cli.Context) error {
 		}
 	}
 	sig, err := signer.Sign(ctx.Context, desc, notation.SignOptions{
-		Expiry: cmd.GetExpiry(ctx),
+		Expiry: cmd.GetExpiry(ctx.Duration(cmd.FlagExpiry.Name)),
 	})
 	if err != nil {
 		return err
--- a/cmd/notation/cert.go
+++ b/cmd/notation/cert.go
@@ -50,12 +50,15 @@ func certAddCommand(opts *certAddOpts) *cobra.Command {
 	command := &cobra.Command{
 		Use:   "add [path]",
 		Short: "Add certificate to verification list",
-		Args:  cobra.ExactArgs(1),
-		PreRun: func(cmd *cobra.Command, args []string) {
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) == 0 {
+				return errors.New("missing certificate path")
+			}
 			opts.path = args[0]
+			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return addCert(cmd, opts)
+			return addCert(opts)
 		},
 	}
 	command.Flags().StringVarP(&opts.name, "name", "n", "", "certificate name")
@@ -68,7 +71,7 @@ func certListCommand() *cobra.Command {
 		Aliases: []string{"ls"},
 		Short:   "List certificates used for verification",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return listCerts(cmd)
+			return listCerts()
 		},
 	}
 	return command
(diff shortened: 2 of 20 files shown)


cmd/notation/cert_gen.go
```diff
@@ -15,30 +15,30 @@ import (
...
 	fmt.Println("generating RSA Key with", bits, "bits")
 	key, keyBytes, err := generateTestKey(bits)
 	if err != nil {
 		return err
 	}
 
 	// generate self-signed certificate
-	cert, certBytes, err := generateTestSelfSignedCert(key, hosts, ctx.Duration("expiry"))
+	expiry := opts.expiry
+	cert, certBytes, err := generateTestSelfSignedCert(key, hosts, expiry)
```
binbin-li, line 41: nit: we can just pass in `opts.expiry` instead of defining a var

cmd/notation/key.go
```diff
@@ -218,7 +266,7 @@ func updateKey(ctx *cli.Context) error {
 	return nil
 }
 
-func listKeys(ctx *cli.Context) error {
+func listKeys(command *cobra.Command) error {
```
binbin-li, line 269: looks like `command` is not used within the method, do we need to keep it?

internal/cmd/options.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+// SignerFlagOpts cmd opts for using cmd.GetSigner
+type SignerFlagOpts struct {
+	Key      string
+	KeyFile  string
+	CertFile string
+}
+
+func (opts *SignerFlagOpts) ApplyFlag(fs *pflag.FlagSet) {
```
binbin-li, line 14: Nit: can you add comments to exported methods though I feel it's self-explained? And `ApplyFlags` probably makes more sense.

cmd/docker-notation/sign.go
```diff
@@ -32,7 +32,15 @@ var signCommand = &cli.Command{
 }
 
 func signImage(ctx *cli.Context) error {
-	signer, err := cmd.GetSigner(ctx)
+	// TODO: make this change only to make sure the code can be compiled
+	// According to the https://github.com/notaryproject/notation/discussions/251,
+	//  we can update/deprecate it later
```
binbin-li, line 37: nit: the indentation is broken

cmd/notation/list.go
```diff
@@ -1,43 +1,51 @@
...
-		return errors.New("no reference specified")
+func listCommand(opts *listOpts) *cobra.Command {
+	if opts == nil {
+		opts = &listOpts{}
+	}
+	cmd := &cobra.Command{
+		Use:     "list [reference]",
+		Aliases: []string{"ls"},
+		Short:   "List signatures from remote",
+		Args:    cobra.ExactArgs(1),
```
patrickzheng200, line 22: Same here. I feel like cobra's ExactArgs error "Error: accepts 1 arg(s), received 0" is kind of vague. The original logic checks ctx.Args().Present(), if not, then it returns error "no reference specified", which explicitly tells the user what is missing.

cmd/notation/pull.go
```diff
@@ -10,66 +11,78 @@ import (
...
+}
+
+func pullCommand(opts *pullOpts) *cobra.Command {
+	if opts == nil {
+		opts = &pullOpts{}
+	}
+	cmd := &cobra.Command{
+		Use:   "pull [reference]",
+		Short: "Pull signatures from remote",
+		Args:  cobra.ExactArgs(1),
```
patrickzheng200, line 32: same here

cmd/notation/push.go
```diff
@@ -1,39 +1,54 @@
...
 
-func runPush(ctx *cli.Context) error {
+func pushCommand(opts *pushOpts) *cobra.Command {
+	if opts == nil {
+		opts = &pushOpts{}
+	}
+	cmd := &cobra.Command{
+		Use:   "push [reference]",
+		Short: "Push signature to remote",
+		Args:  cobra.ExactArgs(1),
```
patrickzheng200, line 28: same here

cmd/notation/sign.go
```diff
@@ -1,78 +1,93 @@
...
+}
+
+func signCommand(opts *signOpts) *cobra.Command {
+	if opts == nil {
+		opts = &signOpts{}
+	}
+	command := &cobra.Command{
+		Use:   "sign [reference]",
+		Short: "Signs artifacts",
+		Args:  cobra.ExactArgs(1),
```
patrickzheng200, line 37: Same here. The original logic's error is "missing reference", which looks more informative.

cmd/notation/cert.go
```diff
@@ -11,94 +11,122 @@ import (
...
+}
+
+func certAddCommand(opts *certAddOpts) *cobra.Command {
+	if opts == nil {
+		opts = &certAddOpts{}
+	}
+	command := &cobra.Command{
+		Use:   "add [path]",
+		Short: "Add certificate to verification list",
+		Args:  cobra.ExactArgs(1),
```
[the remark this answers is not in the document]
chloeyin: If the `path` is an empty string, the command will still fail and the help message shows that it needs a `path` as an argument.
```
path := ctx.Args().First()
if path == "" {
    return errors.New("missing certificate path")
}
```

cmd/notation/list.go
chloeyin: @binbin-li  @qweeah , Do you prefer using `cobra.ExactArgs` or handling args in `PreRunE`?

qweeah: @chloeyin We should customize a arg validation function and provide to the `Args`.
@patrickzheng200 Good catch. I should also apply this to ORAS.

cmd/notation/cert.go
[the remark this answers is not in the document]
patrickzheng200: > If the `path` is an empty string, the command will still fail and the help message shows that it needs a `path` as an argument.
> 
> ```
> path := ctx.Args().First()
> if path == "" {
>     return errors.New("missing certificate path")
> }
> ```

Yes, but if user inputs more than 1 args, cobra.ExactArgs(1) would fail but the orignal logic won't.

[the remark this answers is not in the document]
chloeyin: I will use `Args` field to check args.

patrickzheng200:
LGTM

dtzar:
@gokarnm - can you or someone on your side please review this today?  This is important/blocking to move forward for a number of things, including bug fixes we have.

binbin-li:
lgtm

cmd/notation/login.go
```diff
@@ -1,48 +1,65 @@
...
 Example - Login using $NOTATION_USERNAME $NOTATION_PASSWORD variables:
-	notation login registry.example.com`,
-	ArgsUsage: "[server]",
-	Flags: []cli.Flag{
-		flagUsername,
-		flagPassword,
-		&cli.BoolFlag{
-			Name:  "password-stdin",
-			Usage: "Take the password from stdin",
+  notation login registry.example.com`,
```
shizhMSFT, line 35: nit: The indentation seems need to be revised.

shizhMSFT (approved):
LGTM

PR #256, opened 2022-07-20
test: Add unit tests for notation login

binbin-li:
### What?
Add unit tests for notation login.  https://github.com/notaryproject/notation/pull/218

Coverage:
image
image

### DIFF
--- a/pkg/auth/credential_test.go
+++ b/pkg/auth/credential_test.go
@@ -1,0 +1,82 @@
+package auth
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/notaryproject/notation/pkg/config"
+)
+
+const (
+	errMsg     = "error message"
+	validStore = "pass"
+)
+
+func TestLoadConfig_loadNotationConfigFailed(t *testing.T) {
+	loadOrDefault = func() (*config.File, error) {
+		return nil, fmt.Errorf(errMsg)
+	}
+	_, err := LoadConfig()
+	if err == nil || err.Error() != errMsg {
+		t.Fatalf("Didn't get the expected error, but got: %v", err)
+	}
+}
+
+func TestLoadConfig_NotationConfigContainsAuth(t *testing.T) {
+	loadOrDefault = func() (*config.File, error) {
+		return &config.File{
+			CredentialsStore: validStore,
+		}, nil
+	}
+	file, err := LoadConfig()
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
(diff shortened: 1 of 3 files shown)


pkg/auth/credential_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/notaryproject/notation/pkg/config"
+)
+
+const (
+	errMsg     = "error message"
+	validStore = "pass"
+)
+
+func TestLoadConfig_loadNotationConfigFailed(t *testing.T) {
```
chloeyin, line 15: Should we use an upper-case name `TestLoadConfig_LoadNotationConfigFailed` instead of `TestLoadConfig_loadNotationConfigFailed`?

patrickzheng200:
LGTM

pkg/auth/credential_test.go
binbin-li: Right, I'll get it updated.

pkg/auth/native_store_test.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+}
+
+func TestNativeStore_failedGet(t *testing.T) {
+	s := &nativeAuthStore{
+		programFunc: mockCommandFn,
+	}
+	_, err := s.Get(invalidServerAddress)
+	if err == nil {
+		t.Fatalf("expected err, got %v", err)
```
chloeyin, line 163: Nit: Be consistent with line 173. Either we both use a format string, or we hard code the nil in the message.

pkg/auth/native_store_test.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+}
+
+func TestNativeStore_GetCredentialsStore_NoHelperSet(t *testing.T) {
+	loadConfig = func() (*config.File, error) {
+		return &config.File{}, nil
+	}
+	_, err := GetCredentialsStore(validServerAddress)
+	if err == nil || err.Error() != "could not get the configured credentials store for registry: "+validServerAddress {
+		t.Fatalf("expect error, got nil")
```
chloeyin, line 183: The error may not be nil. Maybe using `t.Fatalf("Didn't get the expected error, but got: %v", err)` is better.

pkg/auth/native_store_test.go
binbin-li: oops, thanks for catching!

chloeyin:
LGTM

shizhMSFT (approved):
LGTM

PR #262, opened 2022-07-25
Remove credential file from spec

### LINKED ISSUE #206: Define local credential experience for Registry Authentication

shizhMSFT:
Resolve #206

### DIFF
--- a/specs/registry-auth.md
+++ b/specs/registry-auth.md
@@ -95,7 +95,7 @@ Notation follows the [Docker Registry v2 authentication][oauth2] specification f
 
 ## Credential Store
 
-As local credentials may be required to access the remote registries, they need to be stored and accessed securely. To achieve maximum security, credential helpers are preferred so that credentials are stored in the system key chain with better protection. If credential helpers are not available, notation will fall back to credential files with proper access control.
+As local credentials may be required to access the remote registries, they need to be stored and accessed securely. To achieve maximum security, credential helpers are preferred so that credentials are stored in the system key chain with better protection. If credential helpers are not available, credentials SHOULD be provided to notation via command line parameters `--username` / `--password` or environment variables `NOTATION_USERNAME` / `NOTATION_PASSWORD`.
 
 ### Credential Helper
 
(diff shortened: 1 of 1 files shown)


specs/registry-auth.md
```diff
@@ -95,7 +95,7 @@ Notation follows the [Docker Registry v2 authentication][oauth2] specification f
 
 ## Credential Store
 
-As local credentials may be required to access the remote registries, they need to be stored and accessed securely. To achieve maximum security, credential helpers are preferred so that credentials are stored in the system key chain with better protection. If credential helpers are not available, notation will fall back to credential files with proper access control.
+As local credentials may be required to access the remote registries, they need to be stored and accessed securely. To achieve maximum security, credential helpers are preferred so that credentials are stored in the system key chain with better protection. If credential helpers are not available, credentials SHOULD be provided to notation via command line parameters `--username` / `--password` or environment variables `NOTATION_USERNAME` / `NOTATION_PASSWORD`.
```
binbin-li, line 98: username/password options are only supported in `notation loging` cmd now. Should we mention it here?

dtzar: Yes, we need to clarify what is or is not supported.

shizhMSFT: username/password options are also supported for `notation sign`, `notation verify`, and other commands which need registry accesses.

gokarnm (approved):

binbin-li (approved):
lgtm

FeynmanZhou:
Do we need to create another issue to update the [CLI baseline](https://github.com/notaryproject/notation/blob/main/specs/notation-cli.md#login) accordingly?

binbin-li:
> Do we need to create another issue to update the [CLI baseline](https://github.com/notaryproject/notation/blob/main/specs/notation-cli.md#login) accordingly?

Yeah, I can make the code change to resolve it. Do we also need to update the cli spec?

FeynmanZhou (approved):

PR #274, opened 2022-07-29
add unit test for cli

chloeyin:

This PR adds unit test for some cli commands, only to make sure the flags are parsed correctly.
This is the last PR of the whole series(https://github.com/notaryproject/notation/pull/255).

### DIFF
--- a/cmd/notation/cache_test.go
+++ b/cmd/notation/cache_test.go
@@ -5,7 +5,7 @@ import (
 	"testing"
 )
 
-func TestCacheListCommand(t *testing.T) {
+func TestCacheListCommand_BasicArgs(t *testing.T) {
 	opts := &cacheListOpts{}
 	cmd := cacheListCommand(opts)
 	expected := &cacheListOpts{
@@ -34,7 +34,7 @@ func TestCacheListCommand(t *testing.T) {
 	}
 }
 
-func TestCachePruneCommand(t *testing.T) {
+func TestCachePruneCommand_BasicArgs(t *testing.T) {
 	opts := &cachePruneOpts{}
 	cmd := cachePruneCommand(opts)
 	expected := &cachePruneOpts{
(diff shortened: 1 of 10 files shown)


JeyJeyGao:
LGTM

cmd/notation/cache_test.go
```diff
@@ -0,0 +1,115 @@
...
+	cmd := cachePruneCommand(nil)
+	if err := cmd.ParseFlags([]string{}); err != nil {
+		t.Fatalf("Parse Flag failed: %v", err)
+	}
+	if err := cmd.Args(cmd, cmd.Flags().Args()); err == nil {
+		t.Fatal("Parse Args expected error, but ok")
+	}
+}
+
+func TestCacheRemoveCommand(t *testing.T) {
```
binbin-li, line 76: Could you specify the testing case like line 107? Same to other tests.

cmd/notation/cache_test.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	if err := cmd.Args(cmd, cmd.Flags().Args()); err != nil {
+		t.Fatalf("Parse Args failed: %v", err)
+	}
+	if !reflect.DeepEqual(*expected, *opts) {
+		t.Fatalf("Expect cache remove opts: %v, got: %v", expected, opts)
+	}
+}
+
+func TestCacheRemoveCommand_MissinArgs(t *testing.T) {
```
binbin-li, line 107: typo: Missing

binbin-li:
Left some tiny comments. Just wanna raise some questions out of scope.
1. Looks like in this PR we focus on the main use cases, successful case and a common failed case which should cover most of cases. But users can fail a command in a lot ways (a combination of flags or args). It's kind of time-wasting to test all of them, just wanna confirm we would do the same to the new PRs. @shizhMSFT 
2. This PR tests the conversion from command input to the internal `Opts` object. For the business logics after it, is there a good way to test it? Maybe just run `cmd.ParseFlags()` like this PR does? Or move the logics to other places for easier testing?

binbin-li:
lgtm overall

shizhMSFT (approved):
LGTM

shizhMSFT:
> Left some tiny comments. Just wanna raise some questions out of scope.
> 
> 1. Looks like in this PR we focus on the main use cases, successful case and a common failed case which should cover most of cases. But users can fail a command in a lot ways (a combination of flags or args). It's kind of time-wasting to test all of them, just wanna confirm we would do the same to the new PRs. @shizhMSFT
> 2. This PR tests the conversion from command input to the internal `Opts` object. For the business logics after it, is there a good way to test it? Maybe just run `cmd.ParseFlags()` like this PR does? Or move the logics to other places for easier testing?

We are still figuring out a better way to test CLI.

Issue #276, opened 2022-08-02
Update Cli spec to reflect username/password options in commands besides login


binbin-li:
Besides `notation login`, we need `username/password` options in other commands as well since the cli doesn't come with `Credentials Store` configured.

The current cli spec shows only `login` command support `username/password` which needs to be updated accordingly.

FeynmanZhou:
Hi @binbin-li , I assume we need to add `username/password` options to the following commands, right?

- notation sign
- notation verify
- notation push
- notation pull
- notation list

From my perspective, if a user provides credentials for authenticated registry and logs into a registry via `notation login`, is it necessary to do authentication in the other commands above? 

I think Notation should allow other operations by default after `notation login`, which means `username/password` options are only needed for the first-time authentication.

binbin-li:
> Hi @binbin-li , I assume you are requesting to add `username/password` options to the following commands, right?
> 
> * notation sign
> * notation verify
> * notation push
> * notation pull
> * notation list
> 
> If a user provides credentials for authenticated registry and logs into a registry via `notation login`, is it necessary to do authentication in the other commands above?
> 
> I think Notation should allow other operations by default after `notation login`, which means `username/password` options are only needed for the first-time authentication.

@FeynmanZhou from the previous discussions, users can use `username/password` in other commands even though logged in. And the user-typed `username/password` would have higher priority than that from first-time login command.

FeynmanZhou:
> > Hi @binbin-li , I assume you are requesting to add `username/password` options to the following commands, right?
> > 
> > * notation sign
> > * notation verify
> > * notation push
> > * notation pull
> > * notation list
> > 
> > If a user provides credentials for authenticated registry and logs into a registry via `notation login`, is it necessary to do authentication in the other commands above?
> > I think Notation should allow other operations by default after `notation login`, which means `username/password` options are only needed for the first-time authentication.
> 
> @FeynmanZhou from the previous discussions, users can use `username/password` in other commands even though logged in. And the user-typed `username/password` would have higher priority than that from first-time login command.

Got you. The current implementation support specifying `username` & `password` for each command option, we can merge https://github.com/notaryproject/notation/pull/293 to make the spec consistent with implementation in alpha.3. 

From an end user's perspective, I think we should recommend that use `notation login` with specifying `username` & `password`  and provide a doc to instruct users to configure the [Credential Store](https://github.com/notaryproject/notation/blob/main/specs/registry-auth.md#credential-store) at the beginning. 

[...]

PR #293, opened 2022-08-10
doc: add missing username/password options to commands

binbin-li:
### What?

Resovles https://github.com/notaryproject/notation/issues/276

For commands other than `notation login`, we should allow them support `username/password` options even without first-time logging.
Since the `login` command requires `credentialsStore` set up which might be missing for users. So we cannot force users always use `login` to save credentials. Other commands should be able to use the credentials directly from input without `credentialsStore` set up.

Updated commands:
1. sign
2. push
3. pull
4. list
5. cache
6. verify

### DIFF
--- a/specs/notation-cli.md
+++ b/specs/notation-cli.md
@@ -67,7 +67,12 @@ COMMANDS:
    help, h     Shows a list of commands or help for one command
 
 OPTIONS:
+   --username value, -u value    Username for registry operations [$NOTATION_USERNAME]
+   --password value, -p value    Password for registry operations [$NOTATION_PASSWORD]
    --help, -h  show help (default: false)
+
+GLOBAL ARGUMENTS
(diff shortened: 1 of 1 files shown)


shizhMSFT (approved):
LGTM

dtzar (approved):
It would be nice to add to the login section something like: "Once login is completed, then -u -p is no longer required for any notation commands against the registry server authenticated."

dtzar:
@yizha1 - since this is just a doc fix and is now at the 1 week mark, I recommend merging it today before notation release.

binbin-li:
> It would be nice to add to the login section something like: "Once login is completed, then -u -p is no longer required for any notation commands against the registry server authenticated."

Good call, I'll add it.

shizhMSFT (approved):
LGTM

PR #301, opened 2022-08-12
bump up version to v0.10.0-alpha.3

yizha1:
Signed-off-by: Yi Zha zhayi@outlook.com

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -2,7 +2,7 @@ package version
 
 var (
 	// Version shows the current notation version, optionally with pre-release.
-	Version = "v1.0.0-alpha.3"
+	Version = "v0.10.0-alpha.3"
 
 	// BuildMetadata stores the build metadata.
 	BuildMetadata = "unreleased"
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -2,7 +2,7 @@ package version
 
 var (
 	// Version shows the current notation version, optionally with pre-release.
-	Version = "v0.7.1-alpha.1"
+	Version = "v1.0.0-alpha.3"
 
 	// BuildMetadata stores the build metadata.
 	BuildMetadata = "unreleased"


shizhMSFT (approved):
LGTM

gokarnm (approved):
LGTM!

internal/version/version.go
```diff
@@ -2,7 +2,7 @@ package version
 
 var (
 	// Version shows the current notation version, optionally with pre-release.
-	Version = "v0.7.1-alpha.1"
+	Version = "v1.0.0-alpha.3"
```
gokarnm, line 5: We should probably not bump major version to v1.0.0 right now, the alpha releases will have patch updates. Which will bump 1.0.0-{prereleasetag} to 1.0.x-{prereleasetag} and require the v1 release to use the same patch version at some point, this would be bit confusing (e.e. v1.0.5-alpha.3 vs 1.0.5)

https://stackoverflow.com/questions/51373151/releasing-alpha-to-beta-to-production-with-semver

```suggestion
	Version = "v0.8.0-alpha.3"
```

gokarnm (changes requested):
Suggesting bumping to v0.8.0-alpha.3 instead.

yizha1:
> Suggesting bumping to v0.8.0-alpha.3 instead.

@gokarnm  I think it should be v0.9.0-alpha.3. We already released v0.9.0-alpha.1 if you check the release page. The version file was not updated by previous release.

priteshbandi:
Can we do `v0.9.0-alpha.3` ?

yizha1:
> Can we do `v0.9.0-alpha.3` ?

v0.10.0-alpha.3 is better, since there are features merged, so the minor version should be stepped.

dtzar:
@gokarnm - it is blocked due to your comment not being resolved.  You have a LGTM, but no approval.

gokarnm (approved):
LGTM!

PR #309, opened 2022-08-19
docs: add a note for dependencies in go.mod file.

yizha1:
Add a note to remind updating the dependencies in go.mod file before starting the release process.

Signed-off-by: Yi Zha <zhayi@outlook.com>

### DIFF
--- a/RELEASE_CHECKLIST.md
+++ b/RELEASE_CHECKLIST.md
@@ -4,8 +4,9 @@
 
 This document describes the checklist to publish a release via GitHub workflow.
 
-## Release Process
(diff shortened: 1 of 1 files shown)


iamsamirzon:
@yizha1  - Should we also call out the need for oras-go dependency in this?  Also, do you want to mention the process to follow for releasing the notation-go and notation-go-core in this same file?

yizha1:
> @yizha1 - Should we also call out the need for oras-go dependency in this? Also, do you want to mention the process to follow for releasing the notation-go and notation-go-core in this same file?

Thanks Samir. I will add oras-go part. I don't think we need to mention the process for go libraries since the process is basically the official Github release process.

yizha1:
@iamsamirzon @vaninrao10   Can you help to review this PR?

dtzar (approved):

dtzar (approved):

PR #315, opened 2022-08-29
refactor:support cose

chloeyin:


internal/cmd/signer.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/notaryproject/notation/pkg/configutil"
 )
 
+// Supported envelope format
 const (
 	CoseFormat = "cose"
 	JswFormat  = "jws"
```
JeyJeyGao, line 19:
```suggestion
	JwsFormat  = "jws"
```

cmd/notation/verify.go
```diff
@@ -105,7 +104,7 @@ func verifySignatures(ctx context.Context, verifier notation.Verifier, manifestD
 		}
```
patrickzheng200, line 104: If reading one path failed, should we continue to read signatures in other paths instead of failing the verification right away?

chloeyin: Continue reading would be better, updated.

PR #330, opened 2022-09-08
feat: write envelopeType default value to config.json

JeyJeyGao:
1. move the envelope format name definition to ./internal/envelope/
2. add default value envelopeType = jws to config.json

### DIFF
--- a/internal/envelope/envelope.go
+++ b/internal/envelope/envelope.go
@@ -1,0 +1,7 @@
+package envelope
+
+// Supported envelope format.
+const (
+	Cose = "cose"
+	Jws  = "jws"
+)
--- a/pkg/configutil/once.go
+++ b/pkg/configutil/once.go
@@ -4,6 +4,7 @@ import (
 	"sync"
 
 	"github.com/notaryproject/notation-go/config"
+	"github.com/notaryproject/notation/internal/envelope"
 )
 
 var (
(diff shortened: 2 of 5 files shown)


pkg/configutil/once.go
```diff
@@ -24,6 +24,10 @@ func LoadConfigOnce() (*config.Config, error) {
 	var err error
 	configOnce.Do(func() {
 		configInfo, err = config.LoadConfig()
+		// set default value
+		if configInfo.EnvelopeType == "" {
+			configInfo.EnvelopeType = "jws"
```
chloeyin, line 29: I found that we can use `JwsFormat` in package `cmd` instead of hard code a string. But it will introduce a cycle import. Do you think we should move `JwsFormat` and `CoseFormat` to another package in the `internal` (Maybe we can create an `internal/envelope`) so that both `cmd` and `configutil` can import them.
Then your code will be
```
import "github.com/notaryproject/notation/internal/evnelope"
// ...
if configInfo.EnvelopeType == "" {
    configInfo.EnvelopeType = envelope.Jws
}
```

JeyJeyGao: Good idea! Thanks.

internal/envelope/envelope.go
```diff
@@ -0,0 +1,7 @@
+package envelope
+
+// Supported envelope format.
+const (
+	Cose = "cose"
+	Jws  = "jws"
```
chloeyin, line 6: nit: Exported const should be capitalize

chloeyin (approved):

PR #331, opened 2022-09-08
feat: An initial implementation of Notation CLI Trust Store

patrickzheng200:
This PR targets the dev-rc.1 branch ONLY.

1. Implemented 'notation cert add', 'notation cert list', 'notation cert show', and 'notation cert delete' commands.
2. Temporarily removed logic of 'notation cert generate-test' as this command needs refactor once we finish trust store and trust policy.
3. Merged new updates by JunJie from the cose branch (i.e. setting up envelope type in config.json).

Signed-off-by: Patrick Zheng [patrickzheng@microsoft.com](mailto:patrickzheng@microsoft.com)


FeynmanZhou:
If we removed `notation cert generate-test` in this PR, we need to consider how to create a test cert with Trust Store implementation.

PR #333, opened 2022-09-08
feat: add policy command

binbin-li:
### What?
Add policy commands. It supports adding/updating/reading/deleting policies.
Spec can be found in https://hackmd.io/AP9Xz9AbRpOy6ASpbok23A?view#Manage-Trust-Policy.

### Test
Manually tested commands. Test cases can be checked in:  https://hackmd.io/@H7a8_rG4SuaKwzu4NLT-9Q/S1j2Yspxj

### Notes
This PR depends on https://github.com/notaryproject/notation-go/pull/132, so the build will fail until it's merged.

### DIFF
--- a/cmd/notation/policy.go
+++ b/cmd/notation/policy.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/notaryproject/notation-go/dir"
 	"github.com/notaryproject/notation-go/verification"
+	"github.com/notaryproject/notation/internal/cmd"
 	"github.com/notaryproject/notation/internal/ioutil"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -18,10 +19,15 @@ type policyOpts struct {
 	name       string
 	scopes     []string
 	level      string
-	override   []string
+	override   string
 	stores     []string
 	identities []string
 	certPath   string
+}
+
+type deleteOpts struct {
+	names     []string
+	confirmed bool
 }
 
 func policyCommand() *cobra.Command {
@@ -33,7 +39,7 @@ func policyCommand() *cobra.Command {
 		policyListCommand(),
 		policyShowCommand(),
 		policyResolveCommand(),
-		policyDeleteCommand(),
+		policyDeleteCommand(nil),
 		policyUpdateCommand(nil),
 		policyAddCommand(nil),
 	)
@@ -64,7 +70,7 @@ func policyShowCommand() *cobra.Command {
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runShowCommand(cmd, policyName)
+			return runShowCommand(policyName)
 		},
 	}
 }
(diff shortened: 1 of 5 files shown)


cmd/notation/policy.go
```diff
@@ -0,0 +1,323 @@
...
+		TrustedIdentities: opts.identities,
+	}, nil
+}
+
+func (opts *policyOpts) ApplyFlags(fs *pflag.FlagSet) {
+	fs.StringSliceVar(&opts.scopes, "scope", nil, "registry scopes")
+	fs.StringVar(&opts.level, "level", "", "verification level")
+	fs.StringSliceVar(&opts.override, "level-override", nil, "config of custom verification level")
+	fs.StringSliceVar(&opts.stores, "trust-store", nil, "trust stores containing trusted roots")
+	fs.StringArrayVar(&opts.identities, "identity", nil, "trusted identities")
```
patrickzheng200, line 314: Question for default values, currently they are just zero values, but most of them are Required fields in the spec to set up a trust policy. Looks like we could early break if user ignored one of those Required fields.

cmd/notation/policy.go
```diff
@@ -0,0 +1,323 @@
...
+	policyDocument, err := verification.LoadDefaultPolicyDocument()
+	if err != nil {
+		return err
+	}
+	policies := policyDocument.ListPolicies()
+
+	return ioutil.PrintPolicyMap(os.Stdout, policies)
+}
+
+func runShowCommand(cmd *cobra.Command, policyName string) error {
```
patrickzheng200, line 195: cmd is not used in this func, can be removed.

cmd/notation/policy.go
```diff
@@ -0,0 +1,323 @@
...
+
+	policy, err := policyDocument.GetPolicy(policyName)
+	if err != nil {
+		return err
+	}
+
+	return ioutil.PrintPolicyMap(os.Stdout, []*verification.TrustPolicy{policy})
+}
+
+func runResolveCommand(cmd *cobra.Command, scope string) error {
```
patrickzheng200, line 209: same here.

cmd/notation/policy.go
```diff
@@ -0,0 +1,323 @@
...
+	if err != nil {
+		return err
+	}
+
+	policies := policyDocument.ListPoliciesWithinScope(scope)
+
+	return ioutil.PrintPolicyMap(os.Stdout, policies)
+}
+
+func runDeleteCommand(cmd *cobra.Command, policyNames []string) error {
```
patrickzheng200, line 220: same here.

cmd/notation/policy.go
```diff
@@ -0,0 +1,323 @@
...
+
+	if err = policyDocument.DeletePolicies(policyNames, dir.UserLevel); err != nil {
+		return err
+	}
+
+	ioutil.PrintDeletedPolicyNames(os.Stdout, policyNames)
+	return nil
+}
+
+func updatePolicy(cmd *cobra.Command, opts *policyOpts) error {
```
patrickzheng200, line 234: same here

cmd/notation/policy.go
```diff
@@ -0,0 +1,323 @@
...
+
+	if err = policyDocument.UpdatePolicies(newPolicies, dir.UserLevel); err != nil {
+		return err
+	}
+
+	ioutil.PrintPolicyNames(os.Stdout, newPolicies, "Updated Policies:")
+	return nil
+}
+
+func addPolicy(cmd *cobra.Command, opts *policyOpts) error {
```
patrickzheng200, line 254: same here.

cmd/notation/policy.go
binbin-li: There is a problem if we set the default values. For the update/add commands, it accepts both a file path to a config file and flags to pass in a policy. The default values only applies to the flag case not the file case. Actually, maybe we can add those default values while we create a policy from flags https://github.com/notaryproject/notation/pull/333/files#diff-25fb7823289ebe97fc1ecc1e5fb30a9799684356ddf8f2e5d0bef4b0006300c1R286. What do you think? @patrickzheng200

patrickzheng200: > There is a problem if we set the default values. For the update/add commands, it accepts both a path to a config file and flags to pass in a policy. The default values only apply to the flag case not the file case. Actually, maybe we can add those default values while we create a policy from flags https://github.com/notaryproject/notation/pull/333/files#diff-25fb7823289ebe97fc1ecc1e5fb30a9799684356ddf8f2e5d0bef4b0006300c1R286. What do you think? @patrickzheng200

Or we just check if user provided those values in the flag case, if one of these flag is not set by the user, we could early break in the flag case.

binbin-li: right, my current implementation works in this way.

patrickzheng200 (approved):
LGTM

dtzar:
Now that https://github.com/notaryproject/notation-go/pull/132 is merged, does this make sense to try to PR to main and kick off a review from everyone else?

binbin-li:
> 

Once we fully test the dev-rc.1 branch, we can make the PR to main.

binbin-li:
> Now that [notaryproject/notation-go#132](https://github.com/notaryproject/notation-go/pull/132) is merged, does this make sense to try to PR to main and kick off a review from everyone else?


Once we fully test the dev-rc.1 branch, we can make the PR to main.

PR #340, opened 2022-09-16
update: updated trust store implementation

patrickzheng200:
This PR targets the dev-rc.1 branch.
1. TrimSpace of storeType, namedStore and cert fileName.
2.  Show all the certificates when a file contains a certificate chain.


cmd/notation/cert.go
```diff
@@ -440,8 +457,8 @@ func printCerts(root string) error {
 	})
 }
 
-// showRootCA displays details of a root certificate
-func showRootCA(cert *x509.Certificate) {
+// showCert displays details of a certificate
+func showCert(cert *x509.Certificate) {
 	fmt.Println("Issuer:", cert.Issuer)
```
chloeyin, line 462: Do we need to use tabwriter to print the result?  For example: `ioutil.PrintPlugins`

cmd/notation/cert.go
```diff
@@ -440,8 +457,8 @@ func printCerts(root string) error {
 	})
 }
 
-// showRootCA displays details of a root certificate
-func showRootCA(cert *x509.Certificate) {
+// showCert displays details of a certificate
+func showCert(cert *x509.Certificate) {
```
chloeyin, line 461: Do we need to use `tabwriter` to print the result?  For example: `ioutil.PrintPlugins`

cmd/notation/cert.go
```diff
@@ -341,17 +345,26 @@ func showCerts(opts *certShowOpts) error {
...
+
+	//write out
+	fmt.Println("Display certificate details. Starting from leaf certificate if it's a certificate chain.")
+	fmt.Println("-----------------------------------------------------------------------------------------")
+	for ind, cert := range certs {
+		showCert(cert)
+		if ind != len(certs)-1 {
+			fmt.Println("-----------------------------------------------------------------------------------------")
+		}
+	}
```
chloeyin, line 357: We can add a function like `DisplayCerts` for showing certs.

chloeyin, line 357: Do you think we need to wrap these to a function like `DisplayCerts`?

patrickzheng200: Oh yeah, sure. will update.

cmd/notation/cert.go
patrickzheng200: Interesting, didn't know we have this. Let me try!

patrickzheng200: I found that tabWriter writes horizontally, which does not work very well for long fields such as cert.PublicKey.  Will stick to vertical list format.

cmd/notation/cert.go
```diff
@@ -179,9 +179,9 @@ func certGenerateTestCommand(opts *certGenerateTestOpts) *cobra.Command {
 
 func addCert(opts *certAddOpts) error {
 	storeType := opts.storeType
-	namedStore := opts.namedStore
+	namedStore := strings.TrimSpace(opts.namedStore)
```
binbin-li, line 182: what about `storeType`?

binbin-li: looks like you check it in `AddCertCore`, but the `errorSlice` would be a slice of same errors. I feel we can break it early.

cmd/notation/cert.go
```diff
@@ -316,17 +320,17 @@ func listCerts(opts *certListOpts) error {
 }
 
 func showCerts(opts *certShowOpts) error {
-	storeType := opts.storeType
+	storeType := strings.TrimSpace(opts.storeType)
```
binbin-li, line 323: I might be wrong, looks like empty check on `storeType` and `namedStore` is required in all commands. Maybe we can move it to a method of `certShowOpts`.

cmd/notation/cert.go
patrickzheng200: Nice catch! Checking storeType as well. As for AddCertCore, since it's export, will need to check storeType again.

binbin-li (approved):
lgtm

PR #342, opened 2022-09-21
feat: use new verify workflow

binbin-li:
### What?

Use the new verification workflow for the `verify` command. Therefore, we have to make some updates to support it.
1. Remove all local verification related flags and arguments, including `signatures`, `certs`, `certFiles`, `pull`, `local` and `mediaType`.
2. Add `config` flag so that users can set up verification plugin configs.
3. Use the new verification workflow.
4. Update the `hello-signing.md`.

### Test
Updated the unit tests.
Command output example:
image


cmd/notation/verify.go
```diff
@@ -45,120 +41,49 @@ func verifyCommand(opts *verifyOpts) *cobra.Command {
...
-		for _, sigDigest := range sigDigests {
-			sigPaths = append(sigPaths, dir.Path.CachedSignature(manifestDigest, sigDigest))
+	// set up verification plugin config.
+	configs := make(map[string]string)
+	for _, c := range opts.config {
+		idx := strings.Index(c, ":")
+		if idx == -1 || idx == 0 || idx == len(c) - 1 {
+			return fmt.Errorf("invalid config option: %s", c)
 		}
+		configs[c[0:idx]] = c[idx+1:]
```
JeyJeyGao, line 63:
```suggestion
		configs[c[:idx]] = c[idx+1:]
```

cmd/notation/verify.go
```diff
@@ -45,120 +41,49 @@ func verifyCommand(opts *verifyOpts) *cobra.Command {
...
-		sigDigests, err := cache.SignatureDigests(manifestDigest)
-		if err != nil {
-			return err
-		}
-		for _, sigDigest := range sigDigests {
-			sigPaths = append(sigPaths, dir.Path.CachedSignature(manifestDigest, sigDigest))
+	// set up verification plugin config.
+	configs := make(map[string]string)
+	for _, c := range opts.config {
+		idx := strings.Index(c, ":")
```
JeyJeyGao, line 59: Is it easier to use `strings.split(c, ":")`

cmd/notation/verify.go
```diff
@@ -45,120 +41,49 @@ func verifyCommand(opts *verifyOpts) *cobra.Command {
...
+		if idx == -1 || idx == 0 || idx == len(c) - 1 {
```
JeyJeyGao, line 60: if c = "key:value1:value2", what should happend?

cmd/notation/verify.go
binbin-li: Good call. It would avoid many potential issues.

cmd/notation/verify.go
```diff
@@ -45,120 +41,49 @@ func verifyCommand(opts *verifyOpts) *cobra.Command {
...
+	configs := make(map[string]string)
+	for _, c := range opts.config {
+		parts := strings.Split(c, ":")
+		if len(parts) != 2 {
+			return fmt.Errorf("invalid config option: %s", c)
 		}
+		configs[parts[0]] = parts[1]
 	}
+	ctx := command.Context()
+	verification.WithPluginConfig(ctx, configs)
```
chloeyin, line 66: Where do we use the return value?

cmd/notation/verify.go
```diff
@@ -45,120 +41,49 @@ func verifyCommand(opts *verifyOpts) *cobra.Command {
...
-			return err
-		}
-		for _, sigDigest := range sigDigests {
-			sigPaths = append(sigPaths, dir.Path.CachedSignature(manifestDigest, sigDigest))
+	// set up verification plugin config.
+	configs := make(map[string]string)
+	for _, c := range opts.config {
+		parts := strings.Split(c, ":")
+		if len(parts) != 2 {
+			return fmt.Errorf("invalid config option: %s", c)
```
patrickzheng200, line 61: Looks like for each string in opts.config, it has to be in a "xxx:xxx" format? Otherwise it is an invalid config option. Should we add this as a reminder in the corresponding --help description?

chloeyin: I found that sign with plugin config uses a string as flag value, e.g., `notation sign --pluginConfig key0=val0,key1=val1,key2=val2`, which is different from verification. Is it by design?

cmd/notation/verify.go
binbin-li: configs would be retrieved via method: https://github.com/notaryproject/notation-go/blob/dev-rc.1/verification/types.go#L225 in https://github.com/notaryproject/notation-go/blob/dev-rc.1/verification/verifier_helpers.go#L299

cmd/notation/verify.go
binbin-li: Updated to be consistent with `sign` command. Thanks for reminder!

cmd/notation/verify.go
chloeyin: > configs would be retrieved via method: https://github.com/notaryproject/notation-go/blob/dev-rc.1/verification/types.go#L225 in https://github.com/notaryproject/notation-go/blob/dev-rc.1/verification/verifier_helpers.go#L299

But `context.WithValue` will create a new context instead of putting value in the current context.
https://pkg.go.dev/context#WithValue
It is more like a linked list.

binbin-li: Fixed it. thanks!

internal/ioutil/print.go
```diff
@@ -52,3 +53,29 @@ func PrintCertificateMap(w io.Writer, v []config.CertificateReference) error {
...
 }
+
+func PrintVerificationResults(w io.Writer, v []*verification.SignatureVerificationOutcome, resultErr error) error {
+	tw := newTabWriter(w)
+
+	overallResult := "success"
+	if resultErr != nil {
+		overallResult = "failure"
+	}
+	fmt.Fprintf(tw, "OVERALL RESULT: %s\n", overallResult)
```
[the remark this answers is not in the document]
binbin-li: The print out format is not decided yet, I'll check with Yi once he has done some manual testing.

PR #359, opened 2022-09-28
feat: support tagged reference

binbin-li:
### What?

1. Support passing in a reference in the format: registry/repository:v1. Before that, only a reference with digest tag is supported.
2. Refactor the print out format to be more readable.

### Test
Success case:
image

Failure case:
failure

### DIFF
--- a/cmd/notation/verify.go
+++ b/cmd/notation/verify.go
@@ -47,8 +47,19 @@ func verifyCommand(opts *verifyOpts) *cobra.Command {
 }
 
 func runVerify(command *cobra.Command, opts *verifyOpts) error {
-	// initialize.
-	verifier, err := getVerifier(opts)
+	// resolve the given reference and set the digest.
+	manifestDesc, err := getManifestDescriptorFromReference(command.Context(), &opts.SecureFlagOpts, opts.reference)
+	if err != nil {
+		return err
+	}
+	ref, err := orasregistry.ParseReference(opts.reference)
+	if err != nil {
+		return err
+	}
+	ref.Reference = manifestDesc.Digest.String()
+
+	// initialize verifier.
+	verifier, err := getVerifier(opts, ref)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


JeyJeyGao:
LGTM

internal/ioutil/print.go
```diff
@@ -54,28 +54,28 @@ func PrintCertificateMap(w io.Writer, v []config.CertificateReference) error {
...
 func printOutcomes(tw *tabwriter.Writer, outcomes []*verification.SignatureVerificationOutcome) {
-	if len(outcomes) > 0 {
-		fmt.Fprintln(tw, "DETAILED ERRORS:")
-		for _, outcome := range outcomes {
-			fmt.Println(outcome.Error)
-		}
+	if len(outcomes) == 1 {
+		fmt.Println("1 signature failed verification, error is listed as below:")
+	} else {
+		fmt.Printf("%d signatures failed verification, errors are listed as below:\n", len(outcomes))
```
patrickzheng200, line 75: If there are more than one failures, should we provide more info besides the errors? For example, when two outcome errors are different, we should print out which signature envelope leads to the corresponding outcome.Error.

yizha1:
Regarding the success case, is there an empty line before the digest? If yes, I suggest removing the empty line.

internal/ioutil/print.go
binbin-li: I discussed about that with Yi yesterday. It would look better if we can print out the signature digest for each error. But the result returned by the `verify` method does not include the digest. So we might need another PR to support that in `notation-go` verify method. We can make the change once @yizha1 completes the `Verify` spec.

binbin-li:
> Regarding the success case, is there an empty line before the digest? If yes, I suggest removing the empty line.

that's the old screenshot, I've updated it.

internal/ioutil/print.go
patrickzheng200: That makes sense. Otherwise, a single print out such as "signature is not produced by a trusted signer" does not really help the user to figure out what's going on. How about adding a TODO in the comment so we won't forget to add this part in the future?

chloeyin:
BTW, could you use `cmd.ParseFlagPluginConfig` to parse the plugin config? https://github.com/notaryproject/notation/blob/4bc23dbe15dd2e4f60ac96189d939771e8ee9081/cmd/notation/sign.go#L123
We don't need to parse the config string manully.

binbin-li:
> BTW, could you use `cmd.ParseFlagPluginConfig` to parse the plugin config?
> 
> https://github.com/notaryproject/notation/blob/4bc23dbe15dd2e4f60ac96189d939771e8ee9081/cmd/notation/sign.go#L123
> 
> 
> We don't need to parse the config string manully.

Sure, thanks for calling it out.

cmd/notation/verify.go
```diff
@@ -47,8 +47,19 @@ func verifyCommand(opts *verifyOpts) *cobra.Command {
...
+	// resolve the given reference and set the digest.
+	manifestDesc, err := getManifestDescriptorFromReference(command.Context(), &opts.SecureFlagOpts, opts.reference)
+	if err != nil {
+		return err
+	}
+	ref, err := orasregistry.ParseReference(opts.reference)
+	if err != nil {
+		return err
+	}
+	ref.Reference = manifestDesc.Digest.String()
```
chloeyin, line 59: I suggest wrap this logic into a function like `resolveReference`

patrickzheng200 (approved):
LGTM

chloeyin (approved):

PR #365, opened 2022-09-29
feat: support cose

chloeyin:
Add cose envelope support


go.mod
```diff
@@ -5,20 +5,25 @@ go 1.19
...
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/opencontainers/distribution-spec/specs-go v0.0.0-20220620172159-4ab4752c3b86 // indirect
 	github.com/opencontainers/image-spec v1.0.3-0.20211202183452-c5a74bcca799 // indirect
 	github.com/oras-project/artifacts-spec v1.0.0-rc.2 // indirect
+	github.com/x448/float16 v0.8.4 // indirect
 	golang.org/x/sync v0.0.0-20220722155255-886fb9371eb4 // indirect
 	golang.org/x/sys v0.0.0-20220825204002-c680a09ffe64 // indirect
 )
+
+replace github.com/notaryproject/notation-go => github.com/chloeyin/notation-go v0.8.0-alpha.1.0.20220929023947-bac6075dccee
```
shizhMSFT, line 29: Since we are referencing the `main` branch, we can remove this `replace`.

cmd/notation/init.go
```diff
@@ -0,0 +1,6 @@
+package main
+
+import (
+	_ "github.com/notaryproject/notation-core-go/signature/cose"
+	_ "github.com/notaryproject/notation-core-go/signature/jws"
+)
```
shizhMSFT, line 6: We can remove this file since those packages are imported by `internal/envelope/envelope.go` and many other places.

internal/envelope/envelope.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/notaryproject/notation-core-go/signature/cose"
+	"github.com/notaryproject/notation-core-go/signature/jws"
+	gcose "github.com/veraison/go-cose"
+)
+
+// Supported envelope format.
+const (
+	Cose = "cose"
+	Jws  = "jws"
```
shizhMSFT, line 14:
```suggestion
	COSE = "cose"
	JWS  = "jws"
```

cmd/notation/init.go
chloeyin: removed

internal/envelope/envelope.go
chloeyin: Updated

go.mod
chloeyin: Removed.

shizhMSFT (approved):
LGTM

cmd/notation/sign_test.go
```diff
@@ -23,9 +24,10 @@ func TestSignCommand_BasicArgs(t *testing.T) {
 			},
 		},
 		SignerFlagOpts: cmd.SignerFlagOpts{
-			Key:      "key",
-			KeyFile:  "keyfile",
-			CertFile: "certfile",
+			Key:          "key",
+			KeyFile:      "keyfile",
+			CertFile:     "certfile",
+			EnvelopeType: envelope.JWS,
```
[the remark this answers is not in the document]
chloeyin: Yes. This test is actually not passing the envelope-type, I expect the default value would be jws.
You can expand the line below
image

cmd/notation/push.go
```diff
@@ -70,7 +71,12 @@ func runPush(command *cobra.Command, opts *pushOpts) error {
 			return err
 		}
 		// pass in nonempty annotations if needed
-		sigDesc, _, err := sigRepo.PutSignatureManifest(command.Context(), sig, manifestDesc, make(map[string]string))
+		// TODO: understand media type in a better way
```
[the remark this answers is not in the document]
chloeyin: Both push and verify will need to speculate format now not only for caching. 
For verifying, the new verification workflow with trust store and trust policy doesn't need to speculate format because it will pull from remote.
https://github.com/notaryproject/notation-go/blob/3d8de5f05facb36c97df60de6e49c936e5ac6b41/verification/verifier.go#L100-L116

For pushing, there is a flag named `signatures`, which is the paths to some signatures. CLI will first read the file and push its content to the remote. 
https://github.com/notaryproject/notation/blob/main/specs/notation-cli.md#push

https://github.com/notaryproject/notation/blob/4666267499a11f5c56af7211fee36a27e78cd582/cmd/notation/push.go#L52-L71

So, for RC1, verify doesn't need speculate, but we still need to speculate format if user pushes a signature from local file system unless we abandon this feature.
If still necessary, I can open an issue to track the push problem.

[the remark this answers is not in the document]
priteshbandi: Yeah unless there is specific need to write signature to disk(cache) we should abandon this feature.

Opened an issue https://github.com/notaryproject/notation/issues/379

docs/hello-signing.md
```diff
@@ -72,48 +72,53 @@ ACME Rockets will only deploy software that's been scanned and approved by the A
...
   At this point, the results are empty, as there are no existing signatures
 
 ## Signing a Container Image
 
-To get things started quickly, the Notation cli supports generating self signed certificates. As you automate the signing of content, you will most likely want to create and store the private keys in a key vault. (Detailed production steps will be covered later)
+To get things started quickly, the Notation cli supports self-generated signing certificate. As you automate the signing of content, you will most likely want to create and store the private keys in a key vault. (Detailed production steps will be covered later)
 
-- Generate a self-signed test certificate for signing artifacts
-  The following will generate a self-signed X.509 certificate under the `~/config/notation/` directory
+- Create a self-signed test certificate chain for signing artifacts
```
priteshbandi, line 84:
```suggestion
- Create a self-signed test certificate for signing artifacts
```

internal/envelope/envelope.go
```diff
@@ -0,0 +1,39 @@
...
+// TODO: do we nned to speculate evenlope format?
+//
+// Both push and verify will need to speculate format now not only for caching.
+//
+// For verifying, the new verification workflow with trust store
+// and trust policy doesn't need to speculate format because it will pull from remote.
+//
+// For pushing, there is a flag named signatures, which is the paths to some signatures. CLI will first read the file and push its content to the remote.
+//
+// For RC1, verify doesn't need speculate, but we still need to speculate format if user pushes a signature from local file system
```
priteshbandi, line 28: Please like issue https://github.com/notaryproject/notation/issues/379 and remove the description here and add it to issue https://github.com/notaryproject/notation/issues/379

As per discussion with Yi we are getting rid of signature cache for RC1.

chloeyin: Add comments to #379

priteshbandi (approved):
LGTM

Issue #304, opened 2022-08-13
UX: Improve the help doc and output for verification


yizha1:
### Summary
Currently only a digest showed as the output after notation verify successfully executed. What is this digest about: the digest of the signature, signature manifest or image manifest? see example below:
```
notation --cert $CERT_NAME verify $IMAGE
sha256:18adff7f255319415112345671bb41076de4a864eb792c35c20f0f6b4aa4c458
```
Currently the digest from the output is actually the digest of image manifest.

The help doc for notation verify is not accurate as well. Currently it is showed as below:
```
notation verify -h
Verifies OCI Artifacts
```
It is actually to verify the signature of an OCI artifact, not SBOM, VA or any other supply chain artifacts. 

### User Scenario
As a user, after I verify a signature of the supplied image or OCI artifact against the certificate successfully, I want to make sure the verification is performed on the correct image or OCI artifact against correct certificate.

### Improvement
Here is one idea of improving the output after a successful verification.

```
notation --cert $CERT_NAME verify $IMAGE
The signature is verified against the certificate $CERT_NAME for the OCI artifact sha256:18adff7f255319415112345671bb41076de4a864eb792c35c20f0f6b4aa4c458 

```
NOTE: Maybe need to consider the case of multiple signatures against the same certificate.

cc @shizhMSFT @dtzar @FeynmanZhou @SteveLasker

priteshbandi:
We also need to make output script-friendly.

yizha1:
My proposal @priteshbandi  @dtzar @iamsamirzon @binbin-li @shizhMSFT 
```
# Verify the artifact identified by digest
> notation verify localhost:5000/net-monitor@sha256:1111111111111111111111111111111111111111111111111111111111111111
Verify success for localhost:5000/net-monitor@sha256:1111111111111111111111111111111111111111111111111111111111111111


# Verify the artifact identified by tag
> notation verify localhost:5000/net-monitor:v1
Warning: Tag is used. Always use digest to identify the reference uniquely and immutably.

Resolve tag `v1` to digest `sha256:1111111111111111111111111111111111111111111111111111111111111111`
Verify success for localhost:5000/net-monitor@sha256:1111111111111111111111111111111111111111111111111111111111111111

# Verify success with --debug flag
> notation verify localhost:5000/net-monitor@sha256:1111111111111111111111111111111111111111111111111111111111111111
Use certificate: <certfile_relative_path>
Use trust policy: <policy_name>
[...]

PR #373, opened 2022-10-11
feat: use new verify workflow

### LINKED ISSUE #216: Implementing verify command
**Summary**

Implementing verify command according to the spec #88 

**Intended Outcome**
`notation verify` command works as defined in the spec

**Additional context**

binbin-li:
### What?

Use the new verification workflow for the `verify` command. Therefore, we have to make some updates to support it.
1. Remove all local verification related flags and arguments, including `signatures`, `certs`, `certFiles`, `pull`, `local` and `mediaType`.
2. Add `config` flag so that users can set up verification plugin configs.
3. Use the new verification workflow.
4. Support both tag and digest reference.
5. Format the result.


cmd/notation/verify.go
```diff
@@ -45,113 +40,60 @@ func verifyCommand(opts *verifyOpts) *cobra.Command {
...
-		}
-		idx := slices.Index(cfg.VerificationCertificates.Certificates, name)
-		if idx < 0 {
-			return nil, errors.New("verification certificate not found: " + name)
-		}
-		paths = append(paths, cfg.VerificationCertificates.Certificates[idx].Path)
+	manifestDesc, err := getManifestDescriptorFromReference(command.Context(), &opts.SecureFlagOpts, opts.reference)
+	if err != nil {
+		return orasregistry.Reference{}, err
 	}
```
shizhMSFT, line 95: We need to resolve the tag only if it is not a digest.

cmd/notation/verify.go
```diff
@@ -1,39 +1,34 @@
...
 type verifyOpts struct {
-	RemoteFlagOpts
-	signatures []string
-	certs      []string
-	certFiles  []string
-	pull       bool
-	reference  string
+	SecureFlagOpts
+	reference string
+	config    string
```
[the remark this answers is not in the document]
binbin-li: The reason for using `string` instead of slice is to be consistent with `sign` and `key` commands, https://github.com/notaryproject/notation/blob/main/cmd/notation/sign.go#L27 so that I can utilize https://github.com/notaryproject/notation/blob/main/internal/cmd/flags.go#L91 to parse the configs.

internal/ioutil/print.go
```diff
@@ -52,3 +53,30 @@ func PrintCertificateMap(w io.Writer, v []config.CertificateReference) error {
 	}
 	return tw.Flush()
 }
+
+func PrintVerificationResults(w io.Writer, v []*verification.SignatureVerificationOutcome, resultErr error, digest string) error {
+	tw := newTabWriter(w)
+
+	if resultErr == nil {
+		fmt.Fprintf(tw, "%s\n", digest)
```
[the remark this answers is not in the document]
priteshbandi: Depending upon the enforcement level in the trust policy configuration, the signature verification might succeed but some validations might have failed(where the enforcement level is logged). In such cases, we want to display the failed validations as warnings.

The signature that passed signature verification would be last signature in `SignatureVerificationOutcome` array.

https://github.com/notaryproject/notaryproject/blob/main/trust-store-trust-policy-specification.md#signature-verification-details

internal/ioutil/print.go
```diff
@@ -52,3 +53,30 @@ func PrintCertificateMap(w io.Writer, v []config.CertificateReference) error {
...
+	if len(outcomes) == 1 {
+		fmt.Println("1 signature failed verification, error is listed as below:")
+	} else {
+		fmt.Printf("%d signatures failed verification, errors are listed as below:\n", len(outcomes))
+	}
+
+	for _, outcome := range outcomes {
+		// TODO: print out the signature digest once the outcome contains it.
+		fmt.Printf("%s\n\n", outcome.Error.Error())
+	}
```
priteshbandi, line 81: IMO this error message is not really helpful and should only be printed in debug level.

In non-debug mode, we should display some generic message like:
`Signature verification failed for all the xx signatures associated with ${reference}. `

In debug mode
```
Signature verification failed for all the xx signatures associated with ${reference}. ` The failure reasons for each signature are as follows:
1. ${signature-reference}: ${error message}
2.  ${signature-reference}: ${error message}
````

Comment (binbin-li) [reply internal/ioutil/print.go:81]: @yizha1 Could you help check if you need to update the spec?

File: internal/ioutil/print.go (hunk 3 above)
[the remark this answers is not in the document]
Comment (binbin-li) [reply internal/ioutil/print.go:61]: right, I was thinking to print out the digest for failed signatures with the error for user experience. But seems the returned outcome doesn't contain it. Please correct me if I miss something.

File: internal/ioutil/print.go (hunk 4 above)
Comment (yizha1) [reply internal/ioutil/print.go:81]: I can update verify CLI spec for this accordingly

Comment (priteshbandi) [reply internal/ioutil/print.go:81]: Tracking it in https://github.com/notaryproject/notation/issues/304

**Review by priteshbandi — COMMENTED** [review]
LGTM

File: internal/ioutil/print.go (hunk 3 above)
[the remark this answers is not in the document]
Comment (priteshbandi) [reply internal/ioutil/print.go]: Tracking it in https://github.com/notaryproject/notation/issues/304

**Review by priteshbandi — APPROVED** [review]

File: cmd/notation/verify.go [hunk 5]
```diff
@@ -1,53 +1,34 @@
...
 func verifyCommand(opts *verifyOpts) *cobra.Command {
 	if opts == nil {
 		opts = &verifyOpts{}
 	}
 	command := &cobra.Command{
-		Use:   "verify [reference]",
-		Short: "Verify OCI artifacts",
-		Long: `Verify OCI artifacts
-
-Prerequisite: a trusted certificate needs to be generated or added using the command "notation cert".
```
JeyJeyGao, line 40: Do we need to delete it?

cmd/notation/verify.go
```diff
@@ -1,53 +1,34 @@
...
-Example - Verify a signature associated with an OCI artifact identified by the digest:
-  notation verify <registry>/<repository>@<digest>
-
-Example - Verify a signature using a trusted certificate in a specified path:
-  notation verify --cert-file <cert_path> <registry>/<repository>:<tag>
-`,
+		Use:   "verify [flags] <reference>",
+		Short: "Verifies OCI Artifacts",
+		Long: `Verifies OCI Artifacts:
+  notation verify [--config <key>=<value>,...] [--username <username>] [--password <password>] <reference>`,
```
rgnote, line 31:
```suggestion
  notation verify [--pluginConfig <key>=<value>,...] [--username <username>] [--password <password>] <reference>`,
```

cmd/notation/verify.go
```diff
@@ -59,119 +40,75 @@ Example - Verify a signature using a trusted certificate in a specified path:
 			return runVerify(cmd, opts)
 		},
 	}
-	setFlagSignature(command.Flags(), &opts.signatures)
-	command.Flags().StringSliceVarP(&opts.certs, "cert", "c", []string{}, "certificate names for verification")
-	command.Flags().StringSliceVar(&opts.certFiles, cmd.PflagCertFile.Name, []string{}, "certificate files for verification")
-	command.Flags().BoolVar(&opts.pull, "pull", true, "pull remote signatures before verification")
 	opts.ApplyFlags(command.Flags())
+	command.Flags().StringVarP(&opts.pluginConfig, "config", "c", "", "list of comma-separated {key}={value} pairs that are passed as is to the plugin")
```
rgnote, line 44:
```suggestion
	command.Flags().StringVarP(&opts.pluginConfig, "pluginConfig", "pc", "", "list of comma-separated {key}={value} pairs that are passed as is to the plugin")
```

internal/ioutil/print.go
```diff
@@ -52,3 +53,24 @@ func PrintCertificateMap(w io.Writer, v []config.CertificateReference) error {
...
+
+	fmt.Fprintf(tw, "ERROR: %s\n\n", resultErr.Error())
+	printOutcomes(tw, v, digest)
+
+	return tw.Flush()
+}
+
+func printOutcomes(tw *tabwriter.Writer, outcomes []*verification.SignatureVerificationOutcome, digest string) {
+	fmt.Printf("Signature verification failed for all the %d signatures associated with digest: %s\n", len(outcomes), digest)
+
```
rgnote, line 74: I understand that we are working on a spec to describe the CLI inputs and outputs, but we should print the error messages associated with the individual outcomes in the short-term to make this command more useful. Can you print the outcomes like below in the interim?

```
Signature verification failed for all the 3 signatures associated with digest: asdf....asdfasdf

Signature #1 : <Error>
Signature #2 : <Error>
Signature #3 : <Error>

```

cmd/notation/verify.go
```diff
@@ -59,119 +40,75 @@ Example - Verify a signature using a trusted certificate in a specified path:
...
 
-	// write out
-	fmt.Println(manifestDesc.Digest)
-	return nil
+	// core verify process.
+	ctx := verification.WithPluginConfig(command.Context(), configs)
+	outcomes, err := verifier.Verify(ctx, ref.String())
+
+	// write out.
+	return ioutil.PrintVerificationResults(os.Stdout, outcomes, err, ref.Reference)
```
rgnote, line 72: We should return the error from `verifier.Verify()` here so that CLI can throw a non-zero error code when signature verification fails.

internal/ioutil/print.go
binbin-li: Good call, I'll add error messages.

cmd/notation/verify.go
binbin-li: I'll keep the `Prerequisite` comment.

cmd/notation/verify.go
binbin-li: Actually Cobra doesn't support shorthand with more than one character.

patrickzheng200 (approved):
LGTM

cmd/notation/verify_test.go
```diff
@@ -49,24 +34,15 @@ func TestVerifyCommand_MoreArgs(t *testing.T) {
...
-		signatures: []string{},
-		pull:       false,
+		pluginConfig: "key1=val1,key2=val2",
 	}
 	if err := command.ParseFlags([]string{
 		expected.reference,
 		"--plain-http",
-		"--pull=false",
-		"--media-type=mediaT"}); err != nil {
+		"--pluginConfig", expected.pluginConfig}); err != nil {
```
priteshbandi, line 45:
```suggestion
		"--plugin-config", expected.pluginConfig}); err != nil {
```

Can we use kebab-case everywhere for command input?

internal/ioutil/print.go
```diff
@@ -52,3 +53,28 @@ func PrintCertificateMap(w io.Writer, v []config.CertificateReference) error {
 	}
 	return tw.Flush()
 }
+
+func PrintVerificationResults(w io.Writer, v []*verification.SignatureVerificationOutcome, resultErr error, digest string) error {
+	tw := newTabWriter(w)
+
+	if resultErr == nil {
+		fmt.Fprintf(tw, "Signature verification succeeded for %s\n", digest)
+		// TODO: print out failed validations as warnings.
```
priteshbandi, line 62: Can you please also add issue with TODO: https://github.com/notaryproject/notation/issues/304
Would allow easier tracking.

PR #375, opened 2022-10-11
moved notation-core-go as indirect dependency

patrickzheng200:
Notation CLI will call notation-go's alias of notation-core-go api.


binbin-li (approved):
lgtm

PR #374, opened 2022-10-11
spec: update notation cli md file as index for sub-commands

yizha1:
Update notation cli md file to reflect the changes in rc.1 release and add references to each sub-command.

Issue #303, opened 2022-08-13
UX: Improve the output for a successful signing


yizha1:
### Summary
Currently only a digest showed as the output after notation sign successfully executed. What is this digest about, the digest of the signature, signature manifest or image manifest? see example below:
```
notation sign --key $KEY_NAME $IMAGE
sha256:18adff7f255319415112345671bb41076de4a864eb792c35c20f0f6b4aa4c458
```
nowadays the digest is actually the digest of image manifest.
### User Scenario
As a user, after I sign an image or artifact successfully, I want to know where the signature is stored and what the signature refers to, so that I can make sure the signature refers to a correct image or artifact, make sure sigature is stored properly, and I can CRUD later.

### Improvement
Here is one idea of improving the output after a successful signing

```
notation sign --key $KEY_NAME $IMAGE
Signature is pushed to regsitry xxx.xxx.xxx, and refers to sha256:18adff7f255319415112345671bb41076de4a864eb792c35c20f0f6b4aa4c458
```

cc @shizhMSFT @dtzar @FeynmanZhou @SteveLasker

shizhMSFT:
We might also need to make output script-friendly.

dtzar:
> We might also need to make output script-friendly.

Yes there could be something like a `-o json` option to output it json format

yizha1:
Any comments? @dtzar @gokarnm @priteshbandi @iamsamirzon @vaninrao10 
```
# sign an artifact identified by digest
> notation sign localhost:5000/net-monitor@sha256:1111111111111111111111111111111111111111111111111111111111111111
Sign success. Signature has been attached to localhost:5000/net-monitor:sha2561111111111111111111111111111111111111111111111111111111111111111

# sign an artifact identified by tag
> notation sign localhost:5000/net-monitor:v1
Warning: Tag is used. Always use digest to identify the reference uniquely and immutably.

Resolve tag "v1" to digest "sha256:1111111111111111111111111111111111111111111111111111111111111111"
Sign success. Signature has been attached to localhost:5000/net-monitor:sha256:1111111111111111111111111111111111111111111111111111111111111111
```

patrickzheng200:
> @yizha1 Similar to my comments in [#304 (comment)](https://github.com/notaryproject/notation/issues/304#issuecomment-1293593766), I suggest updating `Sign success` to `Sign succeeded`. This is more commonly used.
> 
[...]

PR #404, opened 2022-10-18
update: clean up notation CLI

patrickzheng200:
1. Deprecated cache/push/pull commands
2. Removed usage of cache in all commands


JeyJeyGao:
LGTM

cmd/notation/push.go
```diff
@@ -2,91 +2,12 @@ package main
...
-		}
-
-		// write out
-		fmt.Println(sigDesc.Digest)
-	}
-
-	return nil
-}
-
 func pushSignature(ctx context.Context, opts *SecureFlagOpts, ref string, sig []byte) (notation.Descriptor, error) {
```
shizhMSFT, line 11: nit: can we move this function to `sign.go`?

cmd/notation/sign.go
```diff
@@ -98,14 +93,6 @@ func runSign(command *cobra.Command, cmdOpts *signOpts) error {
 	}
 
 	// write out
-	path := cmdOpts.output
-	if path == "" {
-		path = dir.Path.CachedSignature(digest.Digest(desc.Digest), digest.FromBytes(sig))
-	}
-	if err := osutil.WriteFile(path, sig); err != nil {
-		return err
-	}
```
shizhMSFT, line 107: Removing this block of code also removes the capability of outputting the signature to a specified local file path.

cmd/notation/sign.go
```diff
@@ -119,7 +106,7 @@ func runSign(command *cobra.Command, cmdOpts *signOpts) error {
 		}
 	}
 
-	fmt.Println(desc.Digest)
+	fmt.Println("Sign succeeded for", desc.Digest)
```
shizhMSFT, line 109: @FeynmanZhou @yizha1 Is this intended?

cmd/notation/sign.go
shizhMSFT: Looks like we will not support local signing for this milestone. Then we always do remote (content) signing. Could you also remove the `--push` flag?

shizhMSFT: Also, we need to remove `--push-reference` since we are not able to push signatures to other repos for the current workflow.

shizhMSFT (changes requested):

cmd/notation/sign.go
yizha1: We can keep the old output for now. There is one issue #303  addressing this output message. We can update it later.

patrickzheng200: I see. Will remove this change for now.

priteshbandi (approved):
LGTM

specs/notation-cli.md
```diff
@@ -5,14 +5,11 @@ The CLI commands are what's currently available in [notation v0.7.1-alpha.1](htt
 
 ## Table of Contents
```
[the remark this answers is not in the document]
patrickzheng200: I guess #374 might still take a while to be merged? Maybe we could rebase #374 after this PR?

shizhMSFT (approved):
LGTM

PR #405, opened 2022-10-19
feat: Added `notation certificate` command for trust store

patrickzheng200:
This PR is the implementation of `notation certificate` command regarding the use of trust store. 
Based on community meeting, this command only effects on User level.


cmd/notation/sign.go
```diff
@@ -149,3 +128,28 @@ func prepareSigningContent(ctx context.Context, opts *signOpts) (notation.Descri
...
+		return notation.Descriptor{}, err
+	}
+	manifestDesc, err := getManifestDescriptorFromReference(ctx, opts, ref)
+	if err != nil {
+		return notation.Descriptor{}, err
+	}
+
+	// core process
+	// pass in nonempty annotations if needed
+	sigMediaType, err := envelope.SpeculateSignatureEnvelopeFormat(sig)
```
chloeyin, line 145: Can we use envelope type passed to the sign command instead of speculating it manually?

patrickzheng200: This file is no longer in the scope of this PR. Just rebased.

internal/osutil/file.go
```diff
@@ -36,3 +39,92 @@ func WriteFileWithPermission(path string, data []byte, perm fs.FileMode, overwri
...
+// Copy the src file to dst. Existing file will be overwritten.
+func Copy(src, dst string) (int64, error) {
+	sourceFileStat, err := os.Stat(src)
+	if err != nil {
+		return 0, err
+	}
+
+	if !sourceFileStat.Mode().IsRegular() {
+		return 0, fmt.Errorf("%s is not a regular file", src)
+	}
```
chloeyin, line 52: We can't copy from a symbolic link to the store?

patrickzheng200: According to the [spce](https://github.com/notaryproject/notaryproject/blob/main/trust-store-trust-policy-specification.md#trust-store):
Symlinks are not supported for the named store directories or certificate files. Implementation MUST validate that the named store directory or certificate files are not symlinks, and fail if it does not meet this condition.

chloeyin: I thought only the target cannot be a symbolic link, but the source file can be a symbolic link to another file?

internal/certificate/cert.go
```diff
@@ -0,0 +1,202 @@
...
+	if display {
+		fmt.Printf("Successfully added %s to named store %s of type %s\n", filepath.Base(certPath), namedStore, storeType)
+	}
+
+	return nil
+}
+
+// ListCertsCore walks through root and lists all x509 certificates in it
+func ListCertsCore(root string) error {
+	return filepath.WalkDir(root, func(path string, d fs.DirEntry, err error) error {
```
rgnote, line 101: Can we use https://github.com/notaryproject/notation-go/blob/f35c52efb5cc24ba8dc6371b95f69098ff403987/verification/store.go#L23 instead?

That function also performs all the validations on the trust store

cmd/notation/cert.go
```diff
@@ -47,156 +59,296 @@ func certAddCommand(opts *certAddOpts) *cobra.Command {
...
+			opts.path = args
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return addCert(opts)
+			return addCerts(opts)
 		},
 	}
-	command.Flags().StringVarP(&opts.name, "name", "n", "", "certificate name")
+	command.Flags().StringVarP(&opts.storeType, "type", "t", "", "specify trust store type, options: ca, signingAuthority")
```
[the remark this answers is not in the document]
patrickzheng200: Updated `notation cert list` command to use TrustStorePrefixes. For other commands, storeType is a flag set in command line.

cmd/notation/cert.go
```diff
@@ -47,156 +59,296 @@ func certAddCommand(opts *certAddOpts) *cobra.Command {
...
+	command.Flags().StringVarP(&opts.namedStore, "store", "s", "", "specify named store")
 	return command
 }
+
+func certShowCommand(opts *certShowOpts) *cobra.Command {
+	if opts == nil {
+		opts = &certShowOpts{}
+	}
+	command := &cobra.Command{
+		Use:   "show --type <type> --store <name> [flags] <cert_fileName>",
```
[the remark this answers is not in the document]
patrickzheng200: The `<cert_fileName>` aligns with the [spec](https://github.com/notaryproject/notation/blob/main/specs/commandline/certificate.md#show-details-of-a-certain-certificate-file).

cmd/notation/cert.go
```diff
@@ -47,156 +59,296 @@ func certAddCommand(opts *certAddOpts) *cobra.Command {
...
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return generateTestCert(opts)
 		},
 	}
 
-	command.Flags().StringVarP(&opts.name, "name", "n", "", "key and certificate name")
 	command.Flags().IntVarP(&opts.bits, "bits", "b", 2048, "RSA key bits")
-	command.Flags().BoolVar(&opts.trust, "trust", false, "add the generated certificate to the verification list")
+	command.Flags().BoolVar(&opts.trust, "trust", false, "add the generated certificate to the trust store")
```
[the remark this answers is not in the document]
patrickzheng200: With --trust set, the certificate will be saved at truststore/x509/ca/<common_name>.
The private key is saved at notation/localkeys folder.
(Note, this command is for testing only.)

internal/certificate/cert.go
patrickzheng200: These two functions have fundamental differences. [LoadX509TrustStore](https://github.com/notaryproject/notation-go/blob/f35c52efb5cc24ba8dc6371b95f69098ff403987/verification/store.go#L48) will returns an error if joinedPath is not a regular file. However, in `notation cert list`, we won't show an eror in this situation. Instead, it would only list regular files, if there's no such file, `notation cert list` won't print anything in the command line.

cmd/notation/cert.go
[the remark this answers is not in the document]
patrickzheng200: > 1. Can we add these options to the usage text on L#156
> 2. Which trust store will this command save the certificate to?
> 3. Where will the private key file be written to?

The usage text's `[flags]` semantically covers all the optional flags available under this command. @yizha1 Shall we explicitly write out `--trust` in the `Use` line?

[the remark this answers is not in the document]
yizha1: Is this the output message of this command?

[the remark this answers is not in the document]
patrickzheng200: > Is this the output message of this command?

This is the description displayed when `notation cert generate-test --help` is called.

internal/osutil/file.go
```diff
@@ -36,3 +39,92 @@ func WriteFileWithPermission(path string, data []byte, perm fs.FileMode, overwri
...
+
+	source, err := os.Open(src)
+	if err != nil {
+		return 0, err
+	}
+	defer source.Close()
+
+	if err := os.MkdirAll(dst, 0700); err != nil {
```
shizhMSFT, line 60: We can set it to `0777` and then it will be `&` with [umask](https://en.wikipedia.org/wiki/Umask).
```suggestion
	if err := os.MkdirAll(dst, 0777); err != nil {
```

internal/osutil/file.go
```diff
@@ -36,3 +39,92 @@ func WriteFileWithPermission(path string, data []byte, perm fs.FileMode, overwri
 	}
 	return file.Close()
 }
+
+// Copy the src file to dst. Existing file will be overwritten.
+func Copy(src, dst string) (int64, error) {
```
shizhMSFT, line 44: This `Copy` is not a generic file copy. Consider naming it to `CopyToDir`.

internal/osutil/file.go
```diff
@@ -36,3 +39,92 @@ func WriteFileWithPermission(path string, data []byte, perm fs.FileMode, overwri
...
+			fmt.Println(p)
+		}
+	}
+	if len(failure) != 0 {
+		fmt.Println("Failed to delete following certificates from Trust Store:")
+
+		for ind := range failure {
+			fmt.Printf("%s, with error \"%s\"\n", failure[ind], errorSlice[ind])
+		}
+	}
```
shizhMSFT, line 127: Common packages should not contain UI code.

internal/osutil/file.go
```diff
@@ -36,3 +39,92 @@ func WriteFileWithPermission(path string, data []byte, perm fs.FileMode, overwri
...
+	destination, err := os.Create(certFile)
+	if err != nil {
+		return 0, err
+	}
+	defer destination.Close()
+	return io.Copy(destination, source)
+}
+
+// CleanDir removes all regular files under dir
+func CleanDir(dir string) error {
```
shizhMSFT, line 73: Should we just call `os.RemoveAll()`?

cmd/notation/cert.go
```diff
@@ -3,42 +3,55 @@ package main
...
 
 type certGenerateTestOpts struct {
 	name      string
 	bits      int
 	trust     bool
-	hosts     []string
 	isDefault bool
 }
 
 func certCommand() *cobra.Command {
```
shizhMSFT, line 47: The cert command is getting complex. Can we follow the same pattern as `oras` like `cert/add.go`, `cert, show.go`, etc..

shizhMSFT (changes requested):

cmd/notation/cert_gen.go
```diff
@@ -12,16 +12,13 @@ import (
...
 	"github.com/notaryproject/notation-go/dir"
 	"github.com/notaryproject/notation/internal/osutil"
+	"github.com/notaryproject/notation/internal/truststore"
 	"github.com/notaryproject/notation/pkg/configutil"
 )
 
 func generateTestCert(opts *certGenerateTestOpts) error {
 	// initialize
-	hosts := opts.hosts
 	name := opts.name
```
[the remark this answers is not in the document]
patrickzheng200: Nice catch!

internal/truststore/truststore.go
```diff
@@ -0,0 +1,216 @@
+package truststore
```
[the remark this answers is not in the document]
patrickzheng200: Moved into cmd/notation/internal/truststore

internal/truststore/truststore.go
```diff
@@ -0,0 +1,216 @@
...
+
+// AddCertCore adds a single cert file at path to the trust store
+// under dir truststore/x509/storeType/namedStore
+func AddCertCore(path, storeType, namedStore string, display bool) error {
+	// initialize
+	certPath, err := filepath.Abs(path)
+	if err != nil {
+		return err
+	}
+	storeType = strings.TrimSpace(storeType)
```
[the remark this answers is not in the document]
patrickzheng200: The trim space is for input that contains only whitespaces. If I remove the trim space, this command, 
`./notation cert add -t "          " -s Test1 ../../internal/testdata/GlobalSign.der `
will bypass the storeType == "" check. This is similar to other trim space scenarios.

internal/truststore/truststore.go
```diff
@@ -0,0 +1,216 @@
...
+		if err != nil {
+			return err
+		}
+		if info.Mode().IsRegular() {
+			if _, err := corex509.ReadCertificateFile(path); err != nil {
+				return err
+			}
+			fmt.Println(path)
+		}
+		return nil
```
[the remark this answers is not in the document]
patrickzheng200: > `/sub-dir` should be ignored. That is, the depth of walk is limited.
> 
> ```
> $XDG_CONFIG_HOME/notation/trust-store
>     /x509
>         /ca
>             /acme-rockets
>                 cert1.pem
>                 cert2.pem
>                   /sub-dir       # sub directory is ignored
>                     cert-3.pem   # certs under sub directory is ignored
>             /acme-rockets-ca2
>                 cert1.pem
>             /wabbit-networks
>                 cert3.crt
>         /tsa
>             /publicly-trusted-tsa
>                 tsa-cert1.pem
> ```

Yes, that's correct. The current implementation already does so by only listing regular files under root.

internal/truststore/truststore.go
```diff
@@ -0,0 +1,216 @@
...
+	h := sha1.Sum(cert.Raw)
+	fmt.Println("Thumbprints:", hex.EncodeToString(h[:]))
+}
+
+// DeleteAllCerts deletes all certificate files from the trust store
+// under dir truststore/x509/storeType/namedStore
+func DeleteAllCerts(storeType, namedStore string, confirmed bool, errorSlice []error) []error {
+	path, err := dir.Path.UserConfigFS.GetPath(dir.TrustStoreDir, "x509", storeType, namedStore)
+	if err == nil {
+		prompt := fmt.Sprintf("Are you sure you want to remove all certificate files under dir: %q?", path)
```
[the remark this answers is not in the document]
patrickzheng200: Updated to just show truststore/x509/<storeType>/<namedStore>

internal/truststore/truststore.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+	} else {
+		errorSlice = append(errorSlice, fmt.Errorf("%s with error %q", path, err.Error()))
+	}
+
+	return errorSlice
+}
+
+// CheckError returns nil when no err or err is fs.ErrNotExist
+func CheckError(err error) error {
```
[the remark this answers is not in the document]
patrickzheng200: Any suggestion on this one? One option might be CheckNonErrNotExistError, but this looks a bit too verbose?

cmd/notation/cert_test.go
```diff
@@ -1,99 +0,0 @@
-package main
```
rgnote, line 1: Any reason for not having any unit tests in this PR?

rgnote: Apart from not having the unit tests, rest all looks good to me.

patrickzheng200: > Any reason for not having any unit tests in this PR?

Oh yeah, unit tests are in a following PR. I've also updated the comment of this PR.

patrickzheng200: > Please post the PR link here once it is ready.

@rgnote FYI, I just added those unit tests in this PR as well. PTAL.

cmd/notation/internal/truststore/truststore.go
```diff
@@ -0,0 +1,226 @@
...
+// CheckNonErrNotExistError returns nil when no err or err is fs.ErrNotExist
+func CheckNonErrNotExistError(err error) error {
+	if err != nil && !errors.Is(err, fs.ErrNotExist) {
+		return err
+	}
+	return nil
+}
+
+// ValidateStoreType checks if storeType is supported
+func ValidateStoreType(storeType string) bool {
```
priteshbandi, line 219:
```suggestion
func isValidStoreType(storeType string) bool {
```

cmd/notation/cert/add.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	if len(failure) != 0 {
+		fmt.Printf("Failed to add following certificates to named store %s of type %s:\n", namedStore, storeType)
+
+		for ind := range failure {
+			fmt.Printf("%s, with error %q\n", failure[ind], errorSlice[ind])
+		}
+	}
+
+	return nil
```
priteshbandi, line 80: In case where we failed to add certificate, we should return non-zero exit code.  Applicable to multiple CLIs

cmd/notation/internal/truststore/truststore.go
```diff
@@ -0,0 +1,226 @@
...
+	trustStorePath, err := dir.Path.UserConfigFS.GetPath(dir.TrustStoreDir, "x509", storeType, namedStore)
+	if err := CheckNonErrNotExistError(err); err != nil {
+		return err
+	}
+	// check if certificate already in the trust store
+	if _, err := os.Stat(filepath.Join(trustStorePath, filepath.Base(certPath))); err == nil {
+		return errors.New("certificate already exists in the Trust Store")
+	}
+	// add cert to trust store
+	_, err = osutil.CopyToDir(certPath, trustStorePath)
```
priteshbandi, line 88: We should never override existing cert unless explicitly specified by user.

internal/osutil/file.go
```diff
@@ -36,3 +38,32 @@ func WriteFileWithPermission(path string, data []byte, perm fs.FileMode, overwri
...
+		return 0, fmt.Errorf("%s is not a regular file", src)
+	}
+
+	source, err := os.Open(src)
+	if err != nil {
+		return 0, err
+	}
+	defer source.Close()
+
+	if err := os.MkdirAll(dst, 0777); err != nil {
```
priteshbandi, line 59: Should we restrict permission to current user?

internal/osutil/file.go
```diff
@@ -36,3 +38,32 @@ func WriteFileWithPermission(path string, data []byte, perm fs.FileMode, overwri
...
+		return 0, err
+	}
+	certFile := filepath.Join(dst, filepath.Base(src))
+	destination, err := os.Create(certFile)
```
priteshbandi, line 63: Should we create new file with same permission as that of source file? or restrict it to current user?

specs/commandline/certificate.md
```diff
@@ -130,7 +130,6 @@ Flags:
   -b, --bits int      RSA key bits (default 2048)
   -d, --default       mark as default
   -h, --help          help for generate-test
-  -n, --name string   key and certificate name
       --trust         add the generated certificate to the trust store
```
yizha1, line 133: Remove `--trust` flag for generate-test command, and the generated certificate should be added into trust store by default, otherwise user need to get the certificate file path and use `notation cert add` to add the certificate file to trust store, which is not cumbersome.

specs/commandline/certificate.md
```diff
@@ -130,7 +130,6 @@ Flags:
   -b, --bits int      RSA key bits (default 2048)
   -d, --default       mark as default
```
yizha1, line 131: mark as default signing key

cmd/notation/internal/truststore/truststore.go
```diff
@@ -0,0 +1,226 @@
...
+// showCert displays details of a certificate
+func showCert(cert *x509.Certificate) {
+	fmt.Println("Issuer:", cert.Issuer)
+	fmt.Println("Subject:", cert.Subject)
+	fmt.Println("Valid from:", cert.NotBefore)
+	fmt.Println("Valid to:", cert.NotAfter)
+	fmt.Println("IsCA:", cert.IsCA)
+
+	h := sha1.Sum(cert.Raw)
+	fmt.Println("Thumbprints:", hex.EncodeToString(h[:]))
```
priteshbandi, line 148:
```suggestion
	fmt.Println("SHA1 Thumbprint:", strings.ToLower(hex.EncodeToString(h[:])))
```

cmd/notation/internal/truststore/truststore.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	return errorSlice
+}
+
+// DeleteCert deletes a specific certificate file from the
+// trust store, namely truststore/x509/storeType/namedStore/cert
+func DeleteCert(storeType, namedStore, cert string, confirmed bool, errorSlice []error) []error {
+	path, err := dir.Path.UserConfigFS.GetPath(dir.TrustStoreDir, "x509", storeType, namedStore, cert)
+	if err == nil {
+		truststorePath := fmt.Sprintf("truststore/x509/%s/%s/%s", storeType, namedStore, cert)
```
priteshbandi, line 185: truststorePath is not a valid path for windows because widows uses backslashes while Mac and Linux uses slashes.

cmd/notation/cert/delete.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	var errorSlice []error
+
+	if opts.all {
+		// Delete all certificates under storeType/namedStore
+		errorSlice = truststore.DeleteAllCerts(storeType, namedStore, opts.confirmed, errorSlice)
+
+		// write out on failure
+		if len(errorSlice) > 0 {
+			fmt.Println("Failed to clear following named stores:")
```
priteshbandi, line 67:
```suggestion
			fmt.Println("Failed to delete following named stores:")
```

cmd/notation/internal/truststore/truststore.go
```diff
@@ -0,0 +1,226 @@
...
+	fmt.Println("Thumbprints:", hex.EncodeToString(h[:]))
+}
+
+// DeleteAllCerts deletes all certificate files from the trust store
+// under dir truststore/x509/storeType/namedStore
+func DeleteAllCerts(storeType, namedStore string, confirmed bool, errorSlice []error) []error {
+	path, err := dir.Path.UserConfigFS.GetPath(dir.TrustStoreDir, "x509", storeType, namedStore)
+	if err == nil {
+		truststorePath := fmt.Sprintf("truststore/x509/%s/%s", storeType, namedStore)
+		prompt := fmt.Sprintf("Are you sure you want to remove all certificate files under: %q?", truststorePath)
```
priteshbandi, line 157: Also since we are managing trust-store can we obfuscate the trust store location from user? And show trust store location only when there is an error.
```suggestion
		prompt := fmt.Sprintf("Are you sure you want to remove all certificate for %s of type %s?", namedStore, storeType)
```

cmd/notation/internal/truststore/truststore.go
```diff
@@ -0,0 +1,226 @@
...
+		}
+		if !confirmed {
+			return errorSlice
+		}
+
+		if err = os.RemoveAll(path); err != nil {
+			errorSlice = append(errorSlice, fmt.Errorf("%s with error %q", path, err.Error()))
+		} else {
+			// write out on success
+			fmt.Printf("Successfully cleared %s\n", path)
```
priteshbandi, line 171: ditto.

cmd/notation/cert/delete.go
```diff
@@ -0,0 +1,91 @@
...
+
+	// Delete a certain certificate with path storeType/namedStore/cert
+	cert := opts.cert
+	if cert == "" {
+		return errors.New("to delete a specific certificate, certificate fileName cannot be empty")
+	}
+	errorSlice = truststore.DeleteCert(storeType, namedStore, cert, opts.confirmed, errorSlice)
+	// write out on failure
+	if len(errorSlice) > 0 {
+		fmt.Println("Failed to delete following certificates:")
```
priteshbandi, line 84:
```suggestion
		fmt.Println("Failed to delete following certificate:")
```

cmd/notation/cert/generateTest.go
```diff
@@ -68,19 +110,17 @@ func generateTestCert(opts *certGenerateTestOpts) error {
...
 		return err
 	}
 	trust := opts.trust
+	// Add to the trust store
 	if trust {
-		if err := addCertCore(cfg, name, certPath); err != nil {
+		if err := truststore.AddCert(certPath, "ca", name, true); err != nil {
 			return err
 		}
 	}
```
priteshbandi, line 123:
```suggestion
            		fmt.Printf("%s: added to the trust-store %s of type ca", name)
	}
```

cmd/notation/cert/show.go
```diff
@@ -0,0 +1,78 @@
...
+func certShowCommand(opts *certShowOpts) *cobra.Command {
+	if opts == nil {
+		opts = &certShowOpts{}
+	}
+	command := &cobra.Command{
+		Use:   "show --type <type> --store <name> [flags] <cert_fileName>",
+		Short: "Show certificate details given trust store type, named store, and certificate file name. If the certificate file contains multiple certificates, then all certificates are displayed.",
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) == 0 {
+				return errors.New("missing certificate path")
```
priteshbandi, line 28:
```suggestion
				return errors.New("missing certificate file name")
```

cmd/notation/cert/show.go
```diff
@@ -0,0 +1,78 @@
...
+			}
+			if len(args) > 1 {
+				return errors.New("show only supports single certificate file")
+			}
```
priteshbandi, line 32: user can still hit this condition if they specify more than 2 cert files

cmd/notation/cert/show.go
```diff
@@ -0,0 +1,78 @@
...
+	path, err := dir.Path.UserConfigFS.GetPath(dir.TrustStoreDir, "x509", storeType, namedStore, cert)
+	if err != nil {
+		return fmt.Errorf("failed to show details of certificate %s, with error: %s", cert, err.Error())
+	}
+	certs, err := corex509.ReadCertificateFile(path)
+	if err != nil {
+		return fmt.Errorf("failed to show details of certificate %s, with error: %s", cert, err.Error())
+	}
+	if len(certs) == 0 {
+		return fmt.Errorf("failed to show details of certificate %s, with error: no valid certificate presents", cert)
```
priteshbandi, line 71:
```suggestion
		return fmt.Errorf("failed to show details of certificate %s, with error: certificate not found", cert)
```

cmd/notation/internal/truststore/truststore.go
```diff
@@ -0,0 +1,226 @@
...
+			}
+			fmt.Println(path)
+		}
+		return nil
+	})
+}
+
+// ShowCerts writes out details of certificates
+func ShowCerts(certs []*x509.Certificate) {
+	fmt.Println("Display certificate details. Starting from leaf certificate if it's a certificate chain.")
```
priteshbandi, line 129: We are not validating the `certs` array is a certificate chain.
```suggestion
	fmt.Println("Certificate details")
```

cmd/notation/cert/add.go
```diff
@@ -0,0 +1,81 @@
...
+	storeType := opts.storeType
+	if storeType == "" {
+		return errors.New("store type cannot be empty")
+	}
+	if !truststore.ValidateStoreType(storeType) {
+		return fmt.Errorf("unsupported store type: %s", storeType)
+	}
+	namedStore := opts.namedStore
+	if namedStore == "" {
+		return errors.New("named store cannot be empty")
```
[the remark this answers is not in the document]
patrickzheng200: Do you mean values with multiple whitespaces only? The spec doesn't say we must ban such name for a named store.

cmd/notation/internal/truststore/truststore.go
patrickzheng200: Will rename to IsValidStoreType, this func needs to be exported as it is in the truststore package. The cert package needs to use it.

cmd/notation/internal/truststore/truststore.go
patrickzheng200: That's correct. Note, code right above this logic checks existence first. If a cert fileName is already in the specified truststore/x509/storeType/namedStore, `return errors.New("certificate already exists in the Trust Store")`.

internal/osutil/file.go
patrickzheng200: Note the `os.MkdirAll` FileMode is before unmask. The default Unix permission set for newly created directories is 777 (rwxrwxrwx) masked by the permission bits set in the umask of the process. So I think we should follow the default of Unix here?

internal/osutil/file.go
patrickzheng200: I'm not sure. Is it better to maintain the trust store if all files in it have the same permission by default? Same as above, `os.Create` will set the file permission to 0666 before unmask which is the default behavior of Unix system.

cmd/notation/cert/add.go
patrickzheng200: > In case where we failed to add certificate, we should return non-zero exit code. Comment applicable to all cert commands

@priteshbandi Could you provide an example or a suggestion of change here?

specs/commandline/certificate.md
patrickzheng200: Good point!

cmd/notation/internal/truststore/truststore.go
```diff
@@ -0,0 +1,226 @@
...
+	x509.KeyUsageKeyAgreement:      "Key Agreement",
+	x509.KeyUsageCertSign:          "Certificate Sign",
+	x509.KeyUsageCRLSign:           "CRL Sign",
+	x509.KeyUsageEncipherOnly:      "Encipher Only",
+	x509.KeyUsageDecipherOnly:      "Decipher Only",
+}
+
+// ExtKeyUsagesNameMap is a map of x509.Certificate ExtKeyUsages map used
+// in 'notation cert show' command
+var ExtKeyUsagesNameMap = map[x509.ExtKeyUsage]string{
```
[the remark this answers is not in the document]
patrickzheng200: Nice catch! Will remove it.

cmd/notation/internal/truststore/truststore.go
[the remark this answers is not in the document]
patrickzheng200: This is just a prompt while deleting certificates, listing all certificates will be quite verbose here. Besides, listing certificates is `notation cert list` 's job. I think if a user wants to check which certs are going to be deleted, he should run the list command first.

cmd/notation/internal/truststore/truststore.go
patrickzheng200: Oh yeah, truststorePath is a concise path and only used for display purpose. In other words, it's only used during printing to the command line prompt. To operate on OS level, the correct system path in line 183 is used.

cmd/notation/internal/truststore/truststore.go
patrickzheng200: Sure.

cmd/notation/cert/show.go
patrickzheng200: > user can still hit this condition if they specify more than 2 cert files

I'm not sure what do you mean here? By design, `notation cert show` only supports single cert file argument.

cmd/notation/cert/generateTest.go
patrickzheng200: We don't need this line. When `truststore.AddCert` is called, upon successful, `fmt.Printf("Successfully added %s to named store %s of type %s\n", filepath.Base(certPath), namedStore, storeType)` will do the work.

cmd/notation/cert/add.go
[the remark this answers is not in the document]
priteshbandi: IMO we should restrict most, if not all, the user-inputs to the following regex `[a-zA-Z0-9_.-]+` i.e. alphanumeric and `_`, `-`.

This will avoid most of the [cross-platform issues](https://superuser.com/questions/358855/what-characters-are-safe-in-cross-platform-file-names-for-linux-windows-and-os)

cmd/notation/cert/add.go
priteshbandi: For example, user tries to add multiple certificates to trust store using the ` notation certificate add --type ca --store some-store /temp/present.crt /temp/absent.crt` command and one of the certificate(`absent.crt`) doesn't exist 

Since the command will fail to add `absent.crt`, the exit code for the command should be non-zero. Also, the error messages related to failed portion of the command i.e. addition of  `absent.crt` to trust store should be published stderr instead of stdin

internal/osutil/file.go
priteshbandi: Do we need to provide read/write permissions to group and public? IMO to start with we can provide read permission but not write permission to group and public.

cc: @shizhMSFT

internal/osutil/file.go
priteshbandi: Here I would recommend restring it to `644(rw-r--r--)`

cmd/notation/cert/add.go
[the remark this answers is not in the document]
patrickzheng200: > IMO we should restrict most, if not all, the user-inputs to the following regex `[a-zA-Z0-9_.-]+` i.e. alphanumeric and `_`, `-`.
> 
> This will avoid most of the [cross-platform issues](https://superuser.com/questions/358855/what-characters-are-safe-in-cross-platform-file-names-for-linux-windows-and-os). Also, we can either update spec with this PR or later(please open a issue, if we choose to do it later).

The underlying [dir](https://github.com/notaryproject/notation-go/blob/bccecbc83094bffc7c942c16adcdd871bec1f940/dir/fs.go#L85) package in notation-go already utilizes the os standard library to validate the file name.

cmd/notation/cert/add.go
patrickzheng200: Sure. Updated the code.

internal/osutil/file.go
patrickzheng200: code updated.

internal/osutil/file.go
patrickzheng200: code updated.

cmd/notation/cert/add.go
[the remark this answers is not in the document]
priteshbandi: yes but if a user is executing notation verification on multiple operating system, a trust store name like `"` would work on linux but not on windows.

Is there any reason we want to allow all possible characters? Also, not constraining input means we need to have exhaustive test and execute them on multiple operating system.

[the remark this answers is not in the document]
priteshbandi: Example:
```
➜  notation git:(truststore) ✗ color ./notation cert add /Users/pritesh/Desktop/digi-root.pem --store "/" --type ca
Successfully added following certificates to named store / of type ca:
/Users/pritesb/Desktop/digi-root.pem
➜  notation git:(truststore) ✗ color ./notation cert list                                                           
/Users/pritesb/Library/Application Support/notation/truststore/x509/ca/digi-root.pem
```

cmd/notation/cert/add.go
priteshbandi: Looks like now we are emitting error messages twice:

https://ctxt.io/2/AAAQ8Ra_Ew

cmd/notation/cert/add.go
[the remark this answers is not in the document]
patrickzheng200: code updated.

cmd/notation/cert/add.go
patrickzheng200: > Looks like now we are emitting error messages twice:
> 
> https://ctxt.io/2/AAAQ8Ra_Ew

Nice catch! Code updated.

priteshbandi (approved):
LGTM

cmd/notation/cert/delete.go
```diff
@@ -0,0 +1,93 @@
...
+		return errors.New("store type cannot be empty")
+	}
+	if !truststore.IsValidStoreType(storeType) {
+		return fmt.Errorf("unsupported store type: %s", storeType)
+	}
+	var errorSlice []error
+
+	if opts.all {
+		// Delete all certificates under storeType/namedStore
+		errorSlice = truststore.DeleteAllCerts(storeType, namedStore, opts.confirmed, errorSlice)
```
shizhMSFT, line 63: `errorSlice` is `[]error`. We can define a type `type Errors []error` and implement `func (errs Error) Error string()` to reduce duplicated code. This can be done later in the `notation` CLI code refactoring. Of course, we need an issue to track.

shizhMSFT (changes requested):

cmd/notation/cert/delete.go
patrickzheng200: Created an [issue](https://github.com/notaryproject/notation/issues/426) to track this.

shizhMSFT (approved):
LGTM

PR #413, opened 2022-10-25
feat: implement login/logout UX

JeyJeyGao:
implemnt login UI based on https://github.com/notaryproject/notation/blob/main/specs/commandline/login.md fix:
1. added `Login Succeeded` message when login succeeded
2. added username parameter validation
3. removed --plain-http global flag
4. updated spec: removed logout --plain-http & --all flag

Test:
1. login with -u -p flags
2. login with $NOTATION_USERNAME $NOTATION_PASSWORD
3. login with --password-stdin
5. login without username -> `username was not set.`

### DIFF
--- a/cmd/notation/login.go
+++ b/cmd/notation/login.go
@@ -47,7 +47,7 @@ Example - Login using $NOTATION_USERNAME $NOTATION_PASSWORD variables:
 
 			// check username
 			if opts.Username == "" {
-				return errors.New("username was not set.")
+				return errors.New("username is not set.")
 			}
 			return nil
 		},
--- a/cmd/notation/login.go
+++ b/cmd/notation/login.go
@@ -44,11 +44,6 @@ Example - Login using $NOTATION_USERNAME $NOTATION_PASSWORD variables:
 			if err := readPassword(opts); err != nil {
 				return err
 			}
-
-			// check username
-			if opts.Username == "" {
-				return errors.New("username is not set.")
-			}
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
@@ -82,9 +77,7 @@ func runLogin(cmd *cobra.Command, opts *loginOpts) error {
 		return fmt.Errorf("failed to store credentials: %v", err)
 	}
 
-	if err == nil {
-		fmt.Println("Login Succeeded")
-	}
+	fmt.Println("Login Succeeded")
 	return nil
 }
 
--- a/cmd/notation/login.go
+++ b/cmd/notation/login.go
@@ -25,7 +25,7 @@ func loginCommand(opts *loginOpts) *cobra.Command {
 	}
 	command := &cobra.Command{
 		Use:   "login [flags] <server>",
-		Short: "Provides credentials for authenticated registry operations",
+		Short: "Login to registry",
 		Long: `Log in to an OCI registry
 
 Example - Login with provided username and password:
--- a/cmd/notation/login.go
+++ b/cmd/notation/login.go
@@ -24,9 +24,9 @@ func loginCommand(opts *loginOpts) *cobra.Command {
 		opts = &loginOpts{}
 	}
 	command := &cobra.Command{
-		Use:   "login [options] [server]",
+		Use:   "login [flags] <server>",
 		Short: "Provides credentials for authenticated registry operations",
-		Long: `notation login [options] [server]
+		Long: `Log in to an OCI registry
 
 Example - Login with provided username and password:
 	notation login -u <user> -p <password> registry.example.com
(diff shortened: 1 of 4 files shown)


cmd/notation/login.go
```diff
@@ -44,13 +44,18 @@ Example - Login using $NOTATION_USERNAME $NOTATION_PASSWORD variables:
 			if err := readPassword(opts); err != nil {
 				return err
 			}
+
+			// check username
+			if opts.Username == "" {
+				return errors.New("username was not set.")
```
chloeyin, line 50: `was not set` or `is not set`?

cmd/notation/logout.go
```diff
@@ -16,8 +16,9 @@ func logoutCommand(opts *logoutOpts) *cobra.Command {
 		opts = &logoutOpts{}
 	}
 	return &cobra.Command{
-		Use:   "logout [server]",
+		Use:   "logout [flags] [server]",
```
chloeyin, line 19: If server is an optional argument, https://github.com/notaryproject/notation/blob/f747031135a6010d92c737ea611276b18f4188d7/cmd/notation/logout.go#L21-L27 should we remove this logic?

cmd/notation/login.go
JeyJeyGao: Updated: `is not set`

cmd/notation/logout.go
JeyJeyGao: Currently, we only has a -h flag. If -h flag is specified, this logic will be skipped.

patrickzheng200, line 19:
```suggestion
		Use:   "logout [flags] <server>",
```
This doesn't look very correct to me. If server is optional, what happens when `notation logout` is called?

cmd/notation/logout.go
```diff
@@ -16,8 +16,9 @@ func logoutCommand(opts *logoutOpts) *cobra.Command {
...
 		Short: "Log out the specified registry hostname",
```
patrickzheng200, line 20:
```suggestion
		Short: "Log out from the logged in registries",
```
This is stated in the [spec](https://github.com/notaryproject/notation/blob/main/specs/notation-cli.md). Interestingly, this command's Long is shorter than Short.

patrickzheng200 (changes requested):

cmd/notation/logout.go
JeyJeyGao: Updated code and spec.

yizha1:
@dtzar  There are some cases need to be considered for `--all` flag, for example, user will log out from all the registries including those logged in using `docker login` or other tools.  There are also other cases. We need spend more time on investigation. So, I suggest removing `--all` flag for now and add it back later once we figure out the correct behaviors.

cmd/notation/login.go
```diff
@@ -75,6 +81,10 @@ func runLogin(cmd *cobra.Command, opts *loginOpts) error {
 	if err = nativeStore.Store(serverAddress, creds); err != nil {
 		return fmt.Errorf("failed to store credentials: %v", err)
 	}
+
+	if err == nil {
```
shizhMSFT, line 85: This error check seems redundant.

cmd/notation/login.go
```diff
@@ -44,13 +44,18 @@ Example - Login using $NOTATION_USERNAME $NOTATION_PASSWORD variables:
 			if err := readPassword(opts); err != nil {
 				return err
 			}
+
+			// check username
+			if opts.Username == "" {
+				return errors.New("username is not set.")
+			}
```
shizhMSFT, line 51: If username is not set, then the CLI should prompt instead of an error.
@yizha1 Does `notation` follows the same UX with `oras` and `docker`? or it has a different UX that does not support prompting?

cmd/notation/login.go
JeyJeyGao: Updated.

cmd/notation/login.go
JeyJeyGao: Removed this logic.

shizhMSFT (approved):
LGTM

priteshbandi:
> @dtzar There are some cases need to be considered for `--all` flag, for example, user will log out from all the registries including those logged in using `docker login` or other tools. There are also other cases. We need spend more time on investigation. So, I suggest removing `--all` flag for now considering the time of beta.1 release is approaching and add it back later once we figure out the correct behaviors.

Should we mention this in command help?

priteshbandi (approved):
LGTM

cmd/notation/logout.go
```diff
@@ -16,8 +16,9 @@ func logoutCommand(opts *logoutOpts) *cobra.Command {
 		opts = &logoutOpts{}
 	}
 	return &cobra.Command{
-		Use:   "logout [server]",
-		Short: "Log out the specified registry hostname",
```
priteshbandi, line 20: not: `Short` seems to be longer than `Long`.

patrickzheng200 (approved):
LGTM

cmd/notation/logout.go
JeyJeyGao: Updated: both short and long description use `Log out from the logged in registries`

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #415, opened 2022-10-26
feat: implement plugin UX

JeyJeyGao:
According https://github.com/notaryproject/notation/blob/main/specs/commandline/plugin.md.

Updated help message.
Updated spec: replace "\<command>"  with "[command]" because the format is defined in cobra template.

### DIFF
--- a/cmd/notation/plugin.go
+++ b/cmd/notation/plugin.go
@@ -10,7 +10,7 @@ import (
 
 func pluginCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "plugin",
+		Use:   "plugin <command>",
 		Short: "Manage plugins",
 	}
 	cmd.AddCommand(pluginListCommand())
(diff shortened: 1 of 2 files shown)


cmd/notation/plugin.go
```diff
@@ -10,7 +10,7 @@ import (
 
 func pluginCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "plugin",
+		Use:   "plugin [command]",
```
patrickzheng200, line 13: I think this add is not necessary? Cobra would automatically print out [command] when calling `notation plugin -h`

specs/commandline/plugin.md
```diff
@@ -12,15 +12,15 @@ Use `notation plugin` to manage plugins. See notation [plugin documentation](htt
...
+  notation plugin [command]
 
 Available Commands:
   list        List installed plugins
 
 Flags:
   -h, --help          help for plugin
 
-RUN 'notation plugin <command> --help' for more information on a command.
+Use 'notation plugin [command] --help' for more information about a command.
```
patrickzheng200, line 23: We can just remove this line as other specs do not have it.

cmd/notation/plugin.go
JeyJeyGao: Just want to make it clear, but removed now.

specs/commandline/plugin.md
JeyJeyGao: Removed this line.

jondonas (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #419, opened 2022-10-26
feat: implement version command

JeyJeyGao:
- Added version command
- Removed --version flag
- Update Makefile to set GitCommit value by ldflags
- Update the BuildMetadata definition and allow user to override the value by environment variable

Please build it by`make build` command to include git commit information

Example

```
$ ./bin/notation version
Notation: Notary v2, A tool to sign, store, and verify artifacts.

Version:     v0.11.0-alpha.4
Go version:  go1.19
Git commit:  f747031135a6010d92c737ea611276b18f4188d7
```

If it does `not` build by `make build` command, the output will be

```
$ ./bin/notation version
Notation: Notary v2, A tool to sign, store, and verify artifacts.

Version:     v0.11.0-alpha.4
Go version:  go1.19
```

### DIFF
--- a/cmd/notation/version.go
+++ b/cmd/notation/version.go
@@ -1,0 +1,32 @@
+package main
+
+import (
+	"fmt"
+	"runtime"
+
+	"github.com/notaryproject/notation/internal/version"
+	"github.com/spf13/cobra"
+)
+
+func versionCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "version",
+		Short: "Show the notation version information",
+		Args:  cobra.NoArgs,
+		Run: func(cmd *cobra.Command, args []string) {
+			runVersion()
+		},
+	}
+	return cmd
+}
+
+func runVersion() {
+	fmt.Printf("Notation: Notary v2, A tool to sign, store, and verify artifacts.\n\n")
+
+	fmt.Printf("Version:     %s\n", version.GetVersion())
+	fmt.Printf("Go version:  %s\n", runtime.Version())
+
+	if version.GitCommit != "" {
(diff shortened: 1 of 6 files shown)


cmd/notation/version.go
```diff
@@ -0,0 +1,39 @@
...
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runVersion()
+		},
+	}
+
+	return cmd
+}
+
+func runVersion() error {
```
chloeyin, line 25: error is always nil, why do we need to return it.

cmd/notation/version.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/notaryproject/notation/internal/version"
+	"github.com/spf13/cobra"
+)
+
+func versionCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "version",
+		Short: "Show the notation version information",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
```
chloeyin, line 17: If this command always returns a nil, you can use `Run` instead of `RunE`

cmd/notation/version.go
JeyJeyGao: Thanks, Updated.

cmd/notation/version.go
JeyJeyGao: Thanks updated.

rgnote (approved):

patrickzheng200 (approved):
LGTM

shizhMSFT (approved):
LGTM

Issue #410, opened 2022-10-21
When no key is present, make key add be the default key


dtzar:
If there are no keys present for notation and you successfully add a key via `notation key add <key>` it does not make this key the default key and hence when you try to sign something it gives this: `Error: signing key not found`

We should add a check that if there are no keys present, we make the key added the default key without having to pass the --default flag.

priteshbandi:
The experience is right but the error message is wrong. Error should say `default signing key not set. Please either set default singing key or specify a key name`.

yizha1:
> The experience is right but the error message is wrong. Error should say `default signing key not set. Please set default singing key or specify a key name`.

Could there be a quick fix for the error message? I am not sure if the error is printed out from CLI or library

priteshbandi:
Yes, posted PR https://github.com/notaryproject/notation/pull/432

priteshbandi:
PR had been merged, closing the issue.

PR #432, opened 2022-11-04
Improve error message when default signing key is not set

priteshbandi:
Fixes Issue: https://github.com/notaryproject/notation/issues/410

**Test output**
```
➜  notation git:(fix) ✗ ./notation key ls
NAME   KEY PATH   CERTIFICATE PATH   ID   PLUGIN NAME
➜  notation git:(fix) ✗ ./notation sign --plain-http $IMAGE
Error: default signing key not set. Please set default singing key or specify a key name
```

### DIFF
--- a/pkg/configutil/util.go
+++ b/pkg/configutil/util.go
@@ -34,16 +34,16 @@ func ResolveKey(name string) (config.KeySuite, error) {
 	if err != nil {
 		return config.KeySuite{}, err
 	}
-	
+
 	// if name is empty, look for default signing key
 	if name == "" {
 		name = signingKeys.Default
-	}
 
-	// if name is still empty, return error
-	if name == "" {
-		return config.KeySuite{}, errors.New("default signing key not set." +
-			" Please set default singing key or specify a key name")
+		// if name is still empty, return error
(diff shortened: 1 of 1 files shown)


pkg/configutil/util.go
```diff
@@ -34,9 +34,18 @@ func ResolveKey(name string) (config.KeySuite, error) {
...
+	// if name is empty, look for default signing key
 	if name == "" {
 		name = signingKeys.Default
 	}
+
+	// if name is still empty, return error
+	if name == "" {
+		return config.KeySuite{}, errors.New("default signing key not set." +
+			" Please set default singing key or specify a key name")
+	}
```
shizhMSFT, line 47:
```suggestion
	// if name is empty, look for default signing key
	if name == "" {
		name = signingKeys.Default

		// if name is still empty, return error
		if name == "" {
			return config.KeySuite{}, errors.New("default signing key not set. Please set default singing key or specify a key name")
		}
	}
```

rgnote (approved):

shizhMSFT (approved):
LGTM

Issue #438, opened 2022-11-08
Document Typical "Trust Policy" examples that customers may encounter. 


iamsamirzon:
For Notation's RC-1 release, lets create a sample of trust policy templates, that customers can use as a reference to build their trust policies.

These examples should be based on real world examples

iamsamirzon:
@toddysm  - Assigning this one to you first. We all can add other examples later.

cc: @vaninrao10

yizha1:
As discussed in the community meeting, closed this issue as the document is available at https://notaryproject.dev/docs/user-guides/tutorials/trust-policy/

PR #441, opened 2022-11-10
Added scenario for trust policy and trust store

toddysm:
Signed-off-by: Toddy Mladenov <toddysm@gmail.com>

This is related to the #438


specs/scenarios/trust-store-and-trust-policy.md
```diff
@@ -0,0 +1,212 @@
...
+$XDG_CONFIG_HOME/notation/trust-store
+    /x509
+        /ca
+            /wabbit-networks
+                wabbit-networks-kubernetes.crt
+        /tsa
+            /publicly-trusted-tsa
+                tsa-cert1.pem
+```
+However, just the *trust store* is not enough for Notation to validate the containers used for the *platform layer*. A *trust policy* needs to be established for the *platform layer* containers. Here is how the trust policy needs to look for a brand new node in a Kubernetes cluster based on the above example.
```
yizha1, line 76: Where is notation installed?

specs/scenarios/trust-store-and-trust-policy.md
```diff
@@ -0,0 +1,212 @@
...
+            ]
+        }
+    ]
+}
+```
+
+The above *trust store* and *trust policy* are the initial state for each Kubernetes node in the cluster once the node is spun up. Once again, those are baked into the VM image that is used to start the node.
+
+### Adding Trust Certificates to the Trust Store and Trust Policy for the Service Layer
+Once the node is started the Wabbit Networks Batch service needs to be deployed on the node. For that to work, the *trust store* needs to be updated with the Wabbit Networks Batch certificates and a trust policy needs to be added. Here is how the *trust store* should look like:
```
yizha1, line 100: Who will add the trust store and trust policy?

specs/scenarios/trust-store-and-trust-policy.md
```diff
@@ -0,0 +1,212 @@
...
+            "trustedIdentities": [
+              "x509.subject: C=US, ST=WA, L=Seattle, O=wabbitnetworks.com, OU=BatchTeam, CN=BatchTeamSecureBuilder"
+            ]
+        }
+    ]
+}
+```
+
+### Adding Trust Certificates to the Trust Store and Trust Policy for the Application Layer
+Once the node is started and the Wabbit Networks Batch service is deployed on the node, a customer can deploy their workloads. For that to work, the *trust store* needs to be updated with the Wabbit Networks Batch certificates and a trust policy needs to be added. Here is how the *trust store* should look like:
```
yizha1, line 146: same question as previous: Who will add the trust store and trust policy?

specs/scenarios/trust-store-and-trust-policy.md
```diff
@@ -0,0 +1,212 @@
...
+            "trustedIdentities": [
+              "x509.subject: C=US, ST=WA, L=Seattle, O=acmerockets.com, OU=MapReduceTeam, CN=MapReduceTeamSecureBuilder"
+            ]
+        }
+    ]
+}
+```
+
+#### Removing Trust Certificates from the Trust Store and the Trust Policy
+When the customer's workload is shut down, the trust established for the customer's certificate should be removed from the node. In cases as above, the customer workloads may be temporary (for example while running the map/reduce task) and once they are done, the trust to the customer certificate should be removed from the nodes. There shouldn't be expectation that a second batch from the same customer will be scheduled on the same cluster/nodes.
```
yizha1, line 205: who will delete the trust store and trust policy?

specs/scenarios/trust-store-and-trust-policy.md
toddysm: Can you elaborate how the Notation installation location is relevant for the store structure?

specs/scenarios/trust-store-and-trust-policy.md
toddysm: This is up to the tools that interact with Notation. From scenario point of view, we don't want to go into the details who this will be because there may be many tools that need to do that. We should enable tools like Ratify, Kyverno etc to do that.

specs/scenarios/trust-store-and-trust-policy.md
toddysm: Same answer as above :) This should be enabled for any tool to do those changes.

specs/scenarios/trust-store-and-trust-policy.md
toddysm: Same answer as above. Any tool should be able to do that and we should enable it through APIs and CLI commands in Notation.

specs/scenarios/trust-store-and-trust-policy.md
```diff
@@ -0,0 +1,212 @@
...
+    /x509
+        /ca
+            /wabbit-networks
+                wabbit-networks-kubernetes.crt
+                wabbit-networks-batch.crt
+        /tsa
+            /publicly-trusted-tsa
+                tsa-cert1.pem
+```
+And here is how the policy should look like before deployint the Wabbit Networks Batch service:
```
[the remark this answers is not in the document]
toddysm: Thanks :) Fixed

specs/scenarios/trust-store-and-trust-policy.md
yizha1: Yes, it is not relevant to the scenario. No need to discuss details. I am just curious about how notation is installed. Notation here could mean notation CLI or notation API.

FeynmanZhou:
Do you think we should place all generic specs in the [notaryproject]( https://github.com/notaryproject/notaryproject) repo? There is another scenario spec in that repo: https://github.com/notaryproject/notaryproject/blob/main/scenarios.md.

specs/scenarios/trust-store-and-trust-policy.md
```diff
@@ -0,0 +1,212 @@
...
+Once the node is started the Wabbit Networks Batch service needs to be deployed on the node. For that to work, the *trust store* needs to be updated with the Wabbit Networks Batch certificates and a trust policy needs to be added. Here is how the *trust store* should look like:
+
+```
+$XDG_CONFIG_HOME/notation/trust-store
+    /x509
+        /ca
+            /wabbit-networks
+                wabbit-networks-kubernetes.crt
+                wabbit-networks-batch.crt
+        /tsa
```
priteshbandi, line 109: If `wabbit-networks-kubernetes.crt` and `wabbit-networks-batch.crt are derived from same root, you can directly add that root.

specs/scenarios/trust-store-and-trust-policy.md
```diff
@@ -0,0 +1,212 @@
...
+        },
+        {
+            "name": "acme-rockets-mapreduce",
+            "registryScopes": [ "mr.acmecr.io/mr" ],          
+            "signatureVerification": {          
+              "level" : "strict" 
+            },
+            "trustStores": ["ca:acme-rockets"], 
+            "trustedIdentities": [
+              "x509.subject: C=US, ST=WA, L=Seattle, O=acmerockets.com, OU=MapReduceTeam, CN=MapReduceTeamSecureBuilder"
```
priteshbandi, line 197: just to call out if `acme` is using the root `acme-rockets-mr.crt` only for signing application images, then user can just have `"trustedIdentities": ["*"]`

priteshbandi (commented):
It's a good doc, added some nitpicks

specs/scenarios/trust-store-and-trust-policy.md
```diff
@@ -0,0 +1,212 @@
...
+
+| layer | description | example |
+| ----- | ----- | ----- |
+| application | Customer application deployed in the form of containers using the services provided in the previous layer | Map-Reduce Application|
+| service | Containerized services deployed on the platform | Batch Service |
+| platform | Infrastructure management platform allowing elastic expansion of the workloads | Kubernetes | 
+
+To further clarify the distinction between the service workloads and the application workaloads, we can use the following diagram.
+
+![](../../media/trust-store-scenarios-app-stack.svg)
```
FeynmanZhou, line 32: Is [Kubernetes Control Plane](https://kubernetes.io/docs/concepts/overview/components/#control-plane-components) more commonly-used than *Kubernetes Management Plane*?

toddysm: I don't know :) I am used to "control plane"

specs/scenarios/trust-store-and-trust-policy.md
toddysm: That is correct. Though, I normally question the security to trust broad roots.

PR #440, opened 2022-11-08
spec: update cli verify spec for UX improvement

yizha1:
Update cli verify spec for UX improvement

Signed-off-by: Yi Zha <yizha1@microsoft.com>

PR #439, opened 2022-11-08
spec: update cli sign spec for tag to digest translation

yizha1:
This PR is mainly to improve the output message of `notation sign` command for tag to digest translation.

PR #450, opened 2022-11-22
update: updated CLI outputs of sign/verification

patrickzheng200:
This PR updates outputs of sign/verification based on changes in PR #439 and PR #440 specs.
Previous to this PR, outputs only print out digest.
In this PR:
1. warnings will be printed out for tag reference during sign/verification. 
2. outputs will print out in format `<registry>/<repository>@<digest>`.


rgnote (approved):
LGTM

cmd/notation/registry.go
```diff
@@ -5,7 +5,7 @@ import (
 	"errors"
 	"net"
 
-	notationregistry "github.com/notaryproject/notation-go/registry"
+	notationRegistry "github.com/notaryproject/notation-go/registry"
```
shizhMSFT, line 8: By convention, we use all lower letters for package names.

cmd/notation/sign.go
```diff
@@ -88,30 +88,39 @@ func runSign(command *cobra.Command, cmdOpts *signOpts) error {
 	}
 
 	// write out
-	fmt.Println(desc.Digest)
+	fmt.Println("Successfully signed", ref.String())
```
shizhMSFT, line 91: nit:
```suggestion
	fmt.Println("Successfully signed", ref)
```

cmd/notation/manifest.go
```diff
@@ -8,22 +8,26 @@ import (
...
-		return ocispec.Descriptor{}, err
+		return ocispec.Descriptor{}, registry.Reference{}, err
 	}
 	repo, err := getRepositoryClient(opts, ref)
 	if err != nil {
-		return ocispec.Descriptor{}, err
+		return ocispec.Descriptor{}, registry.Reference{}, err
 	}
-	return repo.Resolve(ctx, ref.ReferenceOrDefault())
+	manifestDesc, err := repo.Resolve(ctx, ref.ReferenceOrDefault())
```
shizhMSFT, line 28: We might want to discourage the use of default `latest`. /cc @yizha1 @FeynmanZhou 
Note that `oras` CLI requires the customer to input the tag or the digest. Empty reference means errors.
Meanwhile, `docker` will use the default `latest` tag for empty reference.

@patrickzheng200 if we still want the default `latest`. We should do
```suggestion
	ref.Reference = ref.ReferenceOrDefault()
	manifestDesc, err := repo.Resolve(ctx, ref.Reference)
```

cmd/notation/verify.go
```diff
@@ -2,15 +2,13 @@ package main
...
 import (
 	"errors"
+	"fmt"
 	"math"
-	"os"
-	"strings"
 
 	"github.com/notaryproject/notation-go"
-	notationregistry "github.com/notaryproject/notation-go/registry"
+	notationRegistry "github.com/notaryproject/notation-go/registry"
```
shizhMSFT, line 9: same here

cmd/notation/verify.go
```diff
@@ -90,11 +88,27 @@ func runVerify(command *cobra.Command, opts *verifyOpts) error {
...
+	}
+
+	// on success
+	outcome := outcomes[0]
+	// print out warning for any failed result with logged verification action
+	for _, result := range outcome.VerificationResults {
+		if result.Error != nil {
+			// at this point, the verification action has to be logged and
+			// it's failed
+			fmt.Printf("warning: %v was set to \"logged\" and failed with error: %v\n", result.Type, result.Error)
```
shizhMSFT, line 107:
```suggestion
			fmt.Printf("Warning: %v was set to \"logged\" and failed with error: %v\n", result.Type, result.Error)
```

cmd/notation/manifest.go
yizha1: I vote for "empty means errors, tag or digest is a MUST"

patrickzheng200: Added check to return error if tag or digest is missing.

FeynmanZhou: @patrickzheng200 @shizhMSFT  Agree with avoiding using the default `lastest` if users don't specify a tag or digest. The registry side doesn’t tag anything automatically, so we need to prompt users explicitly on the client side.

patrickzheng200: > @patrickzheng200 @shizhMSFT Agree with avoiding using the default `lastest` if users don't specify a tag or digest. The registry side doesn’t tag anything automatically, so we need to prompt users explicitly on the client side.

updated code to reflect this.

cmd/notation/sign.go
```diff
@@ -88,30 +88,39 @@ func runSign(command *cobra.Command, cmdOpts *signOpts) error {
...
 	if err != nil {
-		return ocispec.Descriptor{}, notation.SignOptions{}, err
+		return notation.SignOptions{}, registry.Reference{}, err
 	}
 	pluginConfig, err := cmd.ParseFlagPluginConfig(opts.pluginConfig)
 	if err != nil {
-		return ocispec.Descriptor{}, notation.SignOptions{}, err
+		return notation.SignOptions{}, registry.Reference{}, err
+	}
+	if ref.ValidateReferenceAsDigest() != nil {
```
shizhMSFT, line 109: nit: code style
```suggestion
	if err := ref.ValidateReferenceAsDigest(); err != nil {
```

cmd/notation/verify.go
```diff
@@ -103,26 +117,19 @@ func resolveReference(command *cobra.Command, opts *verifyOpts) (registry.Refere
 		return registry.Reference{}, err
 	}
 
-	if isDigestReference(opts.reference) {
+	// reference is a digest reference
+	if ref.ValidateReferenceAsDigest() == nil {
```
shizhMSFT, line 121: nit: code style
```suggestion
	if err := ref.ValidateReferenceAsDigest(); err == nil {
```

cmd/notation/verify.go
```diff
@@ -103,26 +117,19 @@ func resolveReference(command *cobra.Command, opts *verifyOpts) (registry.Refere
...
 		return ref, nil
 	}
 
-	// Resolve tag reference to digest reference.
-	manifestDesc, err := getManifestDescriptorFromReference(command.Context(), &opts.SecureFlagOpts, opts.reference)
+	// resolve tag to digest reference
+	manifestDesc, ref, err := getManifestDescriptorFromReference(command.Context(), &opts.SecureFlagOpts, opts.reference)
```
shizhMSFT, line 126: empty reference (i.e. no tag and not digest) is not checked.

cmd/notation/manifest.go
```diff
@@ -8,22 +8,30 @@ import (
...
 	if ref == "" {
-		return ocispec.Descriptor{}, errors.New("missing reference")
+		return ocispec.Descriptor{}, registry.Reference{}, errors.New("missing reference")
 	}
 
 	return getManifestDescriptorFromReference(ctx, opts, ref)
 }
 
-func getManifestDescriptorFromReference(ctx context.Context, opts *SecureFlagOpts, reference string) (ocispec.Descriptor, error) {
+func getManifestDescriptorFromReference(ctx context.Context, opts *SecureFlagOpts, reference string) (ocispec.Descriptor, registry.Reference, error) {
```
shizhMSFT, line 19: looks like those functions can be combined as `getManifestDescriptor()`.

cmd/notation/sign.go
```diff
@@ -88,30 +88,39 @@ func runSign(command *cobra.Command, cmdOpts *signOpts) error {
...
+		// reference is not a digest reference
+		fmt.Printf("Warning: Always sign the artifact using digest(`@sha256:...`) rather than a tag(`:%s`) because tags are mutable and a tag reference can point to a different artifact than the one signed.\n", ref.Reference)
+		fmt.Printf("Resolved artifact tag `%s` to digest `%s` before signing.\n", ref.Reference, manifestDesc.Digest.String())
+
+		// resolve tag to digest reference
+		ref.Reference = manifestDesc.Digest.String()
```
shizhMSFT, line 115: This part of code is similar to the code in `resolveReference()`. Could we do a refactor?

cmd/notation/verify.go
patrickzheng200: Refactored verify.go.

cmd/notation/sign.go
patrickzheng200: Refactored manifest.go and verify.go. However, we might don't want to combine this logic in sign.go with the one in verify.go, because specs for sign and verify are different. And the printing messages are different as well.

patrickzheng200: Updated the code to use call back functions.

cmd/notation/sign.go
```diff
@@ -88,30 +89,35 @@ func runSign(command *cobra.Command, cmdOpts *signOpts) error {
 	}
 
 	// write out
-	fmt.Println(desc.Digest)
+	fmt.Println("Successfully signed", ref)
+
```
shizhMSFT, line 93: nit: empty line

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

PR #453, opened 2022-11-25
Use minimum(user only) file permissions

priteshbandi:
Since we are not implementing system config behavior in rc1, updating code to have only user(r,w,x) permission.
Also, its a good practice to use minimum permission model

Signed-off-by: Pritesh Bandi <pritesb@amazon.com>

### DIFF
--- a/internal/osutil/file.go
+++ b/internal/osutil/file.go
@@ -10,15 +10,15 @@ import (
 
 // WriteFile writes to a path with all parent directories created.
 func WriteFile(path string, data []byte) error {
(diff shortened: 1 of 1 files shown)


shizhMSFT (approved):
LGTM. We need to revisit them when we consider system config again.

shizhMSFT:
@priteshbandi Could you edit the PR tiles to conform with the conventional commit spec?

yizha1:
An example of conventional commit https://github.com/oras-project/community/blob/main/GIT_CONVENTIONS.md

rgnote (approved):
LGTM

PR #466, opened 2022-12-05
spec: use wording remove instead of delete

yizha1:
Use `notation remove` instead of `notation delete`. The reason is:
"Delete and remove are defined quite similarly, but the main difference between them is that delete means erase (i.e. rendered nonexistent or nonrecoverable), while remove connotes take away and set aside (but kept in existence)."

The signing key is still in existence after being removed from the signing key list, so `notation remove` is better.

### DIFF
--- a/specs/commandline/key.md
+++ b/specs/commandline/key.md
@@ -16,7 +16,7 @@ Usage:
 
 Available Commands:
   add         Add key to signing key list
-  delete      Delete key from signing key list
+  remove      Remove key from signing key list
   list        List keys used for signing
   update      Update key in signing key list
 
@@ -40,19 +40,19 @@ Flags:
       --plugin-config strings  {key}={value} pairs that are passed as it is to a plugin, refer plugin's documentation to set appropriate values
 ```
 
-### notation key delete
+### notation key remove
 
 ```text
-Delete key from signing key list
+Remove key from signing key list
 
 Usage:
-  notation key delete [flags] <key_name>...
+  notation key remove [flags] <key_name>...
 
 Aliases:
-  delete, rm
+  remove, rm
 
 Flags:
-  -h, --help                   help for delete
+  -h, --help                   help for remove
 
 ```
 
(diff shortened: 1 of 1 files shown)


shizhMSFT (approved):
LGTM

iamsamirzon:
@shizhMSFT , @yizha1  - Does the same logic hold for "Certificate" add/delete or add/remove ?  I think customers will look for parity in how "Certificate" add/delete and "Key" add/delete operations works.

vaninrao10:
> @shizhMSFT , @yizha1 - Does the same logic hold for "Certificate" add/delete or add/remove ? I think customers will look for parity in how "Certificate" add/delete and "Key" add/delete operations works.

I think remove vs delete makes sense, in Linux rm (remove) means delete, also for key remove will remove the reference of the signature from the list. Will we have delete in future if this is a security concern ? not sure about that.
For certificate I am not sure if we have the concept of remove, it is always delete right ?

yizha1:
> > @shizhMSFT , @yizha1 - Does the same logic hold for "Certificate" add/delete or add/remove ? I think customers will look for parity in how "Certificate" add/delete and "Key" add/delete operations works.
> 
> I think remove vs delete makes sense, in Linux rm (remove) means delete, also for key remove will remove the reference of the signature from the list. Will we have delete in future if this is a security concern ? not sure about that. For certificate I am not sure if we have the concept of remove, it is always delete right ?

For certificate added to trust store,  the command `notation cert delete` always delete it.

yizha1:
> > Use notation remove instead of notation delete. The reason is:
> > "Delete and remove are defined quite similarly, but the main difference between them is that delete means erase (i.e. rendered nonexistent or nonrecoverable), while remove connotes take away and set aside (but kept in existence)."
> > The signing key is still in existence after being removed from the signing key list, so notation remove is better.
> 
> We have same issue with remove, in linux [`rm`](https://www.geeksforgeeks.org/rm-command-linux-examples/) means removing files i.e. actually deleting file from disk. Also this brings up another question, how will this command(delete or remove) work with local keys, when we support singing with local keys? Could they be interpreted by user as deleting the keys from the disk? If so than this would be an security issue.

We didn't support local key yet, we can come up with the solution later. It is described in the key spec before this PR that the current design for notation key is to manage the `signing key list`, not the lifecycle of the signing key. So once you remove the key from the signing key list, it is still in existence. This is the truth for remote key. We didn't delete it from the key vault.

yizha1:
Close this PR as discussed. Keep `notation key delete` command, and remove alias `rm`

PR #468, opened 2022-12-06
update: check if verification is skipped by trust policy

patrickzheng200:
Previous to this PR: if -v flag is not set, and the verification process is skipped, notation would still print out message stating "Successfully verified signature for xxx".

This PR tries to fix this, if verification is actually skipped, notation would print out "Trust policy is configured to skip signature verification for xxx".

### DIFF
--- a/cmd/notation/verify.go
+++ b/cmd/notation/verify.go
@@ -104,6 +104,10 @@ func runVerify(command *cobra.Command, opts *verifyOpts) error {
 
 	// on success
 	outcome := outcomes[0]
+	if outcome.VerificationLevel.Name == "skip" {
+		fmt.Println("Trust policy is configured to skip signature verification for", ref.String())
+		return nil
+	}
 	// print out warning for any failed result with logged verification action
 	for _, result := range outcome.VerificationResults {
 		if result.Error != nil {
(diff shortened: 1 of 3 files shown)


JeyJeyGao (approved):
LGTM

shizhMSFT (approved):
LGTM

vaninrao10 (commented):
LGTM

cmd/notation/verify.go
```diff
@@ -112,7 +114,11 @@ func runVerify(command *cobra.Command, opts *verifyOpts) error {
 			fmt.Printf("Warning: %v was set to \"logged\" and failed with error: %v\n", result.Type, result.Error)
```
rgnote, line 114:
```suggestion
			fmt.Printf("Warning: %v was set to %q and failed with error: %v\n", result.Type, result.Action, result.Error)
```

rgnote, line 114: Out of scope for this PR, but we should be printing the outcome details of ALL signatures, not just the first one.

rgnote (approved):
LGTM

cmd/notation/verify.go
patrickzheng200: We could have this discussion post RC.1, but in short, we decided to only return verify outcome for the succeeded verified signature and early break. On failure, we will use the -v and -d flags to show the details of each signature.

shizhMSFT:
Linking https://github.com/notaryproject/notation-go/pull/231

PR #483, opened 2022-12-13
Fix: fixed notation cert command

### LINKED ISSUE #482: Notation CLI: can add an empty .docx file into the trust store and list it

Notation CLI

Notation CLI can add an empty .docx file into the trust store and list it.

An error should occur saying invalid certificate or cert not found in the file before adding it into the trust store.

### How can we reproduce it?

notation cert add ...
notation cert ls...

### Describe your environment

any

v1.0.0-rc.1

patrickzheng200:
Before adding a file into the trust store using `notation cert add`, notation CLI would check if it's valid first. Only file containing valid certificate(s) will be added. Same rule applies to the `notation cert list` command.

This PR fixes #482.

### DIFF
--- a/cmd/notation/internal/truststore/truststore.go
+++ b/cmd/notation/internal/truststore/truststore.go
@@ -37,9 +37,14 @@ func AddCert(path, storeType, namedStore string, display bool) error {
 		return errors.New("named store name needs to follow [a-zA-Z0-9_.-]+ format")
 	}
 
-	// check if the target path is a cert (support PEM and DER formats)
-	if _, err := corex509.ReadCertificateFile(certPath); err != nil {
+	// check if the target path is a x509 certificate
+	// (support PEM and DER formats)
+	certs, err := corex509.ReadCertificateFile(certPath)
+	if err != nil {
 		return err
+	}
+	if len(certs) == 0 {
+		return errors.New("no valid certificate found in the file")
 	}
 
 	// core process
@@ -48,6 +53,7 @@ func AddCert(path, storeType, namedStore string, display bool) error {
 	if err := CheckNonErrNotExistError(err); err != nil {
 		return err
 	}
+
 	// check if certificate already in the trust store
 	if _, err := os.Stat(filepath.Join(trustStorePath, filepath.Base(certPath))); err == nil {
 		return errors.New("certificate already exists in the Trust Store")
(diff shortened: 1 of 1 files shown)


JeyJeyGao (approved):
LGTM

cmd/notation/internal/truststore/truststore.go
```diff
@@ -37,17 +37,23 @@ func AddCert(path, storeType, namedStore string, display bool) error {
...
-	if _, err := corex509.ReadCertificateFile(certPath); err != nil {
+	// check if the target path is a x509 certificate
+	// (support PEM and DER formats)
+	certs, err := corex509.ReadCertificateFile(certPath)
+	if err != nil {
 		return err
 	}
+	if len(certs) == 0 {
+		return errors.New("no valid certificate found in the file")
+	}
```
priteshbandi, line 48: 1. can you please add unit test for this? Also should we add this check to `ReadCertificateFile` method ?
3. s/`no valid certificate found in the "empty file"

patrickzheng200: > 1. can you please add unit test for this? Also should we add this check to `ReadCertificateFile` method ?
> 2. s/`no valid certificate found in the "empty file"

1. hmm, so both notation CLI and notation-go are using this function. But they have different behaviors and returns different errors when encounter this situation. For the `notation certificate list` command, actually no error is returned, the file is just skipped. So probably we could just add this quick fix in CLI?
2. Code updated.

cmd/notation/internal/truststore/truststore.go
```diff
@@ -37,17 +37,23 @@ func AddCert(path, storeType, namedStore string, display bool) error {
...
-	// check if the target path is a cert (support PEM and DER formats)
-	if _, err := corex509.ReadCertificateFile(certPath); err != nil {
+	// check if the target path is a x509 certificate
+	// (support PEM and DER formats)
+	certs, err := corex509.ReadCertificateFile(certPath)
+	if err != nil {
 		return err
 	}
+	if len(certs) == 0 {
+		return errors.New("no valid certificate found in the empty file")
```
shizhMSFT, line 47: It might not be an empty file. It can be a non-empty file with no certs.

patrickzheng200: Oh yeah, that's true.

shizhMSFT (approved):
LGTM

cmd/notation/internal/truststore/truststore.go
```diff
@@ -83,10 +89,13 @@ func ListCerts(root string, depth int) error {
...
 		if info.Mode().IsRegular() {
-			if _, err := corex509.ReadCertificateFile(path); err != nil {
+			certs, err := corex509.ReadCertificateFile(path)
+			if err != nil {
 				return err
 			}
-			fmt.Println(path)
+			if len(certs) != 0 {
+				fmt.Println(path)
+			}
```
priteshbandi, line 98: Why do we need to display path?

patrickzheng200: Oh, this is the result display of the `notation cert list` command.

PR #497, opened 2022-12-27
chore: improve warning message when signing or verifying with tag

priteshbandi:
### Old
➜  notation git:(main) ✗ ./notation sign $IMAGE
Warning: Always sign the artifact using digest(`@sha256:...`) rather than a tag(`:v1`) because tags are mutable and a tag reference can point to a different artifact than the one signed.
Resolved artifact tag `v1` to digest `sha256:36ca4d6834ed680362327811238b97c687e77c5cf4a04a74d0853d3c0c17e059` before signing.
Successfully signed localhost:6000/net-monitor@sha256:36ca4d6834ed680362327811238b97c687e77c5cf4a04a74d0853d3c0c17e059

➜  notation git:(main) ✗ ./notation verify $IMAGE
Resolved artifact tag `v1` to digest `sha256:36ca4d6834ed680362327811238b97c687e77c5cf4a04a74d0853d3c0c17e059` before verification.
Successfully verified signature for localhost:6000/net-monitor@sha256:36ca4d6834ed680362327811238b97c687e77c5cf4a04a74d0853d3c0c17e059

### New

➜  notation git:(main) ✗ ./notation sign $IMAGE
Warning: Resolved artifact tag `v1` to digest `sha256:36ca4d6834ed680362327811238b97c687e77c5cf4a04a74d0853d3c0c17e059` before signing. Always sign the artifact using digest(`@sha256:...`) rather than a tag(`:v1`) because tags are mutable and a tag reference can point to a different artifact than the one signed.
Successfully signed localhost:6000/net-monitor@sha256:36ca4d6834ed680362327811238b97c687e77c5cf4a04a74d0853d3c0c17e059

➜  notation git:(main) ✗ ./notation verify $IMAGE
Warning: Resolved artifact tag `v1` to digest `sha256:36ca4d6834ed680362327811238b97c687e77c5cf4a04a74d0853d3c0c17e059` before verification. The resolved digest may not point to the same signed artifact, since tags are mutable.
Successfully verified signature for localhost:6000/net-monitor@sha256:36ca4d6834ed680362327811238b97c687e77c5cf4a04a74d0853d3c0c17e059

### DIFF
--- a/cmd/notation/sign.go
+++ b/cmd/notation/sign.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/notaryproject/notation-go"
@@ -100,8 +101,7 @@ func runSign(command *cobra.Command, cmdOpts *signOpts) error {
 
 func prepareSigningContent(ctx context.Context, opts *signOpts, sigRepo notationregistry.Repository) (notation.SignOptions, registry.Reference, error) {
 	ref, err := resolveReference(ctx, &opts.SecureFlagOpts, opts.reference, sigRepo, func(ref registry.Reference, manifestDesc ocispec.Descriptor) {
-		fmt.Printf("Warning: Always sign the artifact using digest(`@sha256:...`) rather than a tag(`:%s`) because tags are mutable and a tag reference can point to a different artifact than the one signed.\n", ref.Reference)
-		fmt.Printf("Resolved artifact tag `%s` to digest `%s` before signing.\n", ref.Reference, manifestDesc.Digest.String())
+		fmt.Fprintf(os.Stderr, "Warning: Always sign the artifact using digest(@sha256:...) rather than a tag(:%s) because tags are mutable and a tag reference can point to a different artifact than the one signed.\n", ref.Reference)
 	})
 	if err != nil {
 		return notation.SignOptions{}, registry.Reference{}, err
--- a/cmd/notation/verify.go
+++ b/cmd/notation/verify.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"os"
 	"reflect"
 
 	"github.com/notaryproject/notation-go"
@@ -72,8 +73,7 @@ func runVerify(command *cobra.Command, opts *verifyOpts) error {
 
 	// resolve the given reference and set the digest
 	ref, err := resolveReference(command.Context(), &opts.SecureFlagOpts, reference, sigRepo, func(ref registry.Reference, manifestDesc ocispec.Descriptor) {
-		fmt.Printf("Resolved artifact tag `%s` to digest `%s` before verification.\n", ref.Reference, manifestDesc.Digest.String())
-		fmt.Println("Warning: The resolved digest may not point to the same signed artifact, since tags are mutable.")
+		fmt.Fprintf(os.Stderr, "Warning: Always verify the artifact using digest(@sha256:...) rather than a tag(:%s) because resolved digest may not point to the same signed artifact, as tags are mutable.\n", ref.Reference)
 	})
 	if err != nil {
 		return err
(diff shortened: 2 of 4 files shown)


patrickzheng200 (commented):
Looks like there's a DCO failure. And should we remove cmd/notation/notation from this PR since it's a binary file?

yizha1:
@priteshbandi  any reason behind this change?

priteshbandi:
> @priteshbandi any reason behind this change?

Hi Yi, thanks for reviewing. Here are the reasons:
1. Consistent  warning message between sign and verify command (see old and new messages in PR description).
2. Also consolidating warning message to one line instead of two, for better readability.

patrickzheng200:
IIRC, the `warning` message for sign and verification are different by design. Because for verification, it's quite nature to verify a tagged manifest, and the corresponding digest will be returned. User could then pull by digest. Therefore, `Resolved artifact tag xxx to digest xxx before verification.` was not a part of the warning in the original spec design. @yizha1 please correct me if I'm wrong.

priteshbandi:
> IIRC, the `warning` message for sign and verification are different by design. Because for verification, it's quite nature to verify a tagged manifest, and the corresponding digest will be returned. User could then pull by digest. Therefore, `Resolved artifact tag xxx to digest xxx before verification.` was not a part of the warning in the original verify spec design. @yizha1 please correct me if I'm wrong.

In both sign and verify we are displaying what image digest was signed so that user can use digest for subsequent operations.

yizha1:
It is a bit different between `sign $tag` and `verify $tag`.
For `sign $tag`, we want users to make sure the artifact that they are signing is the right one and using `$digest` is the only correct way to ensure it. However, we don't fail `sign $tag` if users use `$tag` not `$digest`. So, there is a warning message first to warn the users before notation resolves the `$tag` to `$digest`. If we put the warning after resolving $tag, it looks like it is OK to use `$tag`, actually it is not.

yizha1:
It is a bit different between `sign $tag` and `verify $tag`.
For `sign $tag`, we want users to make sure the artifact that they are signing is the right one and using `$digest` is the only correct way to ensure it. However, we don't fail `sign $tag`, if users use `$tag` not `$digest`. So, there is a warning message first to warn the users before notation resolves the `$tag` to `$digest`. If we put the warning after resolving $tag, it looks like it is OK to use `$tag`, actually it is not.



> ### New
> ➜ notation git:(main) ✗ ./notation sign $IMAGE 
> Warning: Resolved artifact tag `v1` to digest `sha256:36ca4d6834ed680362327811238b97c687e77c5cf4a04a74d0853d3c0c17e059` before signing. Always sign the artifact using digest(`@sha256:...`) rather than a tag(`:v1`) because tags are mutable and a tag reference can point to a different artifact than the one signed. 
> Successfully signed localhost:6000/net-monitor@sha256:36ca4d6834ed680362327811238b97c687e77c5cf4a04a74d0853d3c0c17e059
>

priteshbandi:
> It is a bit different between `sign $tag` and `verify $tag`. For `sign $tag`, we want users to make sure the artifact that they are signing is the right one and using `$digest` is the only correct way to ensure it. However, we don't fail `sign $tag`, if users use `$tag` not `$digest`. So, there is a warning message first to warn the users before notation resolves the `$tag` to `$digest`. If we put the warning after resolving $tag, it looks like it is OK to use `$tag`, actually it is not.
> 
> > ### New
> > ➜ notation git:(main) ✗ ./notation sign $IMAGE
> > Warning: Resolved artifact tag `v1` to digest `sha256:36ca4d6834ed680362327811238b97c687e77c5cf4a04a74d0853d3c0c17e059` before signing. Always sign the artifact using digest(`@sha256:...`) rather than a tag(`:v1`) because tags are mutable and a tag reference can point to a different artifact than the one signed.
> > Successfully signed localhost:6000/net-monitor@sha256:36ca4d6834ed680362327811238b97c687e77c5cf4a04a74d0853d3c0c17e059

Either Sign or Verify the gravity of warning should be same because in both the scenarios the impact is roughly same. Should we just omit line `Resolved artifact tag v1 to digest sha256:36ca4d6834ed680362327811238b97c687e77c5cf4a04a74d0853d3c0c17e059` as this information is duplicate?

shizhMSFT (approved):
LGTM. It makes sense to write all warnings to `stderr` instead of `stdout`.

shizhMSFT:
Definitely, we need to make sure that the users understand that they are signing a particular content addressed by a digest.
Verification is a bit tricky. Given that we don't support tag signing right now, I think we can output a warning on resolving the tag for verification.

shizhMSFT:
One may argue the motivation to verify a content addressed by a known digest. It is about _authenticity_.

For example, Alice wants to pull an image with digest `sha256:abcd...`, which is an image of Bob's company. If the trust policy is set to only allow images from Alice's company, the pull should fail due to verification failure unless this image has a valid signature from Alice's company.

jondonas (approved):
LGTM

PR #498, opened 2023-01-03
doc: update notation sign and verify spec for metadata

byronchien:
Spec update to support https://github.com/notaryproject/roadmap/issues/67

`notation sign`: 
- user will be able to specify additional key value pairs with the `--user-metadata` flag (`-um` short) that will be signed as part of the payload.

`notation verify`:
- user will be able to specify additional key value pairs with the `--user-metadata` flag (`-um` short) that must be present in the signature to pass verification.
- user will be able to configure output format as either plaintext (default) or json with the `--output` flag (`-o` short)

### DIFF
--- a/specs/commandline/sign.md
+++ b/specs/commandline/sign.md
@@ -40,11 +40,6 @@ Flags:
   -um, --user-metadata strings    {key}={value} pairs that are added to the signature
 ```
 
-### User Metadata Restrictions
-- `notation sign` must fail if metadata keys or values are not strings
-- `notation sign` must fail if duplicate metadata keys are provided
-- `notation sign` must fail if metadata keys conflict with reserved Notary or OCI keys
-
 ## Usage
 
 ### Sign an OCI artifact
--- a/specs/commandline/sign.md
+++ b/specs/commandline/sign.md
@@ -29,15 +29,21 @@ Usage:
   notation sign [flags] <reference>
 
 Flags:
-  -e, --expiry duration          optional expiry that provides a "best by use" time for the artifact. The duration is specified in minutes(m) and/or hours(h). For example: 12h, 30m, 3h20m
-  -h, --help                     help for sign
-  -k, --key string               signing key name, for a key previously added to notation's key list.
-  -p, --password string          password for registry operations (default to $NOTATION_PASSWORD if not specified)
-      --plain-http               registry access via plain HTTP
-      --plugin-config strings    {key}={value} pairs that are passed as it is to a plugin, refer plugin's documentation to set appropriate values
-      --signature-format string  signature envelope format, options: 'jws', 'cose' (default "jws")
-  -u, --username string          username for registry operations (default to $NOTATION_USERNAME if not specified)
+  -e,  --expiry duration          optional expiry that provides a "best by use" time for the artifact. The duration is specified in minutes(m) and/or hours(h). For example: 12h, 30m, 3h20m
+  -h,  --help                     help for sign
+  -k,  --key string               signing key name, for a key previously added to notation's key list.
+  -p,  --password string          password for registry operations (default to $NOTATION_PASSWORD if not specified)
+       --plain-http               registry access via plain HTTP
+       --plugin-config strings    {key}={value} pairs that are passed as it is to a plugin, refer plugin's documentation to set appropriate values
+       --signature-format string  signature envelope format, options: 'jws', 'cose' (default "jws")
+  -u,  --username string          username for registry operations (default to $NOTATION_USERNAME if not specified)
+  -um, --user-metadata strings    {key}={value} pairs that are added to the signature
 ```
+
+### User Metadata Restrictions
+- `notation sign` must fail if metadata keys or values are not strings
+- `notation sign` must fail if duplicate metadata keys are provided
+- `notation sign` must fail if metadata keys conflict with reserved Notary or OCI keys
 
 ## Usage
 
@@ -80,6 +86,16 @@ notation sign --signature-format cose <registry>/<repository>@<digest>
 
 # Use a digest that uniquely and immutably identifies an OCI artifact.
 notation sign <registry>/<repository>@<digest>
+```
+
+### Sign an OCI Artifact with user metadata
+
+```shell
+# Prerequisites: 
+# A default signing key is configured using CLI "notation key"
+
+# sign an artifact stored in a registry and add user-metadata io.wabbit-networks.buildId=123 to the payload
+notation sign <registry>/<repository>@<digest> --user-metadata io.wabbit-networks.buildId=123
 ```
 
 ### Sign an OCI artifact stored in a registry and specify the signature expiry duration, for example 24 hours
--- a/specs/commandline/sign.md
+++ b/specs/commandline/sign.md
@@ -37,7 +37,7 @@ Flags:
        --plugin-config strings    {key}={value} pairs that are passed as it is to a plugin, refer plugin's documentation to set appropriate values
        --signature-format string  signature envelope format, options: 'jws', 'cose' (default "jws")
   -u,  --username string          username for registry operations (default to $NOTATION_USERNAME if not specified)
-  -um, --user-metadata strings    {key}={value} pairs that are added to the signature
+  -m,  --user-metadata strings    {key}={value} pairs that are added to the signature payload
 ```
 
 ## Usage
@@ -90,7 +90,10 @@ notation sign <registry>/<repository>@<digest>
 # A default signing key is configured using CLI "notation key"
 
 # sign an artifact stored in a registry and add user-metadata io.wabbit-networks.buildId=123 to the payload
-notation sign <registry>/<repository>@<digest> --user-metadata io.wabbit-networks.buildId=123
+notation sign --user-metadata io.wabbit-networks.buildId=123 <registry>/<repository>@<digest>
+
+# sign an artifact stored in a registry and add user-metadata io.wabbit-networks.buildId=123 and io.wabbit-networks.buildTime=1672944615 to the payload
+notation sign --user-metadata io.wabbit-networks.buildId=123 --user-metadata io.wabbit-networks.buildTime=1672944615 <registry>/<repository>@<digest>
 ```
 
 ### Sign an OCI artifact stored in a registry and specify the signature expiry duration, for example 24 hours
--- a/specs/commandline/verify.md
+++ b/specs/commandline/verify.md
@@ -21,8 +21,10 @@ The signed descriptor may have user defined metadata attached. If the signature 
 ```text
 Successfully verified signature for <registry>/<repository>@<digest>
 
-User Metadata:
-- <key> : <value>
+The artifact is signed with the following user metadata.
+
+KEY    VALUE
+<key>  <value>
 ```
 
 ## Outline
@@ -35,12 +37,11 @@ Usage:
 
 Flags:
   -h,  --help                    help for verify
-  -o,  --output string           output format, options: 'plaintext', 'json' (default: 'plaintext')
   -p,  --password string         password for registry operations (default to $NOTATION_PASSWORD if not specified)
        --plain-http              registry access via plain HTTP
        --plugin-config strings   {key}={value} pairs that are passed as it is to a plugin, if the verification is associated with a verification plugin, refer plugin documentation to set appropriate values
   -u,  --username string         username for registry operations (default to $NOTATION_USERNAME if not specified)
-  -um, --user-metadata strings   user defined {key}={value} pairs that must be present in the signature for successful verification if provided
+  -m,  --user-metadata strings   user defined {key}={value} pairs that must be present in the signature for successful verification if provided
 ```
 
 ## Usage
@@ -129,8 +130,8 @@ Successfully verified signature for localhost:5000/net-monitor@sha256:b94d27b993
 Use the `--user-metadata` flag to verify that provided key-value pairs are present in the payload of the valid signature.
 
 ```shell
-# Verify signatures on the supplied OCI artifact identified by the digest and verify that io.wabbit-networks.data=foo is present in the signed payload
-notation verify localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9 --user-metadata io.wabbit-networks.data=foo
+# Verify signatures on the supplied OCI artifact identified by the digest and verify that io.wabbit-networks.buildId=123 is present in the signed payload
+notation verify --user-metadata io.wabbit-networks.buildId=123 localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
 ```
 
 An example of output messages for a successful verification:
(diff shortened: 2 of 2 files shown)


specs/commandline/sign.md
```diff
@@ -29,16 +29,22 @@ Usage:
...
+       --plugin-config strings    {key}={value} pairs that are passed as it is to a plugin, refer plugin's documentation to set appropriate values
+       --signature-format string  signature envelope format, options: 'jws', 'cose' (default "jws")
+  -u,  --username string          username for registry operations (default to $NOTATION_USERNAME if not specified)
+  -um, --user-metadata strings    {key}={value} pairs that are added to the signature
 ```
 
+### User Metadata Restrictions
+- `notation sign` must fail if metadata keys or values are not strings
+- `notation sign` must fail if duplicate metadata keys are provided
+- `notation sign` must fail if metadata keys conflict with reserved Notary or OCI keys
```
priteshbandi, line 46: Since this is cli spec and the restrictions are coming from notary v2 spec, should we remove them from here?

specs/commandline/verify.md
```diff
@@ -25,11 +34,13 @@ Usage:
...
-      --plain-http              registry access via plain HTTP
-      --plugin-config strings   {key}={value} pairs that are passed as it is to a plugin, if the verification is associated with a verification plugin, refer plugin documentation to set appropriate values
-  -u, --username string         username for registry operations (default to $NOTATION_USERNAME if not specified)
+  -h,  --help                    help for verify
+  -o,  --output string           output format, options: 'plaintext', 'json' (default: 'plaintext')
+  -p,  --password string         password for registry operations (default to $NOTATION_PASSWORD if not specified)
+       --plain-http              registry access via plain HTTP
+       --plugin-config strings   {key}={value} pairs that are passed as it is to a plugin, if the verification is associated with a verification plugin, refer plugin documentation to set appropriate values
+  -u,  --username string         username for registry operations (default to $NOTATION_USERNAME if not specified)
+  -um, --user-metadata strings   {key}={value} pairs that must be present in the signature for successful verification if provided
```
priteshbandi, line 43:
```suggestion
  -um, --user-metadata strings   user defined {key}={value} pairs that must be present in the signature for successful verification if provided
```

specs/commandline/verify.md
```diff
@@ -131,3 +142,42 @@ Resolved artifact tag `v1` to digest `sha256:b94d27b9934d3e08a52e52d7da7dabfac48
 Warning: The resolved digest may not point to the same signed artifact, since tags are mutable.
 Successfully verified signature for localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
 ```
+
+### Verify signatures on an OCI artifact with user metadata
```
priteshbandi, line 146: I would move this to line 127 and update the example to use digest instead of tag in registry reference.

specs/commandline/verify.md
```diff
@@ -131,3 +142,42 @@ Resolved artifact tag `v1` to digest `sha256:b94d27b9934d3e08a52e52d7da7dabfac48
...
+```text
+Resolved artifact tag `v1` to digest `sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9` before verification.
+Warning: The resolved digest may not point to the same signed artifact, since tags are mutable.
+Successfully verified signature for localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
+
+User Metadata:
+- io.wabbit-networks.data : foo
+```
+
+### Verify signatures on an OCI artifact and format output as json
```
priteshbandi, line 167: ditto

specs/commandline/verify.md
```diff
@@ -131,3 +142,42 @@ Resolved artifact tag `v1` to digest `sha256:b94d27b9934d3e08a52e52d7da7dabfac48
...
+# Prerequisites: Signatures are stored in a registry referencing the signed OCI artifact
+# Verify signatures on an OCI artifact identified by the tag
+notation verify localhost:5000/net-monitor:v1 --output json
+```
+
+An example of output messages for a successful verification:
+
+```text
+Resolved artifact tag `v1` to digest `sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9` before verification.
+Warning: The resolved digest may not point to the same signed artifact, since tags are mutable.
```
priteshbandi, line 181: Should we omit warning when output is in json?

specs/commandline/verify.md
```diff
@@ -113,13 +124,46 @@ An example of output messages for a successful verification:
...
+An example of output messages for a successful verification:
+
+```text
+Successfully verified signature for localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
+
+User Metadata:
+- io.wabbit-networks.data : foo
+```
+
+### Verify signatures on an OCI artifact and format output as json
```
yizha1, line 145: I suggest creating a new PR for `--output` flag for both `notation sign` and `notation verify` commands. Besides successful cases, the follows examples are required as well
- example output of failure cases
- example output of successful or failure cases when additional `--debug` (debug logs) or `--verbose` flag (info logs) enabled

specs/commandline/sign.md
```diff
@@ -29,14 +29,15 @@ Usage:
...
-  -u, --username string          username for registry operations (default to $NOTATION_USERNAME if not specified)
+  -e,  --expiry duration          optional expiry that provides a "best by use" time for the artifact. The duration is specified in minutes(m) and/or hours(h). For example: 12h, 30m, 3h20m
+  -h,  --help                     help for sign
+  -k,  --key string               signing key name, for a key previously added to notation's key list.
+  -p,  --password string          password for registry operations (default to $NOTATION_PASSWORD if not specified)
+       --plain-http               registry access via plain HTTP
+       --plugin-config strings    {key}={value} pairs that are passed as it is to a plugin, refer plugin's documentation to set appropriate values
+       --signature-format string  signature envelope format, options: 'jws', 'cose' (default "jws")
+  -u,  --username string          username for registry operations (default to $NOTATION_USERNAME if not specified)
+  -um, --user-metadata strings    {key}={value} pairs that are added to the signature
```
yizha1, line 40: Is this metadata added to the annotation attributes of payload https://github.com/notaryproject/notaryproject/blob/main/specs/signature-specification.md#payload?

yizha1, line 40: I think the short version should contain only one character, maybe using `-m` as metadata

specs/commandline/sign.md
```diff
@@ -82,6 +83,16 @@ notation sign --signature-format cose <registry>/<repository>@<digest>
...
 ```
 
+### Sign an OCI Artifact with user metadata
+
+```shell
+# Prerequisites: 
+
+# sign an artifact stored in a registry and add user-metadata io.wabbit-networks.buildId=123 to the payload
+notation sign <registry>/<repository>@<digest> --user-metadata io.wabbit-networks.buildId=123
```
yizha1, line 93: Can we add an example for adding multiple user metadata? I suppose it is something like 
`--user-metadata io.wabbit-networks.buildId=123 --user-metadata io.wabbit-networks.buildTime="xxx"`

specs/commandline/verify.md
```diff
@@ -25,11 +34,13 @@ Usage:
...
-      --plain-http              registry access via plain HTTP
-      --plugin-config strings   {key}={value} pairs that are passed as it is to a plugin, if the verification is associated with a verification plugin, refer plugin documentation to set appropriate values
-  -u, --username string         username for registry operations (default to $NOTATION_USERNAME if not specified)
+  -h,  --help                    help for verify
+  -o,  --output string           output format, options: 'plaintext', 'json' (default: 'plaintext')
+  -p,  --password string         password for registry operations (default to $NOTATION_PASSWORD if not specified)
+       --plain-http              registry access via plain HTTP
+       --plugin-config strings   {key}={value} pairs that are passed as it is to a plugin, if the verification is associated with a verification plugin, refer plugin documentation to set appropriate values
+  -u,  --username string         username for registry operations (default to $NOTATION_USERNAME if not specified)
+  -um, --user-metadata strings   user defined {key}={value} pairs that must be present in the signature for successful verification if provided
```
yizha1, line 43: same as above, `-um` contains two chars.

specs/commandline/verify.md
```diff
@@ -113,13 +124,46 @@ An example of output messages for a successful verification:
...
+notation verify localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9 --user-metadata io.wabbit-networks.data=foo
+```
+
+An example of output messages for a successful verification:
+
+```text
+Successfully verified signature for localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
+
+User Metadata:
+- io.wabbit-networks.data : foo
```
yizha1, line 142: Can we use the same metadata used in sign.md, which is `io.wabbit-networks.buildId: 123` in this case.

specs/commandline/verify.md
```diff
@@ -113,13 +124,46 @@ An example of output messages for a successful verification:
 Successfully verified signature for localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
 ```
 
+### Verify signatures on an OCI artifact with user metadata
+
+Use the `--user-metadata` flag to verify that provided key-value pairs are present in the payload of the valid signature.
+
+```shell
+# Verify signatures on the supplied OCI artifact identified by the digest and verify that io.wabbit-networks.data=foo is present in the signed payload
+notation verify localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9 --user-metadata io.wabbit-networks.data=foo
```
yizha1, line 133: Can we use the same metadata used in sign.md, which is `io.wabbit-networks.buildId=123` in this case.

yizha1, line 133: What will happen if `--user-metadata` was not specified in `notation verify` command, however the metadata was added during `notation sign`

yizha1 (commented):
Thanks @byronchien for the contribution. I have provided my comments.

specs/commandline/sign.md
byronchien: yes, metadata will be added to the annotation attributes of the payload. is this something that needs to be called out in the spec?

specs/commandline/verify.md
byronchien: if `--user-metadata` is not specified in `notation verify`, then metadata won't be verified as part of overall signature verification even if the metadata is present.

specs/commandline/sign.md
yizha1: I suggest adding a reference to the signature spec, so that people can understand these metadata will be added as annotations for the payload as described in the signature spec payload section. You may be aware, there are also annotation attributes for signature manifest. Check this out Check out this https://github.com/notaryproject/notaryproject/blob/main/specs/signature-specification.md#storage. It may cause confusion.

Maybe we can also make the description of this new flag more clear
`{key}={value} pairs that are added to the signature` -> `{key}={value} pairs that are added to the signature payload`

specs/commandline/verify.md
yizha1: I think we need to document this behavior in the signature spec.  Any comments @priteshbandi @shizhMSFT

priteshbandi: No, user metadata is signature envelope agnostic, it will always be string-string key value pair.

specs/commandline/sign.md
priteshbandi: I would avoid any references in command help because the url might change but yes we should add some information about how they are being stored in the spec.

specs/commandline/verify.md
yizha1: @priteshbandi  I mean should we document this behavior "if --user-metadata is not specified in notation verify, then metadata won't be verified as part of overall signature verification even if the metadata is present."

specs/commandline/sign.md
shizhMSFT: Cobra does not support multi-character short alias.

shizhMSFT: If it is `-m`, why not `--metadata` instead of `--user-metadata`?

specs/commandline/sign.md
shizhMSFT, line 93: It is better if we can put flags before arguments in examples.

specs/commandline/verify.md
```diff
@@ -16,6 +16,15 @@ Warning: The resolved digest may not point to the same signed artifact, since ta
 Successfully verified signature for <registry>/<repository>@<digest>
 ```
 
+The signed descriptor may have user defined metadata attached. If the signature for the OCI artifact contains any metadata, the output message is as follows:
+
+```text
+Successfully verified signature for <registry>/<repository>@<digest>
+
+User Metadata:
+- <key> : <value>
```
shizhMSFT, line 25: What's the output if there is no user metadata?

shizhMSFT, line 25: nit: Should it be
```suggestion
User Metadata:
- <key>: <value>
```

shizhMSFT: or will it be better if it is a table view like
```
The artifact is signed with the following user metadata.

KEY    VALUE
<key>  <value>
```

specs/commandline/verify.md
shizhMSFT: The `--output` flag to specify the display format may be fine for `notation verify`. However, it may not be good for `notation sign`, especially for offline signing scenarios.

For example,
```
notation sign --oci net-monitor.tar@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9 --output net-monitor.sig
```

specs/commandline/verify.md
byronchien: i'm down to change it, this would make it more consistent with other commands like plugin / key list

specs/commandline/sign.md
byronchien: the rationale behind `--user-metadata` over `--metadata` is that the former is more clear that the metadata are user-specified values and just `--metadata` would be too generic.

specs/commandline/verify.md
byronchien: can move this to the output PR when i create that one, but outputting the signature information as json is needed to allow consumption of the metadata programmatically, so for now, just proposing that we add it for `verify` only. is the concern about potential overlap with some other flag for sign?

specs/commandline/verify.md
byronchien: if no user metadata, then this section is omitted (same as existing output)

specs/commandline/verify.md
byronchien: If we're already saying to use `--user-metadata` to verify key value pairs, is it necessary to also specify the opposite? (if you don't specify `--user-metadata` then we won't verify)

specs/commandline/verify.md
```diff
@@ -113,13 +125,59 @@ An example of output messages for a successful verification:
...
+The artifact is signed with the following user metadata.
+
+KEY                         VALUE
+io.wabbit-networks.buildId  123
+```
+
+An example of output messages for an unsuccessful verification:
+
+```text
+Error: signature verification failed for all the signatures associated with localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
```
yizha1, line 151: If the verification failure is due to metadata, there should be error log like you shared under the debug logging.

"Error: specified metadata is not present in the signature."

yizha1 (approved):
LGTM with some comments

shizhMSFT (approved):
LGTM

specs/commandline/verify.md
byronchien: json output removed from this PR, so resolving

specs/commandline/verify.md
byronchien: output reformatting moved to a separate PR, so resolving this

specs/commandline/sign.md
byronchien: not adding a reference to the annotation / signature specs then

specs/commandline/verify.md
```diff
@@ -113,13 +125,62 @@ An example of output messages for a successful verification:
...
+
+```text
+...
+INFO User Metadata flag is present. Checking signature metadata for specified values.
+DEBU[2023-01-05T11:35:07-08:00] Verifying that metadata { "io.wabbit-networks.buildId":"123" } is present in signature metadata.
+DEBU[2023-01-05T11:35:07-08:00] Signature metadata: { "io.wabbit-networks.buildId":"321" }
+DEBU[2023-01-05T11:35:07-08:00] Error: specified metadata is not present in the signature.
+Error: unable to find specified metadata in any signatures associated with localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
+Error: signature verification failed for all the signatures associated with localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
+```
```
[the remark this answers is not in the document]
byronchien: ok, will remove, also matches existing spec that doesn't have verbose/debug examples

priteshbandi (commented):
LGTM

priteshbandi (approved):

PR #502, opened 2023-01-07
doc: update to support OCI image manifest

yizha1:
Updates:

- Add a new section of "Use OCI image manifest to store signatures"
- Add new usage examples

### DIFF
--- a/specs/commandline/sign.md
+++ b/specs/commandline/sign.md
@@ -31,7 +31,7 @@ Usage:
 Flags:
   -e,  --expiry duration          optional expiry that provides a "best by use" time for the artifact. The duration is specified in minutes(m) and/or hours(h). For example: 12h, 30m, 3h20m
   -h,  --help                     help for sign
-       --image-spec string        specify the manifest type for signatures. options: v1.1-artifact, v1.1-image (default: v1.1-artifact)
+       --image-spec string        manifest type for signatures. options: v1.1-artifact, v1.1-image (default: v1.1-artifact)
   -k,  --key string               signing key name, for a key previously added to notation's key list.
   -p,  --password string          password for registry operations (default to $NOTATION_PASSWORD if not specified)
        --plain-http               registry access via plain HTTP
@@ -43,13 +43,13 @@ Flags:
 
 ## Use OCI image manifest to store signatures
 
-By default, Notation uses [OCI artifact manifest][oci-artifact-manifest] to store signatures in the registries. For backward compatibility, Notation supports using `OCI image manifest` to store the signature in the registries that implement partial of the [OCI Image specification v1.1][oci-image-spec]. Use flag `--image-spec v1.1-image` to force Notation store the signatures using OCI image manifest.
+By default, Notation uses [OCI artifact manifest][oci-artifact-manifest] to store signatures in registries. For backward compatibility, Notation supports using `OCI image manifest` to store signatures in registries that partially implement the [OCI Image specification v1.1][oci-image-spec]. Use flag `--image-spec v1.1-image` to force Notation to store the signatures using OCI image manifest.
 
-Registries MAY not implement or enable the `Referrers API`, which is used by the client to fetch the referrers. In the context of Notation, the referrers are signatures. Notation follows the fallback procedure defined in [OCI distribution spec][oci-backward-compatibility] if `Referrers API` is unavailable.
+Registries MAY not implement or enable the `Referrers API`, which is used by clients to fetch referrers. In the context of Notation, the referrers are signatures. Notation follows the fallback procedure defined in [OCI distribution spec][oci-backward-compatibility] if `Referrers API` is unavailable.
 
 ### Set config property for OCI image manifest
 
-OCI image manifest requires additional property `config` of type `descriptor`, which is not required by OCI artifact manifest. Notation creates a default config descriptor for the user if the flag `--image-spec v1.1-image` is used.
+OCI image manifest requires additional property `config` of type `descriptor`, which is not required by OCI artifact manifest. Notation creates a default config descriptor for the user if flag `--image-spec v1.1-image` is used.
 
 Notation uses empty JSON object `{}` as the default configuration content, and thus the default `config` property is fixed, as following:
 
@@ -63,9 +63,9 @@ Notation uses empty JSON object `{}` as the default configuration content, and t
 
 ### When to use OCI image manifest
 
-[Registry support][registry-support] lists registries with different compatibilities. For registries not supporting `OCI artifact manifest`, user can use flag `--image-spec v1.1-image` to sign artifacts stored in those registries.
+[Registry support][registry-support] lists registries with different compatibilities. For registries not supporting `OCI artifact manifest`, users can use flag `--image-spec v1.1-image` to sign artifacts stored in those registries.
 
-In case registries not listed in the page, user can consider using flag `--image-spec v1.1-image` by checking the error message. Note that there is no deterministic way to determine a registry supporting `OCI artifact manifest` or not. The error message is just for reference. The following response status contained in error messages MAY indicate that the registries doesn't support `OCI artifact manifest`:
+For registries not listed in the page, users can consider using flag `--image-spec v1.1-image` by checking the error message. Note that there is no deterministic way to determine whether a registry supports `OCI artifact manifest` or not. The error message is just for reference. The following response status contained in error messages MAY indicate that the registry doesn't support `OCI artifact manifest`:
 
 - Response status `400 BAD Request` with error code `MANIFEST_INVALID` or `UNSUPPORTED`
 


specs/commandline/sign.md
```diff
@@ -37,8 +38,52 @@ Flags:
...
+
+Registries MAY not implement or enable the `Referrers API`, which is used by the client to fetch the referrers. In the context of Notation, the referrers are signatures. Notation follows the fallback procedure defined in [OCI distribution spec][oci-backward-compatibility] if `Referrers API` is unavailable.
+
+### Set config property for OCI image manifest
+
+OCI image manifest requires additional property `config` of type `descriptor`, which is not required by OCI artifact manifest. Notation creates a default config descriptor for the user if the flag `--use-image-manifest` is used. User can specify custom configurations by using flag `--config-file` for specific needs, which overrides the default configurations.
```
shizhMSFT, line 52: > User can specify custom configurations by using flag `--config-file` for specific needs, which overrides the default configurations.

Same as above. What's the purpose of customizing the config blob?

specs/commandline/sign.md
```diff
@@ -29,6 +29,7 @@ Usage:
   notation sign [flags] <reference>
 
 Flags:
+      --config-file path         optional path of image config file. Should only be used with flag --use-image-manifest
```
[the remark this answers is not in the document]
yizha1: Remove this parameter since there is no user requirement on it.

specs/commandline/sign.md
yizha1: Remove it as no user requirement.

specs/commandline/sign.md
```diff
@@ -37,8 +37,44 @@ Flags:
...
+
+Registries MAY not implement or enable the `Referrers API`, which is used by the client to fetch the referrers. In the context of Notation, the referrers are signatures. Notation follows the fallback procedure defined in [OCI distribution spec][oci-backward-compatibility] if `Referrers API` is unavailable.
+
+### Set config property for OCI image manifest
+
+OCI image manifest requires additional property `config` of type `descriptor`, which is not required by OCI artifact manifest. Notation creates a default config descriptor for the user if the flag `--use-image-manifest` is used.
+
+Notation uses empty JSON object `{}` as the default configuration content, and thus the default `config` property is fixed, as following:
```
FeynmanZhou, line 53: @nima Does ECR accept this config?

specs/commandline/sign.md
```diff
@@ -37,8 +37,44 @@ Flags:
       --plugin-config strings    {key}={value} pairs that are passed as it is to a plugin, refer plugin's documentation to set appropriate values
       --signature-format string  signature envelope format, options: 'jws', 'cose' (default "jws")
   -u, --username string          username for registry operations (default to $NOTATION_USERNAME if not specified)
+      --use-image-manifest       optional for using OCI image manifest to store signatures in a registry
 ```
 
+## Use OCI image manifest to store signatures
+
+By default, Notation uses [OCI artifact manifest][oci-image-manifest] to store signatures in the registries. For backward compatibility, Notation supports using `OCI image manifest` to store the signature in the registries that implement partial or older versions of the OCI Image specification. Use flag `--use-image-manifest` to force Notation store the signatures using OCI image manifest. Note that Notation doesn't automatically switch to `OCI image manifest` for a failure of putting `OCI artifact manifest`.
```
patrickzheng200, line 45: nit:
```suggestion
By default, Notation uses [OCI artifact manifest][oci-image-manifest] to store signatures in the registries. For backward compatibility, Notation supports using `OCI image manifest` to store the signature in the registries that implement partial or older versions of the OCI Image specification. Use flag `--use-image-manifest` to force Notation store the signatures using OCI image manifest. Note that Notation doesn't automatically switch to `OCI image manifest` for a failure of using `OCI artifact manifest`.
```

specs/commandline/sign.md
```diff
@@ -116,3 +152,15 @@ Warning: Always sign the artifact using digest(`@sha256:...`) rather than a tag(
...
+
+```shell
+notation sign --use-image-manifest <registry>/<repository>@<digest>
+```
+
+## References
+
+[oci-image-manifest]: https://github.com/opencontainers/image-spec/blob/v1.0.2/manifest.md
+[oci-backward-compatibility]: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#backwards-compatibility
+[registry-support]: https://notaryproject.dev/docs/registrysupport/
```
patrickzheng200, line 166: These lines are not rendered in the markdown page. Hence:
```suggestion
oci-image-manifest: https://github.com/opencontainers/image-spec/blob/v1.0.2/manifest.md</br>
oci-backward-compatibility: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#backwards-compatibility</br>
registry-support: https://notaryproject.dev/docs/registrysupport/</br>
```

yizha1: I should keep the format, but remove the title `##Reference`, this links won't be showed, but replace the reference in the markdown file.

patrickzheng200 (approved):
LGTM

specs/commandline/sign.md
nima: Yes, I have a sample output which verifies this; keeping in mind for the following that `sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a` is the digest of `{}`:

```md
### (W) `PUT blobs/uploads/{orasConfig.blob}` :: `201 Created` ✅
#### Request
DEBU[0002]  Request URL: "https://080565210187.dkr.ecr.us-east-1.amazonaws.com/v2/utnubu/blobs/uploads/4fe80440-6088-4443-b805-c36a5c2ee156?digest=sha256%3A44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a"
DEBU[0002]  Request method: "PUT"
DEBU[0002]  Request headers:
DEBU[0002]    "Content-Type": "application/octet-stream"
DEBU[0002]    "Authorization": "*****"
DEBU[0002]    "User-Agent": "oras/0.16.0+Homebrew"
#### Response
DEBU[0002]  Response Status: "201 Created"
DEBU[0002]  Response headers:
DEBU[0002]    "Sizes": "1.000000"
DEBU[0002]    "Date": "Sun, 18 Dec 2022 01:32:30 GMT"
DEBU[0002]    "Content-Length": "0"
DEBU[0002]    "Docker-Content-Digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a"
DEBU[0002]    "Docker-Distribution-Api-Version": "registry/2.0"
DEBU[0002]    "Location": "https://080565210187.dkr.ecr.us-east-1.amazonaws.com/v2/utnubu/blobs/sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a"
```

specs/commandline/sign.md
```diff
@@ -37,8 +37,44 @@ Flags:
       --plugin-config strings    {key}={value} pairs that are passed as it is to a plugin, refer plugin's documentation to set appropriate values
       --signature-format string  signature envelope format, options: 'jws', 'cose' (default "jws")
   -u, --username string          username for registry operations (default to $NOTATION_USERNAME if not specified)
+      --use-image-manifest       optional for using OCI image manifest to store signatures in a registry
```
priteshbandi, line 40: should we use some user friendly name here instead of internal implementation? like ```--use-oci-1.1``` to reflect that if registry support oci1.1 add this flag ?

I know its not a good name but I'm open to suggestions

specs/commandline/sign.md
```diff
@@ -37,8 +37,44 @@ Flags:
...
 ```
 
+## Use OCI image manifest to store signatures
+
+By default, Notation uses [OCI artifact manifest][oci-image-manifest] to store signatures in the registries. For backward compatibility, Notation supports using `OCI image manifest` to store the signature in the registries that implement partial or older versions of the OCI Image specification. Use flag `--use-image-manifest` to force Notation store the signatures using OCI image manifest. Note that Notation doesn't automatically switch to `OCI image manifest` for a failure of using `OCI artifact manifest`.
```
priteshbandi, line 45: > that implement partial or older versions of the OCI Image specification

Older than what version? Can we please add more information here?

I would remove ```Note that Notation doesn't automatically switch to `OCI image manifest` for a failure of using `OCI artifact manifest``` because its implicit from explicitly using a flag. Also, when we do automatic fallback, we will need to delete it and which we might forget.

specs/commandline/sign.md
```diff
@@ -37,8 +37,44 @@ Flags:
...
+
+OCI image manifest requires additional property `config` of type `descriptor`, which is not required by OCI artifact manifest. Notation creates a default config descriptor for the user if the flag `--use-image-manifest` is used.
+
+Notation uses empty JSON object `{}` as the default configuration content, and thus the default `config` property is fixed, as following:
+
+```json
+"config": {
+    "mediaType": "application/vnd.cncf.notary.signature",
+    "size": 2,
+    "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a"
```
priteshbandi, line 59: why size of 2? Also, can we use 0 to show the intent that its a dummy value?

Similarly for digest, can we use `00000000000...` to show the intent that its a dummy value?

specs/commandline/sign.md
```diff
@@ -37,8 +37,44 @@ Flags:
...
+
+### Set config property for OCI image manifest
+
+OCI image manifest requires additional property `config` of type `descriptor`, which is not required by OCI artifact manifest. Notation creates a default config descriptor for the user if the flag `--use-image-manifest` is used.
+
+Notation uses empty JSON object `{}` as the default configuration content, and thus the default `config` property is fixed, as following:
+
+```json
+"config": {
+    "mediaType": "application/vnd.cncf.notary.signature",
```
priteshbandi, line 57: config should never be of type `application/vnd.cncf.notary.signature` should we use some dummy value like `application/vnd.cncf.notary.dummy.config`?

specs/commandline/sign.md
```diff
@@ -37,8 +37,44 @@ Flags:
...
+
+[Registry support][registry-support] lists registries with different compatibilities. For registries not supporting `OCI artifact manifest`, user can use flag `--user-image-manifest` to sign artifacts stored in those registries.
+
+In case registries not listed in the page, user can consider using flag `--user-image-manifest` by checking the error message. Note that there is no deterministic way to determine a registry supporting `OCI artifact manifest` or not. The error message is just for reference. The following response status contained in error messages MAY indicate that the registries doesn't support `OCI artifact manifest`:
+
+- Response status `400 BAD Request` with error code `MANIFEST_INVALID` or `UNSUPPORTED`
+
+### Copy signatures between registries
+
+User may use tools to copy artifacts and associated signatures between two registries. The copy succeeds under the following conditions:
```
priteshbandi, line 73: Can you please add some examples of tools?

specs/commandline/sign.md
```diff
@@ -37,8 +37,44 @@ Flags:
...
+
+- From registries that support `OCI artifact manifest` to registries that support `OCI artifact manifest`
+- From registries that support `OCI image manifest` to registries that support `OCI image manifest`
```
priteshbandi, line 76: Will a user be able to move signatures from a registry that supports `OCI image manifest` to a registry that supports `OCI artifact manifest` and vice-versa?

specs/commandline/sign.md
yizha1: @shizhMSFT @FeynmanZhou  any suggestions on the flag name?

specs/commandline/sign.md
yizha1: Some registries don't support empty, 0 size config.

The descriptor requirements describe what the value should be set for `size` and `digest`.

Since the config content is an empty json `{}`, so the size of raw content is `2` per the OCI image manifest spec. 

The digest is calculated based on the config content `{}` using the following command:

```shell
echo -n "{}" | sha256sum
```

specs/commandline/sign.md
yizha1: I can add an example of `ORAS`

specs/commandline/sign.md
yizha1: According to the spec, we should do it 

https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pushing-manifests-with-subject

and referrer API also requires https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers

specs/commandline/sign.md
yizha1: how about `--image-spec v1.1-image`

specs/commandline/sign.md
yizha1: No, currently 
- Registry with `OCI artifact manifest` support doesn't mean `OCI image manifest` with `subject` field is supported
- Registry with `OCI image manifest` support doesn't mean support `OCI artifact manifest` is supported.

So the copy could fail

specs/commandline/sign.md
FeynmanZhou: +1. ORAS CLI uses `--image -spec <spec version>-<manifest type>`  for this kind of scenario.

patrickzheng200:
@yizha1 Looks like this PR requires resolving conflicts now.

specs/commandline/sign.md
```diff
@@ -37,6 +37,48 @@ Flags:
...
+
+For example, use [ORAS][oras-land] to copy artifacts and associated signatures between two registries:
+
+```shell
+oras copy -r demo1.myregistry.io/net-monitor:v1 demo2.myregistry.io/net-monitor:v1
```
patrickzheng200, line 81: hmm, just a discussion, should we remove this part as this is a spec in Notation? It looks a bit strange to me that we are mentioning another tool in Notation's spec.
And as suggested above, I think we should move this whole section to another spec file instead of sign.md.

specs/commandline/sign.md
```diff
@@ -37,6 +37,48 @@ Flags:
...
+In case registries not listed in the page, user can consider using flag `--user-image-manifest` by checking the error message. Note that there is no deterministic way to determine a registry supporting `OCI artifact manifest` or not. The error message is just for reference. The following response status contained in error messages MAY indicate that the registries doesn't support `OCI artifact manifest`:
+
+- Response status `400 BAD Request` with error code `MANIFEST_INVALID` or `UNSUPPORTED`
+
+### Copy signatures between registries
+
+User may use tools to copy artifacts and associated signatures between two registries. The copy succeeds under the following conditions:
+
+- From registries that support `OCI artifact manifest` to registries that support `OCI artifact manifest`
+- From registries that support `OCI image manifest` to registries that support `OCI image manifest`
```
patrickzheng200, line 76: We might want to move the `Copy signatures between registries` section out of this spec, because it is named sign.md and it's for `notation sign` CLI command. Perhaps notaryproject repo is a better place for this section?

specs/commandline/sign.md
```diff
@@ -40,6 +41,34 @@ Flags:
   -m,  --user-metadata strings    {key}={value} pairs that are added to the signature payload
 ```
 
+## Use OCI image manifest to store signatures
+
+By default, Notation uses [OCI artifact manifest][oci-artifact-manifest] to store signatures in the registries. For backward compatibility, Notation supports using `OCI image manifest` to store the signature in the registries that implement partial of the [OCI Image specification v1.1][oci-image-spec]. Use flag `--image-spec v1.1-image` to force Notation store the signatures using OCI image manifest.
```
patrickzheng200, line 46:
```suggestion
By default, Notation uses [OCI artifact manifest][oci-artifact-manifest] to store signatures in registries. For backward compatibility, Notation also supports using `OCI image manifest` to store signatures in registries that implement partial of the [OCI Image specification v1.1][oci-image-spec]. Use flag `--image-spec v1.1-image` to force Notation to store signatures using OCI image manifest.
```

specs/commandline/sign.md
```diff
@@ -40,6 +41,34 @@ Flags:
...
+    "size": 2,
+    "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a"
+}
+```
+
+### When to use OCI image manifest
+
+[Registry support][registry-support] lists registries with different compatibilities. For registries not supporting `OCI artifact manifest`, user can use flag `--image-spec v1.1-image` to sign artifacts stored in those registries.
+
+In case registries not listed in the page, user can consider using flag `--image-spec v1.1-image` by checking the error message. Note that there is no deterministic way to determine a registry supporting `OCI artifact manifest` or not. The error message is just for reference. The following response status contained in error messages MAY indicate that the registries doesn't support `OCI artifact manifest`:
```
patrickzheng200, line 68:
```suggestion
For registries not listed in the page, users can consider using flag `--image-spec v1.1-image` by checking the error message. Note that there is no deterministic way to determine whether a registry supports `OCI artifact manifest` or not. The error message is just for reference. The following response status contained in error messages MAY indicate that the registry doesn't support `OCI artifact manifest`:
```

specs/commandline/sign.md
patrickzheng200, line 46: I feel like we should also mention that `Notation will not automatically convert between OCI artifact manifest and OCI image manifest on any occasion`.

specs/commandline/sign.md
```diff
@@ -40,6 +41,34 @@ Flags:
   -m,  --user-metadata strings    {key}={value} pairs that are added to the signature payload
 ```
 
+## Use OCI image manifest to store signatures
+
+By default, Notation uses [OCI artifact manifest][oci-artifact-manifest] to store signatures in the registries. For backward compatibility, Notation supports using `OCI image manifest` to store the signature in the registries that implement partial of the [OCI Image specification v1.1][oci-image-spec]. Use flag `--image-spec v1.1-image` to force Notation store the signatures using OCI image manifest.
+
+Registries MAY not implement or enable the `Referrers API`, which is used by the client to fetch the referrers. In the context of Notation, the referrers are signatures. Notation follows the fallback procedure defined in [OCI distribution spec][oci-backward-compatibility] if `Referrers API` is unavailable.
```
patrickzheng200, line 48:
```suggestion
Registries MAY not implement or enable the `Referrers API`, which is used by clients to fetch referrers. In the context of Notation, the referrers are signatures. Notation follows the fallback procedure defined in [OCI distribution spec][oci-backward-compatibility] if `Referrers API` is unavailable.
```

specs/commandline/sign.md
```diff
@@ -40,6 +41,34 @@ Flags:
...
+
+### Set config property for OCI image manifest
+
+OCI image manifest requires additional property `config` of type `descriptor`, which is not required by OCI artifact manifest. Notation creates a default config descriptor for the user if the flag `--image-spec v1.1-image` is used.
```
patrickzheng200, line 52:
```suggestion
OCI image manifest requires additional property `config` of type `descriptor`, which is not required by OCI artifact manifest. Notation creates a default config descriptor for the user if flag `--image-spec v1.1-image` is used.
```

specs/commandline/sign.md
```diff
@@ -40,6 +41,34 @@ Flags:
...
+"config": {
+    "mediaType": "application/vnd.cncf.notary.signature",
+    "size": 2,
+    "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a"
+}
+```
+
+### When to use OCI image manifest
+
+[Registry support][registry-support] lists registries with different compatibilities. For registries not supporting `OCI artifact manifest`, user can use flag `--image-spec v1.1-image` to sign artifacts stored in those registries.
```
patrickzheng200, line 66:
```suggestion
[Registry support][registry-support] lists registries with different compatibilities. For registries not supporting `OCI artifact manifest`, users can use flag `--image-spec v1.1-image` to sign artifacts stored in those registries.
```

patrickzheng200 (changes requested):

specs/commandline/sign.md
```diff
@@ -31,6 +31,7 @@ Usage:
 Flags:
   -e,  --expiry duration          optional expiry that provides a "best by use" time for the artifact. The duration is specified in minutes(m) and/or hours(h). For example: 12h, 30m, 3h20m
   -h,  --help                     help for sign
+       --image-spec string        specify the manifest type for signatures. options: v1.1-artifact, v1.1-image (default: v1.1-artifact)
```
patrickzheng200, line 34: To align with descriptions of other flags under this command:
```suggestion
       --image-spec string       manifest type for signatures. options: v1.1-artifact, v1.1-image (default: v1.1-artifact)
```

specs/commandline/sign.md
priteshbandi: yeah ```--image-spec``` looks good to me. 

@iamsamirzon would like to get your opinion here

priteshbandi (approved):
LGTM

specs/commandline/sign.md
priteshbandi: IMO we shouldn't add that because thats implicit from explicitly using a flag. Also, when we do automatic fallback, we will need to delete it and which we might forget

yizha1: Keep the wording not changed for now. We can add the restriction in the future if needed.

shizhMSFT (approved):
LGTM

patrickzheng200 (approved):
LGTM

PR #507, opened 2023-01-13
feat: add support for signed user metadata in notation sign and verify cmds

byronchien:
Adds support for signed user metadata in `notation sign` and `notation verify`. [Relevant spec](https://github.com/notaryproject/notation/pull/498)

*This PR depends on https://github.com/notaryproject/notation-go/pull/242* please review [notation-go/pull/242](https://github.com/notaryproject/notation-go/pull/242) first

example sign usage:
```
chienb@a07817b52895 notation % notation sign $IMAGE --user-metadata io.wabbit-networks.buildId=123 --user-metadata io.wabbit-networks.buildTime=123
Warning: Always sign the artifact using digest(`@sha256:...`) rather than a tag(`:v1`) because tags are mutable and a tag reference can point to a different artifact than the one signed.
Resolved artifact tag `v1` to digest `sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b` before signing.
Successfully signed localhost:5000/net-monitor@sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b

```

example metadata displayed on verification (without metadata flag)
```
chienb@a07817b52895 notation % notation verify $IMAGE
Resolved artifact tag `v1` to digest `sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b` before verification.
Successfully verified signature for localhost:5000/net-monitor@sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b

The artifact was signed with the following user metadata.

KEY                          VALUE
io.wabbit-networks.buildTime   123
io.wabbit-networks.buildId     123
```

example verification:
```
chienb@a07817b52895 notation % notation verify $IMAGE --user-metadata io.wabbit-networks.buildTime=123
Resolved artifact tag `v1` to digest `sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b` before verification.
Successfully verified signature for localhost:5000/net-monitor@sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b

The artifact was signed with the following user metadata.

KEY                            VALUE
io.wabbit-networks.buildTime   123
io.wabbit-networks.buildId     123
```

example verification failure
```
chienb@a07817b52895 notation % notation verify $IMAGE --user-metadata foo=bar
Resolved artifact tag `v1` to digest `sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b` before verification.
Error: signature verification failed: signature verification failed
```

Not in this PR: error messaging. In the spec update, we mentioned that there should be an error message if the verification fails due to the metadata not being present, but if there are multiple verifications that fail for different reasons, is there a desired order to prioritize the what the end error message displayed is?

Signed-off-by: Byron Chien <chienb@amazon.com>


byronchien:
related: https://github.com/notaryproject/notation-go/pull/242

internal/cmd/flags.go
```diff
@@ -71,6 +71,18 @@ var (
...
 		fs.StringArrayVarP(p, PflagPluginConfig.Name, PflagPluginConfig.Shorthand, nil, PflagPluginConfig.Usage)
 	}
+
+	PflagUserMetadata = &pflag.Flag{
+		Name:      "user-metadata",
+		Shorthand: "m",
+	}
+
+	PflagUserMetadataSignUsage = "example sign usage"
+	PflagUserMetadataVerifyUsage = "example verify usage"
```
priteshbandi, line 81: Missing actual usage strings?

cmd/notation/verify.go
```diff
@@ -148,3 +160,20 @@ func resolveReference(ctx context.Context, opts *SecureFlagOpts, reference strin
...
+	metadata, _ := outcome.GetUserMetadata()
+
+	if len(metadata) > 0 {
+		fmt.Println("\nThe artifact was signed with the following user metadata.")
+		tw := tabwriter.NewWriter(os.Stdout, 0, 0, 3, ' ', 0)
+		fmt.Fprintln(tw, "\nKEY\tVALUE\t")
+
+		for k, v := range metadata {
+			fmt.Fprintf(tw, "%v\t%v\t\n", k, v)
+		}
```
priteshbandi, line 175: should we move this logic to below file ? https://github.com/notaryproject/notation/blob/767f8ae4a762541dc6bacb531d65d38d212734c9/internal/ioutil/print.go#L11

cmd/notation/verify.go
```diff
@@ -148,3 +160,20 @@ func resolveReference(ctx context.Context, opts *SecureFlagOpts, reference strin
 
 	return ref, nil
 }
+
+func printMetadataIfPresent(outcome *notation.VerificationOutcome) {
+	// ignoring error, getting metadata for a successful outcome
+	metadata, _ := outcome.GetUserMetadata()
```
priteshbandi, line 166: Why are we ignoring error here? We are not calling `GetUserMetadata` anywhere in verification so a code bug can still lead to a error.

cmd/notation/sign.go
```diff
@@ -121,6 +127,7 @@ func prepareSigningContent(ctx context.Context, opts *signOpts, sigRepo notation
 		SignatureMediaType: mediaType,
 		ExpiryDuration:     opts.expiry,
 		PluginConfig:       pluginConfig,
+		UserMetadata: userMetadata,
```
priteshbandi, line 130: nit: indentation

cmd/notation/verify.go
```diff
@@ -97,17 +107,15 @@ func runVerify(command *cobra.Command, opts *verifyOpts) error {
...
 	// core verify process
 	_, outcomes, err := notation.Verify(ctx, verifier, sigRepo, verifyOpts)
 	// write out on failure
 	if err != nil || len(outcomes) == 0 {
 		if err != nil {
-			var errorVerificationFailed *notation.ErrorVerificationFailed
-			if !errors.As(err, &errorVerificationFailed) {
-				return fmt.Errorf("signature verification failed: %w", err)
-			}
+			return fmt.Errorf("signature verification failed: %w", err)
```
priteshbandi, line 118: whats the implication of this change?

byronchien: if there's any error coming from notation.Verify (and not just ErrorVerificationFailed), the message is returned to the user. looking at this again, there's a bug since if `len(outcomes) == 0`, err can't be nil, so this can be simplified. overall, the implication is that error messages returned from notation.Verify will be displayed to users.

patrickzheng200: > if there's any error coming from notation.Verify (and not just ErrorVerificationFailed), the message is returned to the user. looking at this again, there's a bug since if `len(outcomes) == 0`, err can't be nil, so this can be simplified. overall, the implication is that error messages returned from notation.Verify will be displayed to users.

Looks like this is changing the original output behavior.
The original output design:
1. `err != nil`
       `err is NOT notation.ErrorVerificationFailed: "signature verification failed: err"`
       `err is notation.ErrorVerificationFailed: "signature verification failed for all the signatures associated with ref"`
2. `len(outcomes) == 0` (IMO, we shouldn't rely on `if len(outcomes) == 0, err can't be nil` as this is a logic from the underneath library. If the output pattern in notation-go changed for some reason, then our assumption here will no longer be true.)
       `"signature verification failed for all the signatures associated with ref"`

After this change:
1. `err != nil`
        `"signature verification failed: err"`
2. `len(outcomes) == 0`
        `"signature verification failed for all the signatures associated with ref"`

Note, if `err != nil AND err is notation.ErrorVerificationFailed`, this means that we have gone through all the signatures associated with the artifact and all the signatures failed verification. In this case, we do not need to print out the `err` itself, as the error msg is just `"signature verification failed"`.

byronchien: ah, ok then, can match the original output design then

priteshbandi (commented):
LGTM

priteshbandi (approved):

JeyJeyGao:
The pipeline was failed. Please update the code.

priteshbandi:
> The pipeline was failed. Please update the code.

This PR depends on https://github.com/notaryproject/notation-go/pull/242 and that's the reason build is failing.

cmd/notation/sign.go
```diff
@@ -98,30 +100,34 @@ func runSign(command *cobra.Command, cmdOpts *signOpts) error {
...
+		return notation.RemoteSignOptions{}, registry.Reference{}, err
 	}
 
+	signOpts := notation.RemoteSignOptions{}
+	signOpts.ArtifactReference = ref.String()
+	signOpts.SignatureMediaType = mediaType
+	signOpts.ExpiryDuration = opts.expiry
+	signOpts.PluginConfig = pluginConfig
+	signOpts.UserMetadata = userMetadata
+
```
patrickzheng200, line 131: How about:
```suggestion
	signOpts := notation.RemoteSignOptions{
	    ArtifactReference: ref.String(),
	    SignatureMediaType: mediaType,
	    ExpiryDuration: opts.expiry,
	    PluginConfig: pluginConfig,
	    UserMetadata: userMetadata,
	}
```

patrickzheng200 (commented):
LGTM with one comment.

patrickzheng200 (approved):
LGTM

cmd/notation/sign.go
byronchien: i changed it, but this breaks the tests:
```

cmd/notation/sign.go:103: undefined: notation.RemoteSignOptions
cmd/notation/sign.go:109: undefined: notation.RemoteSignOptions
cmd/notation/sign.go:114: undefined: notation.RemoteSignOptions
cmd/notation/sign.go:118: undefined: notation.RemoteSignOptions
cmd/notation/sign.go:122: undefined: notation.RemoteSignOptions
cmd/notation/sign.go:125: undefined: notation.RemoteSignOptions
```

seems like initialization of embedded structs should be

```
notation.RemoteSignOptions{
    SignOptions: notation.SignOptions{
        ...
    }
}
```

internal/cmd/flags.go
```diff
@@ -71,6 +71,18 @@ var (
...
+		Name:      "user-metadata",
+		Shorthand: "m",
+	}
+
+	PflagUserMetadataSignUsage = "{key}={value} pairs that are added to the signature payload"
+	PflagUserMetadataVerifyUsage = "user defined {key}={value} pairs that must be present in the signature for successful verification if provided"
+
+	SetPflagUserMetadata = func(fs *pflag.FlagSet, p *[]string, usage string) {
+		fs.StringArrayVarP(p, PflagUserMetadata.Name, PflagUserMetadata.Shorthand, nil, usage)
+	}
```
shizhMSFT, line 85: nit: Do you want to remove empty lines in between just to make it a code block and consistent?

cmd/notation/verify.go
```diff
@@ -148,3 +160,13 @@ func resolveReference(ctx context.Context, opts *SecureFlagOpts, reference strin
...
 }
+
+func printMetadataIfPresent(outcome *notation.VerificationOutcome) {
+	// the signature envelope is parsed as part of verification.
+	// since user metadata is only printed on successful verification,
+	// this error can be ignored
+	metadata, _ := outcome.UserMetadata()
+
+	fmt.Println("\nThe artifact was signed with the following user metadata.")
+	ioutil.PrintMetadataMap(os.Stdout, metadata)
```
shizhMSFT, line 171: What if `metadata` is empty?

byronchien: added a check on size

yizha1:
@byronchien  could you fix the conflict?

byronchien:
resolved conflict

PR #509, opened 2023-01-16
feat: support OCI image manifest

patrickzheng200:
This PR adds OCI image manifest support to notation CLI according to this [spec](https://github.com/notaryproject/notation/blob/main/specs/commandline/sign.md).

If using OCI artifact manifest to store signatures, notation requires the registry to support Referrers API as well.

Tested with [COSE, JWS] + [OCI artifact manifest, OCI image manifest] on ACR.

This PR would also resolve [issue](https://github.com/notaryproject/notaryproject/issues/219).


cmd/notation/registry.go
```diff
@@ -41,7 +41,7 @@ func getRegistryClient(ctx context.Context, opts *SecureFlagOpts, serverAddress
 	return reg, nil
 }
 
-func getRepositoryClient(ctx context.Context, opts *SecureFlagOpts, ref registry.Reference) (notationregistry.Repository, error) {
+func getRepositoryClient(ctx context.Context, opts *SecureFlagOpts, ref registry.Reference, ociImageManifest bool) (notationregistry.Repository, error) {
```
JeyJeyGao, line 44: Can we add a comment to mention that the `ociImageManifest` only influences on signing logic but not the verifying and listing logic.

cmd/notation/sign.go
```diff
@@ -66,10 +76,16 @@ Example - Sign an OCI artifact stored in a registry and specify the signature ex
...
+	command.Flags().StringVar(&opts.imageSpec, "image-spec", artifactManifest, "manifest type for signatures. options: v1.1-artifact, v1.1-image")
 	return command
 }
 
 func runSign(command *cobra.Command, cmdOpts *signOpts) error {
+	// sanity check
+	if !validateImageSpec(cmdOpts.imageSpec) {
+		return fmt.Errorf("image spec has to be in %v, got %s", supportedImageSpec, cmdOpts.imageSpec)
+	}
+
```
JeyJeyGao, line 88: Can move the section out of runSign() to make the function clear. What about move it to

```go
            // Here
			return runSign(cmd, opts)
		},
```

cmd/notation/sign.go
```diff
@@ -10,18 +10,28 @@ import (
...
 	"github.com/notaryproject/notation/internal/envelope"
+	"github.com/notaryproject/notation/internal/slices"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/spf13/cobra"
 	"oras.land/oras-go/v2/registry"
 )
 
+const (
+	artifactManifest = "v1.1-artifact"
+
```
JeyJeyGao, line 21: do we need the empty line?

cmd/notation/registry.go
patrickzheng200: It's an unexported helper function though. Will add a description comment for this argument.

cmd/notation/registry.go
```diff
@@ -55,6 +59,56 @@ func getRepositoryClient(ctx context.Context, opts *SecureFlagOpts, ref registry
...
+		Reference: ref,
+		PlainHTTP: plainHTTP,
+	}
+	repositoryOpts := notationregistry.RepositoryOptions{
+		OCIImageManifest: ociImageManifest,
+	}
+	// using OCI artifact manifest to store signatures. This requires the
+	// existence of Referrers API for Sign process.
+	if !ociImageManifest {
+		if err := remoteRepo.SetReferrersCapability(true); err != nil {
```
JeyJeyGao, line 94: Why not let oras-go automatically determine whether use referrers api or not?

patrickzheng200: Because oras-go does not have this capability yet. Notation needs to explicitly ping Referrers API.

specs/commandline/sign.md
```diff
@@ -29,16 +29,18 @@ Usage:
...
-  -p,  --password string          password for registry operations (default to $NOTATION_PASSWORD if not specified)
-       --plain-http               registry access via plain HTTP
-       --plugin-config strings    {key}={value} pairs that are passed as it is to a plugin, refer plugin's documentation to set appropriate values
-       --signature-format string  signature envelope format, options: 'jws', 'cose' (default "jws")
-  -u,  --username string          username for registry operations (default to $NOTATION_USERNAME if not specified)
-  -m,  --user-metadata strings    {key}={value} pairs that are added to the signature payload
+  -d,  --debug                       debug mode
+  -e,  --expiry duration             optional expiry that provides a "best by use" time for the artifact. The duration is specified in minutes(m) and/or hours(h). For example: 12h, 30m, 3h20m
+  -h,  --help                        help for sign
+       --artifact-type string        manifest type for signatures. options: v1.1-artifact, v1.1-image (default "v1.1-artifact")
```
shizhMSFT, line 35: According to the spec, should it be `--image-spec`?

internal/slices/slices.go
```diff
@@ -25,3 +25,13 @@ func Contains[E isser](s []E, name string) bool {
 func Delete[S ~[]E, E isser](s S, i int) S {
 	return append(s[:i], s[i+1:]...)
 }
+
+// Contains reports whether v is present in s.
+func ContainsGenerics[E comparable](s []E, v E) bool {
```
shizhMSFT, line 30: nit: The function name is acceptable but still weird. It should named as `Contains` but the name is occupied. How about `ContainerElement`?

cmd/notation/registry.go
```diff
@@ -55,6 +59,56 @@ func getRepositoryClient(ctx context.Context, opts *SecureFlagOpts, ref registry
...
+	authClient, plainHTTP, err := getAuthClient(ctx, opts, ref)
+	if err != nil {
+		return nil, err
+	}
+
+	remoteRepo := &remote.Repository{
+		Client:    authClient,
+		Reference: ref,
+		PlainHTTP: plainHTTP,
+	}
```
shizhMSFT, line 87: This block of code seems duplicated. Can we refactor a bit?

cmd/notation/sign.go
```diff
@@ -10,18 +10,27 @@ import (
...
 	"github.com/notaryproject/notation/internal/envelope"
+	"github.com/notaryproject/notation/internal/slices"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/spf13/cobra"
 	"oras.land/oras-go/v2/registry"
 )
 
+const (
+	artifactManifest = "v1.1-artifact"
+	imageManifest    = "v1.1-image"
```
shizhMSFT, line 21: nit:
```suggestion
	imageSpecArtifact = "v1.1-artifact"
	imageSpecImage    = "v1.1-image"
```

cmd/notation/sign.go
```diff
@@ -78,7 +92,11 @@ func runSign(command *cobra.Command, cmdOpts *signOpts) error {
 	if err != nil {
 		return err
 	}
-	sigRepo, err := getSignatureRepository(ctx, &cmdOpts.SecureFlagOpts, cmdOpts.reference)
+	var ociImageManifest bool
+	if cmdOpts.imageSpec == imageManifest {
+		ociImageManifest = true
+	}
```
shizhMSFT, line 98: nit: code can be simplified
```suggestion
	ociImageManifest := cmdOpts.imageSpec == imageManifest
```

cmd/notation/sign.go
```diff
@@ -58,6 +67,10 @@ Example - Sign an OCI artifact stored in a registry and specify the signature ex
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			// sanity check
+			if !validateImageSpec(opts.imageSpec) {
+				return fmt.Errorf("image spec has to be in %v, got %s", supportedImageSpec, opts.imageSpec)
```
priteshbandi, line 72:
```suggestion
				return fmt.Errorf("image spec must be one of the following %v but got %s", supportedImageSpec, opts.imageSpec)
```

cmd/notation/sign_test.go
```diff
@@ -23,6 +23,7 @@ func TestSignCommand_BasicArgs(t *testing.T) {
 			Key:             "key",
 			SignatureFormat: envelope.JWS,
 		},
+		imageSpec: artifactManifest,
```
priteshbandi, line 26: We should be consistent on using `artifact-type` or `imageSpec` everywhere.

priteshbandi (commented):
This PR contains changes only for [sign](https://github.com/notaryproject/notaryproject/issues/219) command, will there be a separate PR for [verify](https://github.com/notaryproject/notaryproject/issues/220) command? 
Or is that not required because oras-go already handles fallback for verify ?

cmd/notation/registry.go
JeyJeyGao: https://github.com/oras-project/oras-go/blob/main/registry/remote/repository.go#L532
I think oras-go has the capability, right?

specs/commandline/sign.md
yizha1: @priteshbandi I recall that you did agree with the name `--image-spec` during spec review, see https://github.com/notaryproject/notation/pull/502#discussion_r1069095292.
It is more flexible and extensible by this design `--image-spec <spec version>-<manifest type>`  that users can specify the spec version and the manifest type that they want to use. There could be new types in the future. On the other hand, the name of `--artifact-type` could cause confusion, since there is a field named `artifactType` defined in `OCI artifact manifest` spec. User may think this flag `--artifact-type` is to be used to configure the filed `artifactType`.
I suggest we keep the current definition in the spec.

internal/slices/slices.go
JeyJeyGao, line 30: Can we add a unit test for it, so it will not reduce the code coverage.

cmd/notation/registry.go
```diff
@@ -55,6 +59,56 @@ func getRepositoryClient(ctx context.Context, opts *SecureFlagOpts, ref registry
...
+	repositoryOpts := notationregistry.RepositoryOptions{
+		OCIImageManifest: ociImageManifest,
+	}
+	// using OCI artifact manifest to store signatures. This requires the
+	// existence of Referrers API for Sign process.
+	if !ociImageManifest {
```
Wwwsylvia, line 93: Could you help me to understand the scenarios here? Should we list the valid scenarios in the comment? 
From the code it looks like:
1. If OCI artifact manifest is used
  - Referrers API must be available
  - Must not fallback to referrers tag schema (why?)

2. If OCI Image manifest is used
  - Referrers API and tag schema are both OK (?)

patrickzheng200: This change is based on this community [discussion](https://cloud-native.slack.com/archives/CQUH8U287/p1674050115958189?thread_ts=1673362612.471549&cid=CQUH8U287) that notation will not implement hybrid support, for instance, a registry supports OCI artifact manifest but not the Referrers API. 
Therefore, when user wants to store signatures in OCI artifact manifest, we require the Referrers API as well (and this is the default behavior of notation). 
On the other hand, I haven't seen a registry that supports Referrers API but only OCI image manifest. Hence, when user chooses OCI image manifest, notation could utilize oras's fallback logic and use the tag schema. But to be safer, yes, I think setting `remoteRepo.SetReferrersCapability(false)` on OCI image manifest is a better way here. I will update the code to reflect this.

cmd/notation/registry.go
patrickzheng200: Currently, oras-go does not have this capability yet. See https://github.com/oras-project/oras-go/issues/423 for more details.

cmd/notation/registry.go
Wwwsylvia: Thanks @patrickzheng200  for the explanation.

> But to be safer, yes, I think setting remoteRepo.SetReferrersCapability(false) on OCI image manifest is a better way here.

To my understanding, it seems that it won't hurt to use the Referrers API for OCI image manifests. But if the Referrers API is known to be unavailable, setting `remoteRepo.SetReferrersCapability(false)` will bring better performance since the call to Referrers API will be saved.

specs/commandline/sign.md
patrickzheng200: I changed back to `--image-spec` for now based on comments above.

patrickzheng200:
> This PR contains changes only for [sign](https://github.com/notaryproject/notaryproject/issues/219) command, will there be a separate PR for [verify](https://github.com/notaryproject/notaryproject/issues/220) command? Or is that not required because oras-go already handles fallback for verify ?

Yes, this change only take effect on the Sign process. For Verification, we should already have the correct logic based on community discussion.

cmd/notation/registry.go
patrickzheng200: > Thanks @patrickzheng200 for the explanation.
> 
> > But to be safer, yes, I think setting remoteRepo.SetReferrersCapability(false) on OCI image manifest is a better way here.
> 
> To my understanding, it seems that it won't hurt to use the Referrers API for OCI image manifests. But if the Referrers API is known to be unavailable, setting `remoteRepo.SetReferrersCapability(false)` will bring better performance since the call to Referrers API will be saved.

Just had another discussion with Shiwei, Notation would still enforce two paths during Signing:
1. OCI artifact manifest uses the Referrers API
2. OCI image manifest uses the Referrers Tag Schema
(So basically, we do not support hybrid case on both ways.)

cmd/notation/verify.go
```diff
@@ -56,7 +56,7 @@ Example - Verify a signature on an OCI artifact identified by a tag  (Notation w
 	}
 	opts.LoggingFlagOpts.ApplyFlags(command.Flags())
 	opts.SecureFlagOpts.ApplyFlags(command.Flags())
-	command.Flags().StringArrayVarP(&opts.pluginConfig, "plugin-config", "c", nil, "{key}={value} pairs that are passed as it is to a plugin, if the verification is associated with a verification plugin, refer plugin documentation to set appropriate values")
+	command.Flags().StringArrayVar(&opts.pluginConfig, "plugin-config", nil, "{key}={value} pairs that are passed as it is to a plugin, if the verification is associated with a verification plugin, refer plugin documentation to set appropriate values")
```
JeyJeyGao, line 59: why the shorthand "c" was removed?

patrickzheng200: This was a suggestion by @priteshbandi in this [comment](https://github.com/notaryproject/notation/pull/509#discussion_r1081951504).

JeyJeyGao: If we have multiple plugin config key=value pairs, a shorthand is helpful. 
```
notation sign xxxxx \
-c key1=value1
-c key2=value2
```
@priteshbandi is this a rational use case that we need the shorthand for plugin-config flag?

cmd/notation/registry.go
```diff
@@ -52,7 +91,20 @@ func getRepositoryClient(ctx context.Context, opts *SecureFlagOpts, ref registry
 		Reference: ref,
 		PlainHTTP: plainHTTP,
 	}
-	return notationregistry.NewRepository(remoteRepo), nil
+	return remoteRepo, nil
```
shizhMSFT, line 94: nit: code
```go
	remoteRepo := &remote.Repository{
		Client:    authClient,
		Reference: ref,
		PlainHTTP: plainHTTP,
	}
	return remoteRepo, nil
```
can be simplified as
```go
	return &remote.Repository{
		Client:    authClient,
		Reference: ref,
		PlainHTTP: plainHTTP,
	}, nil
```

shizhMSFT (approved):
LGTM

specs/commandline/sign.md
jlbutler: FWIW, I agree with @priteshbandi on this. I think users specifying both an image spec version and artifact type feels like too much implementation detail exposure.

patrickzheng200: @priteshbandi @jlbutler Another option here is making it a boolean flag, for example, `--legacy-artifact-type`. Omitting the flag means using OCI artifact manifest to store signatures. Setting the flag means using OCI image manifest. What do you think?

cmd/notation/registry.go
Wwwsylvia: Got it!

cmd/notation/verify.go
JeyJeyGao: Update: I think "--plugin-config" is fine.

specs/commandline/sign.md
FeynmanZhou: I vote for `--image-spec <spec version>-<manifest type>`. By specifying the OCI image spec version, it will be much clear as the artifact/image manifest may differentiate by spec versions.

Just FYI, ORAS has a similar situation with specifying a tag to custom the pushing behavior of the manifest type, see https://github.com/oras-project/oras/blob/main/docs/proposals/compatibility-mode.md#force-uploading-a-specific-manifest-type-using-a-flag---image-spec

priteshbandi (approved):
LGTM with pending flag name [discussion](https://cloud-native.slack.com/archives/CQUH8U287/p1675391977693439). For merge, I would suggest wait for that discussion to conclude and then based on it update the flag name.

specs/commandline/sign.md
toddysm: I think both `--image-spec` and `--artifact-type` as flag are confusing. I am struggling to understand the meaning of both. Also, from the help text above the expectation seems to be that this parameter will expect some value like `v1.0` or `v1.1`. This is too much details for somebody who doesn't even know what is the difference between OCI 1.0 and OCI 1.1 or has no knowledge of how different registries store the signature. 

The flow is the following:
- User tries to store the signature into the registry and gets an error.
- Based on the descriptiveness of the error, the user may want to retry the operation with the switch

There are two parts here:
- First, we need to make sure that the error the user sees is descriptive enough so they know what is happening. I would expect to get some error like: **Saving the signature failed. It is possible the registry doesn't support the format the signature is stored. Try --THE_SWITCH to use the legacy format for storing the signature**
- The name of the switch should be more user friendly and not requiring the user to know anything about the OCI spec. What comes to my mind is something like `--use-legacy-format`

yizha1: @toddysm Thanks. It makes sense. There is an existing flag named `--signature-format` for `JWS` or `COSE` envelope. So, it's better not to use the word `format` to avoid confusion. How about `--use-legacy-manifest`. Here is the summary for current proposals:
1. flag name `--signature-manifest`, values could be `1.0` or `1.1`
2. flag name `--signature-manifest`, values could be `image` or `artifact`
3. flag name `--use-legacy-format`, this is a Boolean type of flag
4. flag name `--use-legacy-manifest`, this is a Boolean type of flag

yizha1: One more thing, is it appropriate to use the wording `legacy` before OCI spec cut a 1.1 release?

toddysm: I see. So, what about `--use-image-manifest` Boolean flag? This is quite explicit, and it doesn't require deep knowledge of the OCI spec versioning.

lachie83: General agreement on the Notary community call on Feb 6 was to proceed with option 2 above using string values of `image` and `artifact`, `artifact` will be the default. cc @patrickzheng200

patrickzheng200:
The current build would fail due to dependency change in notation-go: https://github.com/notaryproject/notation-go/pull/271

Please review https://github.com/notaryproject/notation-go/pull/271 first.

cmd/notation/registry.go
```diff
@@ -3,56 +3,110 @@ package main
...
+
+	// Notation enforces the following two paths during Sign process:
+	// 1. OCI artifact manifest uses the Referrers API
+	// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers
+	// 2. OCI image manifest uses the Referrers Tag Schema
+	// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#referrers-tag-schema
+	if !ociImageManifest {
+		logger.Info("Use OCI artifact manifest and Referrers API to store signature")
+		var checkReferrerDesc ocispec.Descriptor
+		checkReferrerDesc.Digest = zeroDigest
```
JeyJeyGao, line 68: Looks like it was not used?

patrickzheng200:
Dependency updated. PR ready to be reviewed/merged.

cmd/notation/registry.go
patrickzheng200: Removed.

cmd/notation/sign.go
```diff
@@ -50,6 +59,12 @@ Example - Sign an OCI artifact identified by a tag (Notation will resolve tag to
 
 Example - Sign an OCI artifact stored in a registry and specify the signature expiry duration, for example 24 hours
   notation sign --expiry 24h <registry>/<repository>@<digest>
+
+Example - Sign an OCI artifact and use OCI image manifest to store the signature, with the default JWS envelope:
+  notation sign --signature-manifest image <registry>/<repository>@<digest>
+
+Example - Sign an OCI artifact and use OCI image manifest to store the signature, with the COSE envelope:
+  notation sign --signature-manifest image --signature-format cose <registry>/<repository>@<digest>
```
priteshbandi, line 67: IMO we dont need to show examples for all the signature envelope

```suggestion
Example - Sign an OCI artifact and use OCI image manifest to store the signature:
  notation sign --signature-manifest image <registry>/<repository>@<digest>
```

cmd/notation/registry.go
```diff
@@ -127,3 +179,36 @@ func getSavedCreds(ctx context.Context, serverAddress string) (auth.Credential,
...
+		if !errors.As(err, &errResp) || errResp.StatusCode != http.StatusNotFound {
+			return err
+		}
+		if isErrorCode(errResp, errcode.ErrorCodeNameUnknown) {
+			// The repository is not found.
+			return err
+		}
+		// A 404 returned by Referrers API indicates that Referrers API is
+		// not supported.
+		errMsg := fmt.Sprintf("failed to ping Referrers API with error: %v. This is due to target registry does not support the Referrers API. Try store signatures with OCI image manifest using `--signature-manifest image`", err)
```
shizhMSFT, line 204: The error message can be simplified as
```suggestion
		errMsg := "Target registry does not support the Referrers API. Try the flag `--signature-manifest image` for backwards compatibility", err)
```
where the detailed error message can be in the log.

cmd/notation/sign.go
```diff
@@ -91,6 +112,10 @@ func runSign(command *cobra.Command, cmdOpts *signOpts) error {
 	// core process
 	_, err = notation.Sign(ctx, signer, sigRepo, opts)
 	if err != nil {
+		var errorPushSignatureFailed notation.ErrorPushSignatureFailed
+		if errors.As(err, &errorPushSignatureFailed) {
+			return fmt.Errorf("%v. This might be due to target registry does not support OCI artifact manifest. Try store signatures with OCI image manifest using `--signature-manifest image`", err)
```
shizhMSFT, line 117: nit: The error message can be simplified.
```suggestion
			return fmt.Errorf("%v. Target registry seems not support OCI artifact manifest. Try the flag `--signature-manifest image` to store signatures using OCI image manifest", err)
```

priteshbandi: +1,

`Target registry does not seem to support OCI artifact manifest. Try the flag `--signature-manifest image` to store signatures using OCI image manifest`

cmd/notation/registry.go
priteshbandi: To be consistent should we use `Try the flag --signature-manifest image to store signatures using OCI image manifest` here ?

cmd/notation/registry.go
```diff
@@ -127,3 +179,36 @@ func getSavedCreds(ctx context.Context, serverAddress string) (auth.Credential,
...
+		return nil
+	})
+	if err != nil {
+		var errResp *errcode.ErrorResponse
+		if !errors.As(err, &errResp) || errResp.StatusCode != http.StatusNotFound {
+			return err
+		}
+		if isErrorCode(errResp, errcode.ErrorCodeNameUnknown) {
+			// The repository is not found.
+			return err
```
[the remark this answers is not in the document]
shizhMSFT: We probably need to have a comment block to describe this path.

cmd/notation/registry.go
patrickzheng200: I think we could combine both, i.e. `Try the flag --signature-manifest image to store signatures using OCI image manifest for backwards compatibility`. The benefit to include `for backwards compatibility` is that a user might not know what `OCI image manifest` is. Adding `for backwards compatibility` shows the purpose behind this flag.

shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM, you will need to resolve conflict before change can be merged

PR #517, opened 2023-01-30
fix: optimize login `--password-stdin` flag behavior

JeyJeyGao:
Previously, when login with `--password-stdin` user needs to specify the EOF of the input by Ctrl+d and the password is shown on the screen, which is not an ideal behavior.

This PR fixed the issue by using term.ReadPassword() to hide the password and only read one line.

### DIFF
--- a/cmd/notation/login.go
+++ b/cmd/notation/login.go
@@ -7,10 +7,12 @@ import (
 	"io"
 	"os"
 	"strings"
+	"syscall"
 
 	"github.com/notaryproject/notation/internal/cmd"
 	"github.com/notaryproject/notation/pkg/auth"
 	"github.com/spf13/cobra"
+	"golang.org/x/term"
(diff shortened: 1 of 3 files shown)


cmd/notation/login.go
```diff
@@ -108,11 +110,12 @@ func newCredentialFromInput(username, password string) orasauth.Credential {
 
 func readPassword(opts *loginOpts) error {
 	if opts.passwordStdin {
-		password, err := readLine()
+		fmt.Print("Enter password: ")
+		bytePass, err := term.ReadPassword(syscall.Stdin)
```
priteshbandi, line 114: Should we do something like this?
```
var bytePass []byte
if !term.IsTerminal(int(os.Stdin.Fd())) {
		   return "", fmt.Errorf("error reading input for password")  
}
bytePass, err := term.ReadPassword(int(os.Stdin.Fd())
```

PR #111, opened 2021-10-20
Alpha.1 release updates

SteveLasker:
Updates the docs for Alpha.1

PR #527, opened 2023-02-02
feat: add support for json output for `notation verify`

byronchien:
allows json output for `notation verify`.

fixes https://github.com/notaryproject/roadmap/issues/67, https://github.com/notaryproject/notation/pull/498

Example output:
```
chienb@a07817b52895 notation % ./bin/notation verify $IMAGE --output json
{
    "reference": "localhost:5000/net-monitor@sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b",
    "userMetadata": {
        "foo": "bar"
    },
    "result": "Success"
}
```

Note: PR is on top of changes that haven't been merged into main yet, so there's duplicate code from [this PR adding metadata support](https://github.com/notaryproject/notation/pull/507).

suggested order for review: [notation-go #261](https://github.com/notaryproject/notation-go/pull/261) => [notation-core-go #111](https://github.com/notaryproject/notation-core-go/pull/111) => [notation #527](https://github.com/notaryproject/notation/pull/527) (this one) =>[notation #528](https://github.com/notaryproject/notation/pull/528)


byronchien:
order to review:
https://github.com/notaryproject/notation/pull/507
https://github.com/notaryproject/notation/pull/527
https://github.com/notaryproject/notation/pull/528

cmd/notation/verify.go
```diff
@@ -72,8 +84,10 @@ func runVerify(command *cobra.Command, opts *verifyOpts) error {
 
 	// resolve the given reference and set the digest
 	ref, err := resolveReference(command.Context(), &opts.SecureFlagOpts, reference, sigRepo, func(ref registry.Reference, manifestDesc ocispec.Descriptor) {
-		fmt.Printf("Resolved artifact tag `%s` to digest `%s` before verification.\n", ref.Reference, manifestDesc.Digest.String())
-		fmt.Println("Warning: The resolved digest may not point to the same signed artifact, since tags are mutable.")
+		if opts.outputFormat == ioutil.OutputPlaintext {
+			fmt.Printf("Resolved artifact tag `%s` to digest `%s` before verification.\n", ref.Reference, manifestDesc.Digest.String())
+			fmt.Println("Warning: The resolved digest may not point to the same signed artifact, since tags are mutable.")
```
patrickzheng200, line 89: Due to the merge of PR [497](https://github.com/notaryproject/notation/pull/497), we might need to reword these two printouts.

specs/commandline/verify.md
```diff
@@ -37,6 +37,7 @@ Usage:
 
 Flags:
   -h,  --help                    help for verify
+  -o,  --output string           output format, options: 'json', 'text' (default "text")
```
patrickzheng200, line 40: @yizha1 @FeynmanZhou This PR is adding a new flag called `--output` into the `notation verify` command. Just want to make sure we are on the same page. And does the flag description look good to you?
And just an open discussion, since this flag is asking the user to type `-o json`, an alternative is to make it a Boolean flag with name `--output-json`. An auto-completion would be feasible here and save some user's typing. What do you think?

cmd/notation/verify.go
```diff
@@ -148,3 +171,31 @@ func resolveReference(ctx context.Context, opts *SecureFlagOpts, reference strin
 
 	return ref, nil
 }
+
+func printResult(outputFormat, reference string, outcome *notation.VerificationOutcome) error {
+
```
patrickzheng200, line 176: nit: can remove extra line

cmd/notation/verify.go
```diff
@@ -148,3 +171,31 @@ func resolveReference(ctx context.Context, opts *SecureFlagOpts, reference strin
...
+	if reflect.DeepEqual(outcome.VerificationLevel, trustpolicy.LevelSkip) {
+		if outputFormat == ioutil.OutputPlaintext {
+			fmt.Println("Trust policy is configured to skip signature verification for", reference)
+			return nil
+		} else {
+			output := verifyOutput{Reference: reference, Result: "Skipped", UserMetadata: map[string]string{}}
+			return ioutil.PrintObjectAsJson(output)
+		}
```
patrickzheng200, line 184:
```suggestion
		if outputFormat == ioutil.OutputPlaintext {
			fmt.Println("Trust policy is configured to skip signature verification for", reference)
			return nil
		}
		output := verifyOutput{Reference: reference, Result: "Skipped", UserMetadata: map[string]string{}}
		return ioutil.PrintObjectAsJson(output)
```

cmd/notation/verify.go
```diff
@@ -148,3 +171,31 @@ func resolveReference(ctx context.Context, opts *SecureFlagOpts, reference strin
...
 }
+
+func printResult(outputFormat, reference string, outcome *notation.VerificationOutcome) error {
+
+	if reflect.DeepEqual(outcome.VerificationLevel, trustpolicy.LevelSkip) {
+		if outputFormat == ioutil.OutputPlaintext {
+			fmt.Println("Trust policy is configured to skip signature verification for", reference)
+			return nil
+		} else {
+			output := verifyOutput{Reference: reference, Result: "Skipped", UserMetadata: map[string]string{}}
```
patrickzheng200, line 182: Looks like when verification is skipped, the two output formats are printing out different messages.
1. Plain text: `Trust policy is configured to skip signature verification for xxx`
2. JSON: `Reference: reference, Result: "Skipped", UserMetadata: map[string]string{}`
IMO, both of them should explicitly state that the trust policy configuration triggered the skip.

specs/commandline/verify.md
priteshbandi: To add some context, decision to use `--output json` came out of [this](https://hackmd.io/_vrqBGAOSUC_VWvFzWruZw?view#Meeting-Notes) meeting. Also,
* Using `--output` flag provides us flexibility to extend it to yaml or table in future. 
* [AWS](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-output-format.html), [Azure](https://learn.microsoft.com/en-us/cli/azure/format-output-azure-cli),  [Oracle](https://blogs.oracle.com/cloud-infrastructure/post/exploring-the-search-and-query-features-of-oracle-cloud-infrastructure-command-line-interface), [k8](https://kubernetes.io/docs/reference/kubectl/#formatting-output) all uses `--output` flag, which mean most users already familiar with them.

patrickzheng200: > To add some context, decision to use `--output json` came out of [this](https://hackmd.io/_vrqBGAOSUC_VWvFzWruZw?view#Meeting-Notes) meeting. Also,
> 
> * Using `--output` flag provides us flexibility to extend it to yaml or table in future.
> * [AWS](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-output-format.html), [Azure](https://learn.microsoft.com/en-us/cli/azure/format-output-azure-cli),  [Oracle](https://blogs.oracle.com/cloud-infrastructure/post/exploring-the-search-and-query-features-of-oracle-cloud-infrastructure-command-line-interface), [k8](https://kubernetes.io/docs/reference/kubectl/#formatting-output) all uses `--output` flag, which mean most users already familiar with them.

I see. Yeah, I agree, `--output` provides more extensibility here.

cmd/notation/verify.go
byronchien: since the JSON output is meant to be programmatically consumed, we should avoid full messages. would a "Result" value of "TrustPolicySkipped" (or something similar) be ok? suggestions on naming welcome

cmd/notation/verify.go
byronchien: merged from main into my branch, so these messages are now matching

byronchien: the merge resulted in a bloated diff, so undid the merge and just updated the warnings to match instead

shizhMSFT (commented):
There is no issue linked with this PR. Could you link at least one issue to this PR so that reviewers can have context?

Since this is a new feature, could you add E2E tests accordingly?

cmd/notation/verify.go
byronchien: > There is no issue linked with this PR. Could you link at least one issue to this PR so that reviewers can have context?
> 
> Since this is a new feature, could you add E2E tests accordingly?

linked the related metadata PR and issue in the initial comment. will add E2E tests for this PR

byronchien:
> There is no issue linked with this PR. Could you link at least one issue to this PR so that reviewers can have context?
> 
> Since this is a new feature, could you add E2E tests accordingly?

linked the related metadata PR and issue in the initial comment. will add E2E tests for this PR

cmd/notation/verify.go
patrickzheng200: > since the JSON output is meant to be programmatically consumed, we should avoid full messages. would a "Result" value of "TrustPolicySkipped" (or something similar) be ok? suggestions on naming welcome

IMO, `TrustPolicySkipped` sounds like the trust policy is skipped instead of the verification. How about `SkippedByTrustPolicy`?

byronchien: sounds good to me, will add

patrickzheng200 (approved):
LGTM

cmd/notation/verify.go
```diff
@@ -26,6 +26,13 @@ type verifyOpts struct {
 	reference    string
 	pluginConfig []string
 	userMetadata []string
+	outputFormat string
+}
+
+type verifyOutput struct {
+	Reference    string            `json:"reference"`
+	UserMetadata map[string]string `json:"userMetadata"`
```
priteshbandi, line 34: should we add omitempty to not not print userMetadata if there is no userMetadata?

cmd/notation/verify.go
```diff
@@ -76,7 +84,9 @@ func runVerify(command *cobra.Command, opts *verifyOpts) error {
 
 	// resolve the given reference and set the digest
 	ref, err := resolveReference(command.Context(), &opts.SecureFlagOpts, reference, sigRepo, func(ref registry.Reference, manifestDesc ocispec.Descriptor) {
-		fmt.Fprintf(os.Stderr, "Warning: Always verify the artifact using digest(@sha256:...) rather than a tag(:%s) because resolved digest may not point to the same signed artifact, as tags are mutable.\n", ref.Reference)
+		if opts.outputFormat == ioutil.OutputPlaintext {
```
priteshbandi, line 87: Do we need to do this if warnings are printed to stderr?

cmd/notation/verify.go
```diff
@@ -100,6 +110,10 @@ func runVerify(command *cobra.Command, opts *verifyOpts) error {
 		return err
 	}
 
+	if opts.outputFormat != ioutil.OutputJson && opts.outputFormat != ioutil.OutputPlaintext {
+		return fmt.Errorf("unrecognized output format: %v", opts.outputFormat)
+	}
+
```
priteshbandi, line 116: Should we do this as first thing in Verify operation like https://github.com/notaryproject/notation/blob/main/cmd/notation/sign.go#L76-L78?

cmd/notation/verify.go
byronchien: good call, will update

cmd/notation/verify.go
byronchien: sure, makes sense to fail fast to avoid wasting effort on resolving the reference etc

cmd/notation/verify.go
byronchien: i guess it's not required if customers just need to read from stdout, verified this locally w/
`./bin/notation verify $IMAGE --output json >> stdout.txt 2> stderr.txt`

specs/commandline/verify.md
```diff
@@ -168,3 +169,23 @@ An example of output messages for a successful verification:
...
+
+An example of output messages for a successful verification:
+
+```text
+{
+    "reference": "localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9",
+    "userMetadata": {
+        "io.wabbit-networks.buildId": "123"
+    },
+    "result": "Success"
```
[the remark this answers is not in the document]
byronchien: for verification failure,  no json is written to stdout, and the failure is logged to stderr.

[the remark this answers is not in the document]
shizhMSFT: @byronchien If JSON is printed only if the verification passes, what's the meaning of showing `"result": "Success"`?

/cc @priteshbandi @yizha1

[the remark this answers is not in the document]
shizhMSFT: When verification fails, I expect that the result would be `failure` or `failed`, and the failure reason will be included in the JSON object so that other scripts can parse it correctly.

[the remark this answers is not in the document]
shizhMSFT: Since this PR is merged, I've created https://github.com/notaryproject/notation/issues/546 to track.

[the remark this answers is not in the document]
priteshbandi: `result` can have two values, one is `skip` and `success`. IMO for failure case displaying JSON for wont be useful because it wont contain any useful information which automation/script can use(non-zero exit code should suffice to show failure)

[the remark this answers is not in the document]
shizhMSFT: The JSON output is usually consumed by scripts or programs. How can them obtain the structured error message and detailed verification outcomes?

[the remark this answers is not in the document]
priteshbandi: To display error message/detailed verification output we need more than `result` field something like `resultReason` because `result` will only say `failure`. Also, in current state it wont be of much help since as there is only one genuine/expected failure use case i.e failed signature verification for all the signatures. Apart from genuine/expected error, all other errors should always be emitted as stdErr.

Are you suggest we should emit json for expected failure use cases or all failure use cases?

PR #528, opened 2023-02-02
feat: add implementation for `notation inspect`

### LINKED ISSUE #219: Implementing inspect command
Summary

Implementing `notation inspect` command according to the spec #213 

Intended Outcome
`notation inspect` command works as defined in the spec

Additional context

byronchien:
Adds support for `notation inspect` ([spec here](https://github.com/notaryproject/notation/blob/main/specs/commandline/inspect.md))

Example output:
```
chienb@a07817b52895 notation % ./bin/notation inspect $IMAGE
Resolved artifact tag `v1` to digest `sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b` before inspect.
Inspecting all signatures for signed artifact
localhost:5000/net-monitor@sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b
    └── sha256:34e5843a1a8b1607d2ba7da9b61c6a5b3953f5680751c160fb944df87b01b2b2
        ├── signature algorithm : RSASSA-PSS-SHA-256
        ├── signed attributes
        │   ├── expiry : 0001-01-01 00:00:00 +0000 UTC
        │   ├── signingScheme : notary.x509
        │   └── signingTime : 2023-01-27 17:02:22 -0800 PST
        ├── user defined attributes
        │   └── io.wabbit-networks.buildId : 123
        ├── unsigned attributes
        │   └── signingAgent : Notation/1.0.0
        ├── certificates
        │   └── SHA1 fingerprint e1ef7b0f984d1f8222d6bf297e1ad10047997b54
        │       ├── issued to : CN=byron.test,O=Notary,L=Seattle,ST=WA,C=US
        │       ├── issued by : CN=byron.test,O=Notary,L=Seattle,ST=WA,C=US
        │       └── expiry : 2023-01-29 01:02:13 +0000 UTC
        └── signed artifact
            ├── media type : application/vnd.docker.distribution.manifest.v2+json
            ├── digest : sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b
            └── size : 942
chienb@a07817b52895 notation % ./bin/notation inspect $IMAGE --output json
{
    "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
    "Signatures": [
        {
            "digest": "sha256:34e5843a1a8b1607d2ba7da9b61c6a5b3953f5680751c160fb944df87b01b2b2",
            "signatureAlgorithm": "RSASSA-PSS-SHA-256",
            "signedAttributes": {
                "expiry": "0001-01-01 00:00:00 +0000 UTC",
                "signingScheme": "notary.x509",
                "signingTime": "2023-01-27 17:02:22 -0800 PST"
            },
            "userDefinedAttributes": {
                "io.wabbit-networks.buildId": "123"
            },
            "unsignedAttributes": {
                "signingAgent": "Notation/1.0.0"
            },
            "certificates": [
                {
                    "SHA1Fingerprint": "e1ef7b0f984d1f8222d6bf297e1ad10047997b54",
                    "issuedTo": "CN=byron.test,O=Notary,L=Seattle,ST=WA,C=US",
                    "issuedBy": "CN=byron.test,O=Notary,L=Seattle,ST=WA,C=US",
                    "expiry": "2023-01-29 01:02:13 +0000 UTC"
                }
            ],
            "signedArtifact": {
                "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
                "digest": "sha256:5a07385af4e6b6af81b0ebfd435aedccdfa3507f0609c658209e1aba57159b2b",
                "size": 942
            }
        }
    ]
}
```

NOTE: this PR is on top of the changes for user-metadata verifictation and json output from there PRs:
- https://github.com/notaryproject/notation/pull/527
- https://github.com/notaryproject/notation/pull/507

suggested order for review: [notation-go #261](https://github.com/notaryproject/notation-go/pull/261) => [notation-core-go #111](https://github.com/notaryproject/notation-core-go/pull/111) => [notation #527](https://github.com/notaryproject/notation/pull/527) =>[notation #528](https://github.com/notaryproject/notation/pull/528) (this one)

Signed-off-by: Byron Chien <chienb@amazon.com>

### DIFF
--- a/cmd/notation/inspect.go
+++ b/cmd/notation/inspect.go
@@ -1,0 +1,302 @@
+package main
+
+import (
+	"crypto/sha1"
+	b64 "encoding/base64"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+	"strconv"
+	"time"
+
+	"github.com/notaryproject/notation-core-go/signature"
+	"github.com/notaryproject/notation-go/plugin/proto"
+	"github.com/notaryproject/notation-go/registry"
+	"github.com/notaryproject/notation/internal/cmd"
+	"github.com/notaryproject/notation/internal/envelope"
+	"github.com/notaryproject/notation/internal/ioutil"
+	"github.com/notaryproject/notation/internal/tree"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/spf13/cobra"
+)
+
+type inspectOpts struct {
+	cmd.LoggingFlagOpts
+	SecureFlagOpts
+	reference    string
+	outputFormat string
+}
+
+type inspectOutput struct {
+	MediaType  string `json:"mediaType"`
+	Signatures []signatureOutput
+}
+
+type signatureOutput struct {
+	MediaType             string              `json:"mediaType"`
+	Digest                string              `json:"digest"`
+	SignatureAlgorithm    string              `json:"signatureAlgorithm"`
+	SignedAttributes      map[string]string   `json:"signedAttributes"`
+	UserDefinedAttributes map[string]string   `json:"userDefinedAttributes"`
+	UnsignedAttributes    map[string]string   `json:"unsignedAttributes"`
+	Certificates          []certificateOutput `json:"certificates"`
+	SignedArtifact        ocispec.Descriptor  `json:"signedArtifact"`
+}
+
+type certificateOutput struct {
+	SHA1Fingerprint string `json:"SHA1Fingerprint"`
+	IssuedTo        string `json:"issuedTo"`
+	IssuedBy        string `json:"issuedBy"`
+	Expiry          string `json:"expiry"`
+}
+
+func inspectCommand(opts *inspectOpts) *cobra.Command {
+	if opts == nil {
+		opts = &inspectOpts{}
+	}
+	command := &cobra.Command{
+		Use:   "inspect [reference]",
+		Short: "Inspect all signatures associated with the signed artifact",
+		Long: `Inspect all signatures associated with the signed artifact.
+
+Example - Inspect signatures on an OCI artifact identified by a digest:
+  notation inspect <registry>/<repository>@<digest>
+
+Example - Inspect signatures on an OCI artifact identified by a tag  (Notation will resolve tag to digest):
+  notation inspect <registry>/<repository>:<tag>
+
+Example - Inspect signatures on an OCI artifact identified by a digest and output as json:
+  notation inspect --output json <registry>/<repository>@<digest>
+`,
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) == 0 {
+				return errors.New("missing reference")
+			}
+			opts.reference = args[0]
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runInspect(cmd, opts)
+		},
+	}
+
+	opts.LoggingFlagOpts.ApplyFlags(command.Flags())
+	opts.SecureFlagOpts.ApplyFlags(command.Flags())
+	cmd.SetPflagOutput(command.Flags(), &opts.outputFormat, cmd.PflagOutputUsage)
+	return command
+}
+
+func runInspect(command *cobra.Command, opts *inspectOpts) error {
+	// set log level
+	ctx := opts.LoggingFlagOpts.SetLoggerLevel(command.Context())
+
+	if opts.outputFormat != cmd.OutputJSON && opts.outputFormat != cmd.OutputPlaintext {
+		return fmt.Errorf("unrecognized output format %s", opts.outputFormat)
+	}
+
+	// initialize
+	reference := opts.reference
+	sigRepo, err := getSignatureRepository(ctx, &opts.SecureFlagOpts, reference)
+	if err != nil {
+		return err
+	}
+
+	manifestDesc, ref, err := getManifestDescriptor(ctx, &opts.SecureFlagOpts, reference, sigRepo)
+	if err != nil {
+		return err
+	}
+
+	// reference is a digest reference
+	if err := ref.ValidateReferenceAsDigest(); err != nil {
+		fmt.Fprintf(os.Stderr, "Warning: Always inspect the artifact using digest(@sha256:...) rather than a tag(:%s) because resolved digest may not point to the same signed artifact, as tags are mutable.\n", ref.Reference)
+		ref.Reference = manifestDesc.Digest.String()
+	}
+
+	output := inspectOutput{MediaType: manifestDesc.MediaType, Signatures: []signatureOutput{}}
+	skippedSignatures := false
+	err = sigRepo.ListSignatures(ctx, manifestDesc, func(signatureManifests []ocispec.Descriptor) error {
+		for _, sigManifestDesc := range signatureManifests {
+			sigBlob, sigDesc, err := sigRepo.FetchSignatureBlob(ctx, sigManifestDesc)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "Warning: unable to fetch signature %s due to error: %v\n", sigManifestDesc.Digest.String(), err)
+				skippedSignatures = true
+				continue
+			}
+
+			sigEnvelope, err := signature.ParseEnvelope(sigDesc.MediaType, sigBlob)
+			if err != nil {
+				logSkippedSignature(sigManifestDesc, err)
+				skippedSignatures = true
+				continue
+			}
+
+			envelopeContent, err := sigEnvelope.Content()
+			if err != nil {
+				logSkippedSignature(sigManifestDesc, err)
+				skippedSignatures = true
+				continue
+			}
+
+			signedArtifactDesc, err := envelope.DescriptorFromSignaturePayload(&envelopeContent.Payload)
+			if err != nil {
+				logSkippedSignature(sigManifestDesc, err)
+				skippedSignatures = true
+				continue
+			}
+
+			signatureAlgorithm, err := proto.EncodeSigningAlgorithm(envelopeContent.SignerInfo.SignatureAlgorithm)
+			if err != nil {
+				logSkippedSignature(sigManifestDesc, err)
+				skippedSignatures = true
+				continue
+			}
+
+			sig := signatureOutput{
+				MediaType:             sigDesc.MediaType,
+				Digest:                sigManifestDesc.Digest.String(),
+				SignatureAlgorithm:    string(signatureAlgorithm),
+				SignedAttributes:      getSignedAttributes(opts.outputFormat, envelopeContent),
+				UserDefinedAttributes: signedArtifactDesc.Annotations,
+				UnsignedAttributes:    getUnsignedAttributes(envelopeContent),
+				Certificates:          getCertificates(opts.outputFormat, envelopeContent),
+				SignedArtifact:        *signedArtifactDesc,
+			}
+
+			// clearing annotations from the SignedArtifact field since they're already
+			// displayed as UserDefinedAttributes
+			sig.SignedArtifact.Annotations = nil
+
+			output.Signatures = append(output.Signatures, sig)
+		}
+		return nil
+	})
+
+	if err != nil {
+		return err
+	}
+
+	err = printOutput(opts.outputFormat, ref.String(), output)
+	if err != nil {
+		return err
+	}
+
+	if skippedSignatures {
+		return errors.New("at least one signature was skipped and not displayed")
+	}
+
+	return nil
+}
+
+func logSkippedSignature(sigDesc ocispec.Descriptor, err error) {
+	fmt.Fprintf(os.Stderr, "Warning: Skipping signature %s because of error: %v\n", sigDesc.Digest.String(), err)
+}
+
+func getSignedAttributes(outputFormat string, envContent *signature.EnvelopeContent) map[string]string {
+	signedAttributes := map[string]string{
+		"signingScheme": string(envContent.SignerInfo.SignedAttributes.SigningScheme),
+		"signingTime":   formatTimestamp(outputFormat, envContent.SignerInfo.SignedAttributes.SigningTime),
+		"expiry":        formatTimestamp(outputFormat, envContent.SignerInfo.SignedAttributes.Expiry),
+	}
+
+	for _, attribute := range envContent.SignerInfo.SignedAttributes.ExtendedAttributes {
+		signedAttributes[fmt.Sprint(attribute.Key)] = fmt.Sprint(attribute.Value)
+	}
+
+	return signedAttributes
+}
+
+func getUnsignedAttributes(envContent *signature.EnvelopeContent) map[string]string {
+	unsignedAttributes := map[string]string{}
+
+	if envContent.SignerInfo.UnsignedAttributes.TimestampSignature != nil {
+		unsignedAttributes["timestampSignature"] = b64.StdEncoding.EncodeToString(envContent.SignerInfo.UnsignedAttributes.TimestampSignature)
+	}
+
+	if envContent.SignerInfo.UnsignedAttributes.SigningAgent != "" {
+		unsignedAttributes["signingAgent"] = envContent.SignerInfo.UnsignedAttributes.SigningAgent
+	}
+
+	return unsignedAttributes
+}
+
+func formatTimestamp(outputFormat string, t time.Time) string {
+	switch outputFormat {
+	case cmd.OutputJSON:
+		return t.Format(time.RFC3339)
+	default:
+		return t.Format(time.ANSIC)
+	}
+}
+
+func getCertificates(outputFormat string, envContent *signature.EnvelopeContent) []certificateOutput {
+	certificates := []certificateOutput{}
+
+	for _, cert := range envContent.SignerInfo.CertificateChain {
+		h := sha1.Sum(cert.Raw)
+		fingerprint := strings.ToLower(hex.EncodeToString(h[:]))
+
+		certificate := certificateOutput{
+			SHA1Fingerprint: fingerprint,
+			IssuedTo:        cert.Subject.String(),
+			IssuedBy:        cert.Issuer.String(),
+			Expiry:          formatTimestamp(outputFormat, cert.NotAfter),
+		}
+
+		certificates = append(certificates, certificate)
+	}
+
+	return certificates
+}
+
+func printOutput(outputFormat string, ref string, output inspectOutput) error {
+	if outputFormat == cmd.OutputJSON {
+		return ioutil.PrintObjectAsJSON(output)
+	}
+
+	fmt.Println("Inspecting all signatures for signed artifact")
+	root := tree.New(ref)
+	cncfSigNode := root.Add(registry.ArtifactTypeNotation)
+
+	for _, signature := range output.Signatures {
+		sigNode := cncfSigNode.Add(signature.Digest)
+		sigNode.AddPair("media type", signature.MediaType)
+		sigNode.AddPair("signature algorithm", signature.SignatureAlgorithm)
+
+		signedAttributesNode := sigNode.Add("signed attributes")
+		addMapToTree(signedAttributesNode, signature.SignedAttributes)
+
+		userDefinedAttributesNode := sigNode.Add("user defined attributes")
+		addMapToTree(userDefinedAttributesNode, signature.UserDefinedAttributes)
+
+		unsignedAttributesNode := sigNode.Add("unsigned attributes")
+		addMapToTree(unsignedAttributesNode, signature.UnsignedAttributes)
+
+		certListNode := sigNode.Add("certificates")
+		for _, cert := range signature.Certificates {
+			certNode := certListNode.AddPair("SHA1 fingerprint", cert.SHA1Fingerprint)
+			certNode.AddPair("issued to", cert.IssuedTo)
+			certNode.AddPair("issued by", cert.IssuedBy)
+			certNode.AddPair("expiry", cert.Expiry)
+		}
+
+		artifactNode := sigNode.Add("signed artifact")
+		artifactNode.AddPair("media type", signature.SignedArtifact.MediaType)
+		artifactNode.AddPair("digest", signature.SignedArtifact.Digest.String())
+		artifactNode.AddPair("size", strconv.FormatInt(signature.SignedArtifact.Size, 10))
+	}
+
+	root.Print()
+	return nil
+}
+
+func addMapToTree(node *tree.Node, m map[string]string) {
+	if len(m) > 0 {
+		for k, v := range m {
+			node.AddPair(k, v)
+		}
+	} else {
+		node.Add("(empty)")
+	}
+}
--- a/internal/envelope/envelope.go
+++ b/internal/envelope/envelope.go
@@ -1,17 +1,29 @@
 package envelope
 
 import (
+	"encoding/json"
+	"errors"
 	"fmt"
 
+	"github.com/notaryproject/notation-core-go/signature"
 	"github.com/notaryproject/notation-core-go/signature/cose"
 	"github.com/notaryproject/notation-core-go/signature/jws"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
-// Supported envelope format.
 const (
+	// Supported envelope format.
 	COSE = "cose"
 	JWS  = "jws"
+
+	// MediaTypePayloadV1 is the supported content type for signature's payload.
+	MediaTypePayloadV1 = "application/vnd.cncf.notary.payload.v1+json"
 )
+
+// Payload describes the content that gets signed.
+type Payload struct {
+	TargetArtifact ocispec.Descriptor `json:"targetArtifact"`
+}
 
 // GetEnvelopeMediaType converts the envelope type to mediaType name.
 func GetEnvelopeMediaType(sigFormat string) (string, error) {
@@ -22,4 +34,35 @@ func GetEnvelopeMediaType(sigFormat string) (string, error) {
 		return cose.MediaTypeEnvelope, nil
 	}
 	return "", fmt.Errorf("signature format %q not supported", sigFormat)
+}
+
+// ValidatePayloadContentType validates signature payload's content type.
+func ValidatePayloadContentType(payload *signature.Payload) error {
+	switch payload.ContentType {
+	case MediaTypePayloadV1:
+		return nil
+	default:
+		return fmt.Errorf("payload content type %q not supported", payload.ContentType)
+	}
+}
+
+// DescriptorFromPayload parses a signature payload and returns the descriptor
+// that was signed. Note: the descriptor was signed but may not be trusted
+func DescriptorFromSignaturePayload(payload *signature.Payload) (*ocispec.Descriptor, error) {
+	if payload == nil {
+		return nil, errors.New("empty payload")
+	}
+
+	err := ValidatePayloadContentType(payload)
+	if err != nil {
+		return nil, err
+	}
+
+	var parsedPayload Payload
+	err = json.Unmarshal(payload.Content, &parsedPayload)
+	if err != nil {
+		return nil, errors.New("failed to unmarshall the payload content to Payload")
+	}
+
+	return &parsedPayload.TargetArtifact, nil
 }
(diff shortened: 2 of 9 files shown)


byronchien:
order to review:
https://github.com/notaryproject/notation/pull/507
https://github.com/notaryproject/notation/pull/527
https://github.com/notaryproject/notation/pull/528

byronchien:
depends on
https://github.com/notaryproject/notation-go/pull/261
https://github.com/notaryproject/notation-core-go/pull/111

cmd/notation/inspect.go
```diff
@@ -0,0 +1,288 @@
...
+	manifestDesc, ref, err := getManifestDescriptor(ctx, &opts.SecureFlagOpts, reference, sigRepo)
+	if err != nil {
+		return err
+	}
+
+	// reference is a digest reference
+	if err := ref.ValidateReferenceAsDigest(); err != nil {
+		if opts.outputFormat == ioutil.OutputPlaintext {
+			fmt.Printf("Resolved artifact tag `%s` to digest `%s` before inspect.\n", ref.Reference, manifestDesc.Digest.String())
+			fmt.Println("Warning: The resolved digest may not point to the same signed artifact, since tags are mutable.")
```
patrickzheng200, line 108: I'm trying to understand this printout, since `ref` is a digest reference (validated by `ref.ValidateReferenceAsDigest()`), we are not resolving any tag here right?

cmd/notation/inspect.go
```diff
@@ -0,0 +1,288 @@
...
+	return unsignedAttributes
+}
+
+func getCertificates(envContent *signature.EnvelopeContent) []certificateOutput {
+	certificates := []certificateOutput{}
+
+	for _, cert := range envContent.SignerInfo.CertificateChain {
+		h := sha1.New()
+		h.Write(cert.RawTBSCertificate)
+		fingerprint := fmt.Sprintf("%x", h.Sum(nil))
```
patrickzheng200, line 202: Should we use `cert.Raw` to compute SHA1 Thumbprint?
```suggestion
		h := sha1.Sum(cert.Raw)
		fingerprint := strings.ToLower(hex.EncodeToString(h[:]))
```

cmd/notation/inspect.go
```diff
@@ -0,0 +1,288 @@
...
+				Digest:                sigDesc.Digest.String(),
+				SignatureAlgorithm:    envelopeContent.SignerInfo.SignatureAlgorithm.String(),
+				SignedAttributes:      getSignedAttributes(envelopeContent),
+				UserDefinedAttributes: signedArtifactDesc.Annotations,
+				UnsignedAttributes:    getUnsignedAttributes(envelopeContent),
+				Certificates:          getCertificates(envelopeContent),
+				SignedArtifact: artifact{
+					MediaType: signedArtifactDesc.MediaType,
+					Digest:    signedArtifactDesc.Digest.String(),
+					Size:      signedArtifactDesc.Size,
```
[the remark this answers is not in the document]
byronchien: since the descriptor contains annotations, can't users add different user-metadata per signature?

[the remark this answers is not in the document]
byronchien: ah but the user-metadata would appear under user-defined attributes instead, so the the signed artifact here would be the same. in that case, yes, printing it multiple times seems redundant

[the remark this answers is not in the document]
byronchien: discussed separately, in the scenario that a signature for image A is attached to image B in a remote repository, it's possible for the signed artifact to be different across signatures despite being listed for the same reference. in that case, it makes sense to have the artifact displayed for all signatures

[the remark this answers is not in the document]
patrickzheng200: > discussed separately, in the scenario that a signature for image A is attached to image B in a remote repository, it's possible for the signed artifact to be different across signatures despite being listed for the same reference. in that case, it makes sense to have the artifact displayed for all signatures

This is a very interesting point. Just curious, do we have any use case of attaching a signature of image A to image B? Because I think this signature would fail the verification process on verifying image B. So for the `inspect` command, do we want to show signatures that truly signed image B only?

cmd/notation/inspect.go
```diff
@@ -0,0 +1,288 @@
...
+				return fmt.Errorf("unable to retrieve digital signature with digest %q associated with %q from the registry, error : %v", sigManifestDesc.Digest, manifestDesc.Digest, err.Error())
+			}
+
+			envelope, err := signature.ParseEnvelope(sigDesc.MediaType, sigBlob)
+			envelopeContent, err := envelope.Content()
+			if err != nil {
+				return err
+			}
+
+			signedArtifactDesc, err := notation.GetDescriptorFromPayload(&envelopeContent.Payload)
```
patrickzheng200, line 133: As discussed in [this PR](https://github.com/notaryproject/notation-go/pull/261#issuecomment-1419846276), we should have an implement of this method in notation CLI.

cmd/notation/inspect.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+	// reference is a digest reference
+	if err := ref.ValidateReferenceAsDigest(); err != nil {
+		if opts.outputFormat == ioutil.OutputPlaintext {
+			fmt.Printf("Resolved artifact tag `%s` to digest `%s` before inspect.\n", ref.Reference, manifestDesc.Digest.String())
+			fmt.Println("Warning: The resolved digest may not point to the same signed artifact, since tags are mutable.")
+		}
+		ref.Reference = manifestDesc.Digest.String()
+	}
```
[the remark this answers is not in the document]
byronchien: the reason for not reusing the resolveReference logic defined there was to get the manifestDescriptor (which isn't returned from resolveReference)

cmd/notation/inspect.go
byronchien:
```
	// reference is a digest reference
	if err := ref.ValidateReferenceAsDigest(); err != nil {
		if opts.outputFormat == ioutil.OutputPlaintext {
			fmt.Printf("Resolved artifact tag `%s` to digest `%s` before inspect.\n", ref.Reference, manifestDesc.Digest.String())
			fmt.Println("Warning: The resolved digest may not point to the same signed artifact, since tags are mutable.")
		}
		ref.Reference = manifestDesc.Digest.String()
	}
```

the if condition is different from the one in resolveReference. the warning is printed if the err from ref.ValidateReferenceAsDigest() != nil (so the reference is a tag)

cmd/notation/inspect.go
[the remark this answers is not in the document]
byronchien: my impression was that we're just showing all signatures that are present in the sig repo for the reference, whether the payload truly matches or not. i am not sure about a specific use case around attaching signatures for different images, so i'd defer that decision to someone more familiar with the original motivation in the spec

cmd/notation/inspect.go
byronchien: added (renamed to DescriptorFromSignaturePayload)

internal/envelope/envelope.go
```diff
@@ -23,3 +43,37 @@ func GetEnvelopeMediaType(sigFormat string) (string, error) {
...
+	for _, mediaType := range supportedMediaTypes {
+		if intPayload.TargetArtifact.MediaType == mediaType {
+			validMediaType = true
+			break
+		}
+	}
+
+	if !validMediaType {
+		return nil, errors.New(fmt.Sprintf("unsupported media type %s", intPayload.TargetArtifact.MediaType))
+	}
```
patrickzheng200, line 70: Looks like you are trying to validate the `TargetArtifact.MediaType`. Should we instead validate the `payload.ContentType`? To validate the content type: https://github.com/notaryproject/notation-go/blob/1b2ec27aea7b2565b9adfb8d5d32e6df88ef4325/internal/envelope/envelope.go#L19. 
(Since it's unexported in notation-go, you could add it in internal/envelope/envlope.go at the CLI side as well.)

byronchien: the original comment that this was added for was specifically about validating the Payload MediaType ([link](https://github.com/notaryproject/notation-go/pull/261#discussion_r1098062259)), but i may have also misunderstood the ask there. can clarify whether we still want to validate the media type of the target artifact, but otherwise will add the helper for validating the payload content type

patrickzheng200: > the original comment that this was added for was specifically about validating the Payload MediaType ([link](https://github.com/notaryproject/notation-go/pull/261#discussion_r1098062259)), but i may have also misunderstood the ask there. can clarify whether we still want to validate the media type of the target artifact, but otherwise will add the helper for validating the payload content type

We should always validate the `payload.ContentType` based on our [spec](https://github.com/notaryproject/notaryproject/blob/main/specs/signature-specification.md#payload). (Note in the example of the spec, the `payload.TargetArtifact` MediaType can be `application/vnd.oci.image.manifest.v1+json` or `sbom/example`)

byronchien: confirmed with shiwei that the comment was intended to be for ContentType, will remove the media type check

internal/envelope/envelope.go
```diff
@@ -1,18 +1,38 @@
...
+	TargetArtifact ocispec.Descriptor `json:"targetArtifact"`
+}
+
+type intermediatePayload struct {
+	TargetArtifact intermediateTargetArtifact `json:"targetArtifact"`
+}
+
+type intermediateTargetArtifact struct {
+	MediaType string `json:"mediaType"`
+}
```
patrickzheng200, line 34: Looks like we are not using these structs anymore? Could remove them.

cmd/notation/inspect.go
```diff
@@ -0,0 +1,312 @@
...
+
+	manifestDesc, ref, err := getManifestDescriptor(ctx, &opts.SecureFlagOpts, reference, sigRepo)
+	if err != nil {
+		return err
+	}
+
+	// reference is a digest reference
+	if err := ref.ValidateReferenceAsDigest(); err != nil {
+		fmt.Fprintf(os.Stderr, "Warning: Always verify the artifact using digest(@sha256:...) rather than a tag(:%s) because resolved digest may not point to the same signed artifact, as tags are mutable.\n", ref.Reference)
+		ref.Reference = manifestDesc.Digest.String()
```
priteshbandi, line 122:
```suggestion
		fmt.Fprintf(os.Stderr, "Warning: Always inspect the artifact using digest(@sha256:...) rather than a tag(:%s) because resolved digest may not point to the same signed artifact, as tags are mutable.\n", ref.Reference)
		ref.Reference = manifestDesc.Digest.String()
```

cmd/notation/inspect.go
```diff
@@ -0,0 +1,312 @@
...
+			sigBlob, sigDesc, err := sigRepo.FetchSignatureBlob(ctx, sigManifestDesc)
+			if err != nil {
+				return fmt.Errorf("unable to retrieve digital signature with digest %q associated with %q from the registry, error : %v", sigManifestDesc.Digest, manifestDesc.Digest, err)
+			}
+
+			sigEnvelope, err := signature.ParseEnvelope(sigDesc.MediaType, sigBlob)
+			envelopeContent, err := sigEnvelope.Content()
+			if err != nil {
+				return err
+			}
```
priteshbandi, line 143: We should continue even when we are not able to parse signature.

cmd/notation/inspect.go
```diff
@@ -0,0 +1,312 @@
...
+
+			err = envelope.ValidatePayloadContentType(&envelopeContent.Payload)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "Warning: Skipping signature %s : %s", sigDesc.Digest.String(), err)
```
priteshbandi, line 147: Should we return non-zero exit code for this usecase?

priteshbandi, line 147:
```suggestion
				fmt.Fprintf(os.Stderr, "Warning: Skipping signature %s because of error: %v", sigDesc.Digest.String(), err)
```

cmd/notation/inspect.go
```diff
@@ -0,0 +1,312 @@
...
+				continue
```
patrickzheng200, line 148: If the content type is not supported, we choose to skip this signature and continue the `inspect` process. I notice that we won't `continue` on other failures in this function, for example, `DescriptorFromSignaturePayload` and `EncodeSigningAlgorithm`. Is this by design?

cmd/notation/inspect.go
```diff
@@ -0,0 +1,312 @@
...
+
+			signedArtifactDesc, err := envelope.DescriptorFromSignaturePayload(&envelopeContent.Payload)
+			if err != nil {
+				return err
+			}
+
+			signatureAlgorithm, err := proto.EncodeSigningAlgorithm(envelopeContent.SignerInfo.SignatureAlgorithm)
+			if err != nil {
+				return err
+			}
```
priteshbandi, line 159: ditto, we should continue for any error and display warning.

internal/envelope/envelope.go
priteshbandi, line 34: Can we remove them?

internal/envelope/envelope.go
```diff
@@ -23,3 +43,29 @@ func GetEnvelopeMediaType(sigFormat string) (string, error) {
...
+	}
+
+	var parsedPayload Payload
+	err := json.Unmarshal(payload.Content, &parsedPayload)
+	if err != nil {
+		return nil, errors.New("failed to unmarshall the payload content to Payload")
+	}
+
+	return &parsedPayload.TargetArtifact, nil
+}
```
patrickzheng200, line 71: IMO, from code reusability perspective, we could call `ValidatePayloadContentType` inside `DescriptorFromSignaturePayload`. Otherwise, using `DescriptorFromSignaturePayload` would require calling `ValidatePayloadContentType` beforehand.

cmd/notation/inspect.go
```diff
@@ -0,0 +1,315 @@
...
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/spf13/cobra"
+)
+
+const (
+	treeItemPrefix     = "├──"
+	treeItemPrefixLast = "└──"
+	subTreePrefix      = "│  "
+	subTreePrefixLast  = "   "
+)
```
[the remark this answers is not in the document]
priteshbandi: opened issue to track this https://github.com/notaryproject/notation/issues/545

cmd/notation/inspect.go
byronchien: discussed w/ pritesh, will be updating to continue and log for all errors

priteshbandi (dismissed):
LGTM

cmd/notation/inspect.go
```diff
@@ -0,0 +1,327 @@
...
+
+	return certificates
+}
+
+func printOutput(outputFormat string, output inspectOutput) error {
+	if outputFormat == cmd.OutputJson {
+		return ioutil.PrintObjectAsJSON(output)
+	}
+
+	fmt.Println("└── application/vnd.cncf.notary.signature")
```
shizhMSFT, line 260: Do we have a const for `application/vnd.cncf.notary.signature`?

cmd/notation/inspect.go
```diff
@@ -0,0 +1,327 @@
...
+		}
+
+		certificates = append(certificates, certificate)
+	}
+
+	return certificates
+}
+
+func printOutput(outputFormat string, output inspectOutput) error {
+	if outputFormat == cmd.OutputJson {
```
shizhMSFT, line 256: In golang, `Json` should be named as `JSON`. Therefore, the flag `OutputJson` should be named as `OutputJSON`.

cmd/notation/inspect.go
```diff
@@ -0,0 +1,327 @@
...
+			sigTree = treeItemPrefixLast
+			sigTreePrefix = subTreePrefixLast
+		}
+
+		fmt.Printf("    %s %s\n", sigTree, signature.Digest)
+
+		sigSubTree := treeItemPrefix
+		sigSubTreePrefix := subTreePrefix
+
+		fmt.Printf("    %s %s %s : %s\n", sigTreePrefix, sigSubTree, "media type", signature.MediaType)
```
shizhMSFT, line 275: According to the spec, there is no space before the colon.
```suggestion
		fmt.Printf("    %s %s %s: %s\n", sigTreePrefix, sigSubTree, "media type", signature.MediaType)
```
Same applies to other lines.

cmd/notation/inspect.go
```diff
@@ -0,0 +1,327 @@
...
+
+	for _, cert := range envContent.SignerInfo.CertificateChain {
+		h := sha1.Sum(cert.Raw)
+		fingerprint := strings.ToLower(hex.EncodeToString(h[:]))
+
+		certificate := certificateOutput{
+			SHA1Fingerprint: fingerprint,
+			IssuedTo:        cert.Subject.String(),
+			IssuedBy:        cert.Issuer.String(),
+			Expiry:          cert.NotAfter.String(),
```
shizhMSFT, line 246: According to spec, those datetime should be in the RFC3339 form for the JSON output

cmd/notation/inspect.go
```diff
@@ -0,0 +1,327 @@
...
+func logSkippedSignature(sigDesc ocispec.Descriptor, err error) {
+	fmt.Fprintf(os.Stderr, "Warning: Skipping signature %s because of error: %v\n", sigDesc.Digest.String(), err)
+}
+
+func getSignedAttributes(envContent *signature.EnvelopeContent) map[string]string {
+	signedAttributes := map[string]string{}
+
+	signedAttributes["signingScheme"] = string(envContent.SignerInfo.SignedAttributes.SigningScheme)
+	signedAttributes["signingTime"] = envContent.SignerInfo.SignedAttributes.SigningTime.String()
+	signedAttributes["expiry"] = envContent.SignerInfo.SignedAttributes.Expiry.String()
```
shizhMSFT, line 208: According to spec, those datetime should be in the RFC3339 form for the JSON output

shizhMSFT: Please note that the current format does not match the format for the tree view as in the spec.

cmd/notation/inspect.go
shizhMSFT: Please note that the current format does not match the format for the tree view as in the spec.

cmd/notation/inspect.go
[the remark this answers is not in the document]
shizhMSFT: If you don't consider printing tree while loading signatures pages, you may consider https://github.com/need-being/go-tree as it is MIT.

cmd/notation/inspect.go
```diff
@@ -0,0 +1,327 @@
...
+	})
+
+	if err != nil {
+		return err
+	}
+
+	err = printOutput(opts.outputFormat, output)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
shizhMSFT: Note: It is a nice to have feature, which can come after this PR.

test/e2e/suite/scenario/quickstart.go
```diff
@@ -74,7 +74,7 @@ var _ = Describe("notation quickstart E2E test", Ordered, func() {
 
 	It("Verify the container image with jws format", func() {
 		notation.Exec("verify", artifact.ReferenceWithDigest()).
-			MatchContent(fmt.Sprintf("Successfully verified signature for %s\n", artifact.ReferenceWithDigest()))
+			MatchKeyWords(fmt.Sprintf("Successfully verified signature for %s\n", artifact.ReferenceWithDigest()))
```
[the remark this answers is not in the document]
byronchien: when running the E2E tests, sometimes the tests with user-metadata would run before this one, causing the output to mismatch since the signature with user-metadata would be verified first and cause the output to not match exactly. Changing it so that the test just checks that the verification passes successfully regardless of whether tags are present or not.

cmd/notation/inspect.go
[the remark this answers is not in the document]
priteshbandi: IMO instead of adding dependency on not a well known library such as `need-being/go-tree` is risky and since the amount of code we are taking about is less than 100 lines I would prefer that we we write our own.

cmd/notation/inspect.go
```diff
@@ -0,0 +1,294 @@
...
+
+	for _, signature := range output.Signatures {
+		root.AddPath(NotarySignatureMediaType, signature.Digest, fmt.Sprintf("media type: %s", signature.MediaType))
+		root.AddPath(NotarySignatureMediaType, signature.Digest, fmt.Sprintf("signature algorithm: %s", signature.SignatureAlgorithm))
+
+		for k, v := range signature.SignedAttributes {
+			root.AddPath(NotarySignatureMediaType, signature.Digest, "signed attributes", fmt.Sprintf("%s: %s", k, v))
+		}
+
+		for k, v := range signature.UserDefinedAttributes {
```
patrickzheng200, line 272: Looks like when the signature does not have any user defined attribute, the phrase `user defined attributes` would still be printed out. Shall we just skip the phrase in this case?
```
  │   │   ├── expiry: 0001-01-01 00:00:00 +0000 UTC
    │   │   └── signingScheme: notary.x509
    │   ├── user defined attributes
    │   ├── unsigned attributes
    │   │   └── signingAgent: Notation/1.0.0
```

cmd/notation/inspect.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+func printOutput(outputFormat string, ref string, output inspectOutput) error {
+	if outputFormat == cmd.OutputJSON {
+		return ioutil.PrintObjectAsJSON(output)
+	}
+
+	fmt.Println("Inspecting all signatures for signed artifact")
+	root := tree.New(ref)
+	root.Add(NotarySignatureMediaType)
```
shizhMSFT, line 262: The [Add()](https://pkg.go.dev/github.com/need-being/go-tree#Node.Add) does return a node.
```suggestion
	mediaTypeNode := root.Add(NotarySignatureMediaType)
```
Later, you can use `mediaTypeNode` like
```go
digestNode := mediaTypeNode.Add(signature.Digest)
digestNode.Add(fmt.Sprintf("media type: %s", signature.MediaType))
digestNode.Add(fmt.Sprintf("signature algorithm: %s", signature.SignatureAlgorithm))
```
instead of
```go
root.AddPath(NotarySignatureMediaType, signature.Digest, fmt.Sprintf("media type: %s", signature.MediaType))
root.AddPath(NotarySignatureMediaType, signature.Digest, fmt.Sprintf("signature algorithm: %s", signature.SignatureAlgorithm))
```

cmd/notation/inspect.go
shizhMSFT: We can keep as it is or just show `(empty)`.

cmd/notation/inspect.go
byronchien: oh that makes things much nicer, thank you

cmd/notation/inspect.go
byronchien: will change to `(empty)`, i personally prefer having some indicator that the field is present without any value

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.20
 
 require (
 	github.com/docker/docker-credential-helpers v0.7.0
+	github.com/need-being/go-tree v0.1.0
```
priteshbandi, line 7: Adding not a well known library such as need-being/go-tree is risky and since the amount of code we are taking about is less than 100 lines I would prefer that we we write our own.

internal/ioutil/tree.go
```diff
@@ -0,0 +1,58 @@
...
+	subTreePrefix      = "│   "
+	subTreePrefixLast  = "    "
+)
+
+type Node struct {
+	Value    string
+	Children []*Node
+}
+
+func NewTree(value string) *Node {
```
patrickzheng200, line 19: Based on the code, looks like this func should be called `NewNode`.

internal/ioutil/tree.go
```diff
@@ -0,0 +1,58 @@
...
+	n.Children = append(n.Children, node)
+	return node
+}
+
+func (n *Node) Print() {
+	fmt.Println(n.Value)
+
+	if len(n.Children) > 0 {
+		for i, child := range n.Children {
+			print("", !(i < len(n.Children)-1), child)
```
patrickzheng200, line 35: how about:
```suggestion
			print("", i == len(n.Children)-1, child)
```

internal/ioutil/tree.go
shizhMSFT, line 19: If the package name is `tree` instead of `ioutil`, it can just be `New()`. Later, it will be used as `tree.New()`.

internal/ioutil/tree.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+type Node struct {
+	Value    string
+	Children []*Node
+}
+
+func NewTree(value string) *Node {
+	node := Node{Value: value, Children: []*Node{}}
+	return &node
```
shizhMSFT, line 21: Code can be simplified as
```suggestion
	return &Node{Value: value}
```

internal/ioutil/tree.go
```diff
@@ -0,0 +1,58 @@
+package ioutil
```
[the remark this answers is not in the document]
shizhMSFT: Of course, please document and write unit tests accordingly.

internal/ioutil/tree.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func (n *Node) Print() {
+	fmt.Println(n.Value)
+
+	if len(n.Children) > 0 {
+		for i, child := range n.Children {
+			print("", !(i < len(n.Children)-1), child)
+		}
+	}
```
shizhMSFT, line 37: This code somehow duplicates with line 53 - 56.

go.mod
shizhMSFT: Since it is MIT, we can borrow their code.

internal/ioutil/tree.go
```diff
@@ -0,0 +1,58 @@
...
+	Value    string
+	Children []*Node
+}
+
+func NewTree(value string) *Node {
+	node := Node{Value: value, Children: []*Node{}}
+	return &node
+}
+
+func (n *Node) Add(value string) *Node {
```
shizhMSFT, line 24: Since we home brew the package `tree`, we can have functions specific to `notation` like
```go
func (n *Node) AddPair(key, value string) *Node {
    return n.Add(key + ": " + value)
}
```
so that code like
```go
sigNode.Add(fmt.Sprintf("media type: %s", signature.MediaType))
```
can be simplified as
```go
sigNode.AddPair("media type", signature.MediaType)
```

cmd/notation/inspect.go
```diff
@@ -0,0 +1,301 @@
...
+			certNode := certListNode.AddPair("SHA1 fingerprint", cert.SHA1Fingerprint)
+			certNode.AddPair("issued to", cert.IssuedTo)
+			certNode.AddPair("issued by", cert.IssuedBy)
+			certNode.AddPair("expiry", cert.Expiry)
+		}
+
+		artifactNode := sigNode.Add("signed artifact")
+		artifactNode.Add(fmt.Sprintf("media type: %s", signature.SignedArtifact.MediaType))
+		artifactNode.Add(fmt.Sprintf("digest: %s", signature.SignedArtifact.Digest))
+		artifactNode.Add(fmt.Sprintf("size: %d", signature.SignedArtifact.Size))
```
shizhMSFT, line 286: nit: `AddPair()`?

shizhMSFT (dismissed):
LGTM with suggestions

cmd/notation/inspect.go
shizhMSFT: Integers can be formated by https://pkg.go.dev/strconv#FormatInt

priteshbandi (dismissed):
LGTM

priteshbandi (dismissed):

patrickzheng200 (dismissed):
LGTM with all comments addressed.

shizhMSFT (dismissed):
LGTM

priteshbandi (dismissed):
LGTM

shizhMSFT (dismissed):
LGTM

patrickzheng200 (dismissed):
LGTM

patrickzheng200 (approved):
LGTM

shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

PR #541, opened 2023-02-08
Dont access value of default pointer if it is nil

priteshbandi:
Don't access value of default pointer if it is nil. This is actually a bug(unable to delete key if defualt key is not present) fix.

Signed-off-by: Pritesh Bandi <pritesb@amazon.com>


patrickzheng200 (approved):
LGTM

byronchien (approved):

Issue #537, opened 2023-02-05
Simplify signing experience


priteshbandi:
### What is the areas you would like to add the new feature to?

Notation CLI

### Is your feature request related to a problem?

Currently, user need to perform some per-requisites like adding key before they can start signing, which is not the best CX.
Signing steps are
1. Add Key `notation key add --plugin ${plugin} --id ${keyId} ${keyName}`
2. Sign `notation sign -k ${keyName}`

### What solution do you propose?

Notation can accept `plugin` and `id` flag in `sign` command so that user wont need to add key for signing (if they dont want to save it) and they can directly call
```shell
notation sign --plugin ${plugin} --id ${keyId}
```

### What alternatives have you considered?

N/A

### Any additional context?

N/A

mintbomb27:
I was thinking of trying to implement this. I took a look at the code, and I think one approach to do this is to modify the [`GetSigner()`](https://github.com/notaryproject/notation/blob/c47a4522c490f4b1cdcadeb0abd9307b63952394/internal/cmd/signer.go#L15) to create a signer there using `NewFromPlugin()`, by taking the id, plugin name, and configs from the `opts`, and returning that.

I'm new to notation, and would like some help here. Is there any other method to go about this, or should I make any change to my method? Do let me know. @priteshbandi @yizha1

iamsamirzon:
+1 Good idea.

priteshbandi:
@mintbomb27 I've raised a PR https://github.com/notaryproject/notation/pull/553 to update the spec for supporting this scenario. If community agrees then we can merge my PR and then you can work on implementation.

priteshbandi:
Fixed in https://github.com/notaryproject/notation/pull/579

PR #553, opened 2023-02-15
feat(doc): simplify signing experience

priteshbandi:
Related issue https://github.com/notaryproject/notation/issues/537

Signed-off-by: Pritesh Bandi <priteshbandi@gmail.com>

### DIFF
--- a/specs/commandline/sign.md
+++ b/specs/commandline/sign.md
@@ -31,10 +31,12 @@ Flags:
   -d,  --debug                      debug mode
   -e,  --expiry duration            optional expiry that provides a "best by use" time for the artifact. The duration is specified in minutes(m) and/or hours(h). For example: 12h, 30m, 3h20m
   -h,  --help                       help for sign
-  -k,  --key string                 signing key name, for a key previously added to notation's key list.
+       --id string                  key id (required if --plugin is set). This is mutually exclusive with the --key flag
+  -k,  --key string                 signing key name, for a key previously added to notation's key list. This is mutually exclusive with the --id and --plugin flags
   -p,  --password string            password for registry operations (default to $NOTATION_PASSWORD if not specified)
        --plain-http                 registry access via plain HTTP
-       --plugin-config stringArray  {key}={value} pairs that are passed as it is to a plugin, refer plugin's documentation to set appropriate values
+       --plugin string              signing plugin name. This is mutually exclusive with the --key flag
+       --plugin-config stringArray  {key}={value} pairs that are passed as it is to a plugin, refer plugin's documentation to set appropriate values.
        --signature-format string    signature envelope format, options: "jws", "cose" (default "jws")
        --signature-manifest string  manifest type for signature, options: "image", "artifact" (default "artifact")
   -u,  --username string            username for registry operations (default to $NOTATION_USERNAME if not specified)
@@ -66,7 +68,7 @@ Notation uses empty JSON object `{}` as the default configuration content, and t
 
 ## Usage
 
-### Sign an OCI artifact
+### Sign an OCI artifact by adding new key
 
 ```shell
 # Prerequisites:
@@ -85,6 +87,12 @@ An example for a successful signing:
 ```console
 $ notation sign localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
 Successfully signed localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
+```
+
+### Sign an OCI artifact with on-demand remote key
+
+```shell
+notation sign --plugin <plugin_name> --id <remote_key_id> <registry>/<repository>@<digest>
 ```
 
 ### Sign an OCI artifact using COSE signature format


iamsamirzon (approved):

specs/commandline/sign.md
```diff
@@ -31,9 +31,11 @@ Flags:
   -d,  --debug                      debug mode
   -e,  --expiry duration            optional expiry that provides a "best by use" time for the artifact. The duration is specified in minutes(m) and/or hours(h). For example: 12h, 30m, 3h20m
   -h,  --help                       help for sign
+       --id string                  key id (required if --plugin is set)
```
JeyJeyGao, line 34: Is `keyid` a better name for it?

priteshbandi: Don't have strong opinion here but was trying to keep it consistent with `keys add` command

specs/commandline/sign.md
```diff
@@ -73,20 +75,24 @@ Notation uses empty JSON object `{}` as the default configuration content, and t
...
 ```
 
+### Sign an OCI artifact by adding new key
+```shell
+# Add a default signing key referencing the remote key identifier, and the plugin associated with it.
+notation key add --default --name <key_name> --plugin <plugin_name> --id <remote_key_id>
+
+# sign an artifact stored in a registry
+notation sign <registry>/<repository>@<digest>
+```
```
shizhMSFT, line 94: This method should be the suggested way to do signing.

We should call `notation sign --plugin <plugin_name> --id <remote_key_id>` as something like "sign an OCI artifact with on-demand remote key" and make it less perferrable than "Sign an OCI artifact by adding new key".

Otherwise, users will do
```sh
notation sign --plugin <plugin_name> --id <remote_key_id> localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
notation sign --plugin <plugin_name> --id <remote_key_id> localhost:5000/net-monitor@sha256:2d5154312d3727d3d6f3c548ccd96158b001bc713a573cf5569941e0f5c331c8
notation sign --plugin <plugin_name> --id <remote_key_id> localhost:5000/net-monitor@sha256:2033f049a95d2227a5d16b16bc0ef197870b90f1bf95d452531f15c902c4e9bf
notation sign --plugin <plugin_name> --id <remote_key_id> localhost:5000/net-monitor@sha256:e6c2ade679277cb93dacf8b3cb23d05c6368dbfeefd212d3a9b7f7b67e67bc58
```
instead of
```sh
notation key add --default --name <key_name> --plugin <plugin_name> --id <remote_key_id>
notation sign localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
notation sign localhost:5000/net-monitor@sha256:2d5154312d3727d3d6f3c548ccd96158b001bc713a573cf5569941e0f5c331c8
notation sign localhost:5000/net-monitor@sha256:2033f049a95d2227a5d16b16bc0ef197870b90f1bf95d452531f15c902c4e9bf
notation sign localhost:5000/net-monitor@sha256:e6c2ade679277cb93dacf8b3cb23d05c6368dbfeefd212d3a9b7f7b67e67bc58
```

specs/commandline/sign.md
```diff
@@ -73,20 +75,24 @@ Notation uses empty JSON object `{}` as the default configuration content, and t
 # - A signing plugin is installed. See plugin documentation (https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md) for more details.
 # - Configure the signing plugin as instructed by plugin vendor.
 
-# Add a default signing key referencing the remote key identifier, and the plugin associated with it.
-notation key add --default --name <key_name> --plugin <plugin_name> --id <remote_key_id>
-
-# sign an artifact stored in a registry
-notation sign <registry>/<repository>@<digest>
+notation sign --plugin <plugin_name> --id <remote_key_id> <registry>/<repository>@<digest>
```
patrickzheng200, line 78: By looking at this design, basically we are going to support two paths:
1. User adds a new key into `signingkeys.json` with `notation key add` -> Sign using `notation sign --key <key_name>`
2. User does NOT add a new key into `signingkeys.json`, instead Sign directly with `notation sign --plugin <plugin_name> --id <remote_key_id>`

Based on above, I have a couple of questions:
1. What is the expected behavior of this command: `notation sign --plugin <plugin_name> --id <remote_key_id> --key <key_name>`, where the key with <key_name> might have a different <plugin_name> and/or <remote_key_id> than the one specified by user. Looks like `[--plugin, --id]` needs to be mutual exclusive with `--key`?
2. In the 1st path, when adding a key using `notation key add` command, we can do `notation key add --plugin-config`, later on sign with `notation sign --plugin-config`. Under the hood, Notation would merge these two plugin-config maps, and configs from `notation sign` would override the ones from `notation key add`. (Related code: https://github.com/notaryproject/notation-go/blob/8c3ed9217dc8025983f64a5510249a94dc009a86/signer/plugin.go#L170) 
Looks like in the 2nd path, we are asking users to do it themselves by combining all the necessary plugin configs (key + sign) into `notation sign` command's `--plugin-config`. Do we need to mention it somewhere in our command line help page?

specs/commandline/sign.md
priteshbandi: Yeah that sounds reasonable, will update PR to make `notation key add --default --name <key_name> --plugin <plugin_name> --id <remote_key_id>` as preferable method.

specs/commandline/sign.md
priteshbandi: >  1. What is the expected behavior of this command: `notation sign --plugin <plugin_name> --id <remote_key_id> --key <key_name>`, where the key with <key_name> might have a different <plugin_name> and/or <remote_key_id> than the one specified by user. Looks like `[--plugin, --id]` needs to be mutual exclusive with `--key`?

Fail the sign request of both `--key` is used wither with `--plugin` or `--id`


>  2. In the 1st path, when adding a key using `notation key add` command, we can do `notation key add --plugin-config`, later on sign with `notation sign --plugin-config`. Under the hood, Notation would merge these two plugin-config maps, and configs from `notation sign` would override the ones from `notation key add`. (Related code: https://github.com/notaryproject/notation-go/blob/8c3ed9217dc8025983f64a5510249a94dc009a86/signer/plugin.go#L170)
>   When a user is not using `--key` and using `--plugin` + `--id` flag(s) for signing, it means that the user doesn't want a lookup in `sinigngkeys.json` file. Also, technically notation cannot perform a lookup in `sinigngkeys.json`  because there is no `--key` for lookup. This means that we don't have the problem of merging `--plugin-config`. To summarize we will only use plugin-config passed in the command flag

specs/commandline/sign.md
yizha1: +1 To @shizhMSFT.

yizha1 (commented):
I vote for @shizhMSFT 's comment.
Thanks @priteshbandi to introduce a new option for users to sign using on-demand key

specs/commandline/sign.md
```diff
@@ -31,9 +31,11 @@ Flags:
   -d,  --debug                      debug mode
   -e,  --expiry duration            optional expiry that provides a "best by use" time for the artifact. The duration is specified in minutes(m) and/or hours(h). For example: 12h, 30m, 3h20m
   -h,  --help                       help for sign
+       --id string                  key id (required if --plugin is set)
   -k,  --key string                 signing key name, for a key previously added to notation's key list.
   -p,  --password string            password for registry operations (default to $NOTATION_PASSWORD if not specified)
        --plain-http                 registry access via plain HTTP
+       --plugin string              signing plugin name
```
patrickzheng200, line 38: Based on our previous discussion, we should mention this flag is mutual exclusive with the `--key` flag in the description.

specs/commandline/sign.md
```diff
@@ -31,9 +31,11 @@ Flags:
...
        --plugin-config stringArray  {key}={value} pairs that are passed as it is to a plugin, refer plugin's documentation to set appropriate values
```
patrickzheng200, line 39:
```suggestion
       --plugin-config stringArray  {key}={value} pairs that are passed as it is to a plugin, refer plugin's documentation to set appropriate values. If --plugin is set, plugin configs for both key and sign should be included.
```
As I mentioned in previous comments, we might need to add this extra reminder for user. (Related spec: https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md#plugin-configuration)

priteshbandi: IMO we don't need to mention this because it should be evident from not using the key name `-k`. If the key name is not passed then there is no plugin config for the key.

patrickzheng200 (approved):
LGTM

yizha1 (approved):
LGTM but IANAM

shizhMSFT (approved):
LGTM

JeyJeyGao (approved):
LGTM

Issue #549, opened 2023-02-10
Improved UX for Plugin installation/uninstallation


yizha1:
### What is the areas you would like to add the new feature to?

Notation CLI

### Is your feature request related to a problem?

For remote signing, users need to install a signing plugin into notation. Here is the current UX of plugin installation:

1. Download the plugin from the plugin vender repository
2. Figure out what is notation configuration directory based on the OS type
3. Make a directory named `plugins` under notation configuration directory
4. Make a sub-directory under `plugins` for the vendor plugin (of cause, users can use command `mkdir -p`)
5. Extract the plugin package (if needed) and copy the plugin file into directory created in step 4
6. Use command `notation plugin ls` to make sure the plugin is installed

It takes so many steps to just install a plugin, and it is also error prone since most steps are manual steps and user need to figure out the directory path accordingly.

### What solution do you propose?

To improve the plugin installation UX and meet the timeline of v1.0 release. A phased approach is recommended. This issue only addressed the solution for the first phase. 

The goal of the first phase is that users don't need to create directories, remember the names, and verify the installation using `notation plugin ls`. The solution is to introduce two new sub-commands for `notation plugin`:
- `notation plugin --name <plugin name> import <plugin file>`
- `notation plugin delete <plugin name>`

With that, the steps of plugin installation can be reduced to two steps

1. Users download the plugin from the plugin vendor repository.
2. Users execute command `notation plugin import <plugin package>`. The result of the execution indicates the success or failure of installation.

Users can easily delete plugin as well using just one command: `notation plugin delete <plugin name>`. If users forget the plugin name when deleting, users can use the existing command `notation plugin ls` to check the name.

### What alternatives have you considered?

None

yizha1:
Update to RC-3 milestone per community discussion on `Configuration UX improvement`

Duffney:
@yizha1 is there anyone actively working on this feature? If not, I'd like to start working on it. My team and I ran into this exact scenario where each of us were on a different OS and had to figure out how to install the plugins. Imo, adding this feature would greatly improve the user experience and I'd like to help. :)

yizha1:
@Duffney Great and your contributions are welcome! 👍 

Currently it is not assigned to anyone yet. You can take it. Here are some steps we are now practicing on a feature development:
- Share the design/solution of the feature in the community.
[...]

PR #555, opened 2023-02-15
doc(spec): add subcommands to Notation plugin spec

Duffney:
Spec changes to improve the UX for plugin installation and removal using the NotationCLI.

### DIFF
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -36,15 +36,55 @@ Aliases:
   list, ls
 ```
 
+### notation plugin install
+
+```text
+Installs a plugin
+
+Usage:
+  notation plugin install --name <plugin name> --plugin-package <plugin package> [flags]
+
+Flags:
+  -n, --name           string   name of the plugin
+  -p, --plugin-package string   path to the plugin package
+  -h, --help                    help for install
+
+Aliases:
+  install, add
+```
+
+### notation plugin remove 
+
+```text
+Removes a plugin
+
+Usage:
+  notation plugin remove <plugin name> [flags]
+
+Flags:
+  -h, --help          help for remove
+
+Aliases:
+  remove, rm
+```
+
 ## Usage
 
 ### Install a plugin
 
-Currently there is no subcommand available for plugin installation. Plugin publisher should provide instructions to download and install the plugin.
+```shell
+notation plugin install --name <plugin name> --plugin-package <plugin package>
+```
+
+Unpon successful execution, the plugins directory is created under the default plugin directory if it does not exist, and the plugin package is extracted to the plugins directory. The plugin installation is then verified. If the verification fails, the plugin is removed from the plugins directory and an error is returned. 
 
 ### Uninstall a plugin
 
-Currently there is no subcommand available for plugin un-installation. Plugin publisher should provide instructions to uninstall the plugin.
+```shell
+notation plugin remove <plugin name>
+```
+
+Upon successful execution, the plugin is removed from the plugins directory. If the plugin is not found, an error is returned showing the syntax for the plugin list command to show the installed plugins.
 
 ### List installed plugins
 
@@ -58,5 +98,5 @@ An example of output from `notation plugin list`:
 
 ```text
 NAME       DESCRIPTION                                   VERSION             CAPABILITIES                ERROR
-azure-kv   Sign artifacts with keys in Azure Key Vault   v0.3.1-alpha.1      [SIGNATURE_GENERATOR.RAW]   <nil>
+azure-kv   Sign artifacts with keys in Azure Key Vault   v0.5.0-rc.1     [SIGNATURE_GENERATOR.RAW]   <nil>
 ```
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -16,6 +16,8 @@ Usage:
 
 Available Commands:
   list        List installed plugins
+  install     Installs a plugin
+  remove      Removes a plugin
 
 Flags:
   -h, --help          help for plugin
@@ -42,11 +44,9 @@ Aliases:
 Installs a plugin
 
 Usage:
-  notation plugin install [--name <plugin name>] [flags] <plugin package> 
+  notation plugin install [flags] <plugin path>
 
 Flags:
-  -n, --name string             name of the plugin
-  -p, --plugin-package string   path to the plugin package
   -h, --help                    help for install
 
 Aliases:
@@ -73,10 +73,10 @@ Aliases:
 ### Install a plugin
 
 ```shell
-notation plugin install --name <plugin name> --plugin-package <plugin package>
+notation plugin install <plugin path>
 ```
 
-Unpon successful execution, the plugins directory is created under the default plugin directory if it does not exist, and the plugin package is extracted to the plugins directory. The plugin installation is then verified. If the verification fails, the plugin is removed from the plugins directory and an error is returned. 
+Upon successful execution, the plugin is moved to plugins directory. If the plugin directory does not exist, it will be created. The plugin installation is then verified and if the verification fails, the plugin is removed from the plugins directory and an error is returned. When an existing plugin is detected, the versions are compared and if the existing plugin is a lower version then it is replaced by the newer version.
 
 ### Uninstall a plugin
 
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -77,7 +77,7 @@ Aliases:
 notation plugin install <plugin path>
 ```
 
-Upon successful execution, the plugin is moved to plugins directory and name+version of plugin is displayed. If the plugin directory does not exist, it will be created. When an existing plugin is detected, the versions are compared and if the existing plugin is a lower version then it is replaced by the newer version.
+Upon successful execution, the plugin is copied to plugins directory and name+version of plugin is displayed. If the plugin directory does not exist, it will be created. When an existing plugin is detected, the versions are compared and if the existing plugin is a lower version then it is replaced by the newer version.
 
 ### Uninstall a plugin
 
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -65,7 +65,7 @@ Flags:
   -h, --help          help for remove
 
 Aliases:
-  remove, rm
+  remove, rm, uninstall, delete
 ```
 
 ## Usage
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -42,7 +42,7 @@ Aliases:
 Installs a plugin
 
 Usage:
-  notation plugin install --name <plugin name> --plugin-package <plugin package> [flags]
+  notation plugin install [--name <plugin name>] [flags] <plugin package> 
 
 Flags:
   -n, --name           string   name of the plugin
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -45,7 +45,7 @@ Usage:
   notation plugin install [--name <plugin name>] [flags] <plugin package> 
 
 Flags:
-  -n, --name           string   name of the plugin
+  -n, --name string             name of the plugin
   -p, --plugin-package string   path to the plugin package
   -h, --help                    help for install
 
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -48,6 +48,7 @@ Usage:
 
 Flags:
   -h, --help                    help for install
+  -f, --force                   force the installation of a plugin
 
 Aliases:
   install, add
@@ -76,7 +77,7 @@ Aliases:
 notation plugin install <plugin path>
 ```
 
-Upon successful execution, the plugin is moved to plugins directory. If the plugin directory does not exist, it will be created. The plugin installation is then verified and if the verification fails, the plugin is removed from the plugins directory and an error is returned. When an existing plugin is detected, the versions are compared and if the existing plugin is a lower version then it is replaced by the newer version.
+Upon successful execution, the plugin is moved to plugins directory and name+version of plugin is displayed. If the plugin directory does not exist, it will be created. When an existing plugin is detected, the versions are compared and if the existing plugin is a lower version then it is replaced by the newer version.
 
 ### Uninstall a plugin
 
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -59,7 +59,7 @@ Aliases:
 Removes a plugin
 
 Usage:
-  notation plugin remove <plugin name> [flags]
+  notation plugin remove [flags] <plugin name>
 
 Flags:
   -h, --help          help for remove


specs/commandline/plugin.md
```diff
@@ -36,15 +36,55 @@ Aliases:
...
+Removes a plugin
+
+Usage:
+  notation plugin remove <plugin name> [flags]
+
+Flags:
+  -h, --help          help for remove
+
+Aliases:
+  remove, rm
```
shizhMSFT, line 68: How about adding `uninstall` to the aliases?

specs/commandline/plugin.md
```diff
@@ -36,15 +36,55 @@ Aliases:
...
+  install, add
+```
+
+### notation plugin remove 
+
+```text
+Removes a plugin
+
+Usage:
+  notation plugin remove <plugin name> [flags]
```
shizhMSFT, line 62: nit: flags comes before the arguments.
```suggestion
  notation plugin remove [flags] <plugin name>
```

specs/commandline/plugin.md
```diff
@@ -36,15 +36,55 @@ Aliases:
   list, ls
 ```
 
+### notation plugin install
+
+```text
+Installs a plugin
+
+Usage:
+  notation plugin install --name <plugin name> --plugin-package <plugin package> [flags]
```
shizhMSFT, line 45: How about
```suggestion
  notation plugin install [--name <plugin name>] [flags] <plugin package> 
```

shizhMSFT: Later we can have
```console
$ notation plugin install notation-example-foo.tgz
The NEW plugin `example-foo` will be installed.
Do you want to continue? [y/N] y
The plugin `example-foo` is successfully installed.
$ notation plugin install --name example-bar notation-example-bar.tgz
The plugin `example-bar` is successfully installed.
$ notation plugin install --name example-hello notation-example-foo.tgz
Error: notation-example-foo.tgz: example-hello not found
```

shizhMSFT:
Linking the parent issue #549

specs/commandline/plugin.md
Duffney: I like it, that makes a lot of sense. +100

specs/commandline/plugin.md
Duffney: I'm not sure how many aliases are too many, but I added uninstall and delete. :)

specs/commandline/plugin.md
```diff
@@ -36,15 +36,55 @@ Aliases:
   list, ls
 ```
 
+### notation plugin install
+
+```text
+Installs a plugin
```
priteshbandi, line 42: I like this command but have some questions

1. How will install/uninstall work aka whats your vision for this commands? Will it allow execution of custom steps which might be required for setup/cleanup?
2. what should be the format of plugin package?
3. IMO user doesn't needs to specify plugin name, we can extract it using [plugin metada](https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md#plugin-metadata) command

Duffney: Hey @priteshbandi happy to share my thoughts.

1. The Install subcommand would create the necessary directory under the default plugin path and extract/move the plugin to the directory. The delete subcommand would remove the plugin from the directory. If there's more than needs to happen, please lmk.
2. By format of the plugin package do you mean is compressed or not? I was planning to support both formats and optionally expand the compressed version if needed and move the file into the plugin directory.
3. That's a great idea, is that metadata extractable from the executable?

priteshbandi: 1. IMO just providing copy and delete plugin file would be a good start; I was thinking about following enhancemenets and they can be be future work.
    1. add pre-install/post-install(similar for uninstall) script support since plugin might want to perform setup/cleanup.
    2. On install, verifying signature of plugin and displaying its data to user. This will help user verify authenticity and integrity of plugin.
2. yes, by format I meant compressed or not.
3. Yes we/notation should be able to call that command on the executable.

yizha1:
@Duffney Could you fix the DCO issue, see below check results.

specs/commandline/plugin.md
Duffney: @priteshbandi since the `get-plugin-metadata` subcommand can get the plugin-name I agree that the user shouldn't have to specify the plugin-name. @yizha1 and @shizhMSFT what do you think?

yizha1: @Duffney @priteshbandi I like the idea of getting the plugin name from metadata, so that users don't need to specify it. We need to consider the output of installation command, so that users know the plugin name and other information as well.

Some questions:
- How do we know if two plugin files with different names are the same plugin? What is the desired behavior if users install the same plugin multiple times?
- Users install the same plugin but with a higher version, what is the desired behavior?

Duffney: Here are my thoughts:

- Assuming the plugin naming standard of `notation-plugin-name` is followed you can use a string parser to get the plugin name from the binary and then use the `get-metadata` function to return the plugin details and compare the values. Imo if the user installs the same plugin multiple times the desired behavior should be to have the new plugin overwrite\replace the existing plugin.
- If the user installs the same plugin, the plugin version should be compared then displayed to the user with a prompt confirming they wish to overwrite the existing plugin.

specs/commandline/plugin.md
```diff
@@ -36,15 +36,55 @@ Aliases:
...
+
+Usage:
+  notation plugin install [--name <plugin name>] [flags] <plugin package> 
+
+Flags:
+  -n, --name           string   name of the plugin
```
JeyJeyGao, line 48: We may don't need the name of the plugin because the name is the required field of `get-plugin-metadata` plugin command. [plugin spec](https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md#plugin-metadata)
Also, the name of the directory for storing the plugin should be the same with the name returned by `get-plugin-metadata` command, so the user cannot change the name and there is no need to set it.

shizhMSFT: The `--name` flag is optional. It is possible that the plugin in the package is not the intended plugin. Having `--name` adds an additional assurance.

See also https://github.com/notaryproject/notation/pull/555#discussion_r1108298855

shizhMSFT, line 48: nit: spacing
```suggestion
  -n, --name string             name of the plugin
```

specs/commandline/plugin.md
```diff
@@ -36,15 +36,55 @@ Aliases:
...
+  remove, rm, uninstall, delete
+```
+
 ## Usage
 
 ### Install a plugin
 
-Currently there is no subcommand available for plugin installation. Plugin publisher should provide instructions to download and install the plugin.
+```shell
+notation plugin install --name <plugin name> --plugin-package <plugin package>
```
shizhMSFT, line 76: If the `--name` option does not present, we need to prompt the user if the plugin is intended.

yizha1:
@Duffney Could you resolve the comments, so that we can move on?

specs/commandline/plugin.md
Duffney: The [plugin extensibility](https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md#plugin-lifecycle-management) spec lays out a naming standard for the plugin folder and the binary, will notation load the plugin if it doesn't follow the naming convention?

Duffney: Resolving based on the Outdated label, but please feel free to start a new thread or reopen if desired. :)

specs/commandline/plugin.md
Duffney: Resolving per request of @yizha1, please add additional comments if you disagree or have questions with my previous response.

Duffney:
> @Duffney Could you resolve the comments, so that we can move on?

I resolved the open comments, but there are a few unanswered questions around the `--plugin-name` argument. Mainly around what to do when the option isn't provided and how notation would handle plugin-names that doesn't follow the naming standard.

yizha1:
> > @Duffney Could you resolve the comments, so that we can move on?
> 
> I resolved the open comments, but there are a few unanswered questions around the `--plugin-name` argument. Mainly around what to do when the option isn't provided and how notation would handle plugin-names that doesn't follow the naming standard.

@Duffney  I think it is connected to the previous comment from Shiwei, see here https://github.com/notaryproject/notation/pull/555#discussion_r1108298855. The `--plugin-name` is optional and used to confirm it is the plugin that users want to install. That is to say, Users cannot use a different name for the plugin. I would suggest the explanation text for the `--plugin-name` would be used for confirming it is the right plugin, not a new name that users can use.

Here is a copy with additional comments added from me. IMO, Case-1 is the most common case.

Case-1: Name is retrieved from the plugin metadata, but need users to confirm it is the right one.
$ notation plugin install notation-example-foo.tgz
The NEW plugin `example-foo` will be installed.           
Do you want to continue? [y/N] y
The plugin `example-foo` is successfully installed.

Case-2: users specify the name which is the same as the name from plugin metadata, then it will be installed directly
$ notation plugin install --name example-bar notation-example-bar.tgz
The plugin `example-bar` is successfully installed.

Case-3: users specify the name, which is not the same from plugin metadata
$ notation plugin install --name example-hello notation-example-foo.tgz
Error: notation-example-foo.tgz: example-hello not found

specs/commandline/plugin.md
```diff
@@ -36,15 +36,55 @@ Aliases:
...
```
yizha1, line 76: Since `--name` is optional, the main usage example should be the one without the `--name` flag. 
We can add a new usage that for some users know the exact plugin name, and they want to confirm it during installation.

Duffney:
> > > @Duffney Could you resolve the comments, so that we can move on?
> > 
> > 
> > I resolved the open comments, but there are a few unanswered questions around the `--plugin-name` argument. Mainly around what to do when the option isn't provided and how notation would handle plugin-names that doesn't follow the naming standard.
> 
> @Duffney I think it is connected to the previous comment from Shiwei, see here [#555 (comment)](https://github.com/notaryproject/notation/pull/555#discussion_r1108298855). The `--plugin-name` is optional and used to confirm it is the plugin that users want to install. That is to say, Users cannot use a different name for the plugin. I would suggest the explanation text for the `--plugin-name` would be used for confirming it is the right plugin, not a new name that users can use.
> 
> Here is a copy with additional comments added from me. IMO, Case-1 is the most common case.
> 
> Case-1: Name is retrieved from the plugin metadata, but need users to confirm it is the right one. $ notation plugin install notation-example-foo.tgz The NEW plugin `example-foo` will be installed. Do you want to continue? [y/N] y The plugin `example-foo` is successfully installed.
> 
> Case-2: users specify the name which is the same as the name from plugin metadata, then it will be installed directly $ notation plugin install --name example-bar notation-example-bar.tgz The plugin `example-bar` is successfully installed.
> 
> Case-3: users specify the name, which is not the same from plugin metadata $ notation plugin install --name example-hello notation-example-foo.tgz Error: notation-example-foo.tgz: example-hello not found

@yizha1 having the name option confirm the correct plugin name makes perfect sense to me, thanks for the clarification. :)

shizhMSFT (dismissed):
LGTM

patrickzheng200 (dismissed):
LGTM

JeyJeyGao (dismissed):
LGTM

specs/commandline/plugin.md
priteshbandi: Instead of prompting should we just emit the plugin name ?

specs/commandline/plugin.md
```diff
@@ -36,15 +36,55 @@ Aliases:
...
+```
+
+Unpon successful execution, the plugins directory is created under the default plugin directory if it does not exist, and the plugin package is extracted to the plugins directory. The plugin installation is then verified. If the verification fails, the plugin is removed from the plugins directory and an error is returned.
```
priteshbandi, line 79: What are the package formats we are planning to support? .zip?

Duffney: I was thinking `.zip`, `.tar.gz`, `.exe` and the non-extension binary for Darwin and Linux. If the plugin is compressed, only the binary would be extracted and moved to the plugin directory. What are some of the sharp edges you've run into before? I have a prototype working that extracts zip and tar, but I'm certain I haven't tested all the use cases. :)

specs/commandline/plugin.md
Duffney: I agree with @priteshbandi. Having the user type `--name pluginName` validate the plugin adds enough friction that it's a deterrent which reduces the chances that it'll be used for validation. 

Emitting the plugin name might be a better option until it adds more value. For example, if the plugin-name provided a list of trusted plugins you could select from and then the CLI tool downloads and installs it for you.

specs/commandline/plugin.md
priteshbandi: Since `.zip` and `.tar.gz` specification are ambiguous/open, their implementations often suffers from security issues( see [cves](https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=unzip)), here are some the high level issues that I can think of
1. What if there are more than one files(multiple plugins, unrelated files, etc) inside zip, should we copy everything? 
2. What if zip is a [zip bomb](https://en.wikipedia.org/wiki/Zip_bomb) ? 
3. What if unzipping tries to write outside the designated pugin folder ([zip slip](https://security.snyk.io/research/zip-slip-vulnerability))?
4. Should we support symlinks in zip, (see [symlink-vulnerability]( https://effortlesssecurity.in/2020/08/05/zip-symlink-vulnerability/))?

Duffney: I see, it's not as simple as it appears. Starting with the .exe and binary for Darwin and Linux sounds good to me. :)

specs/commandline/plugin.md
Duffney: I'll update the spec to emit the `--name` option.

patrickzheng200 (approved):
LGTM

specs/commandline/plugin.md
```diff
@@ -36,15 +38,53 @@ Aliases:
...
 ## Usage
 
 ### Install a plugin
 
-Currently there is no subcommand available for plugin installation. Plugin publisher should provide instructions to download and install the plugin.
+```shell
+notation plugin install <plugin path>
+```
+
+Upon successful execution, the plugin is moved to plugins directory. If the plugin directory does not exist, it will be created. The plugin installation is then verified and if the verification fails, the plugin is removed from the plugins directory and an error is returned. When an existing plugin is detected, the versions are compared and if the existing plugin is a lower version then it is replaced by the newer version.
```
priteshbandi, line 79:
```suggestion
Upon successful execution, the plugin is moved to plugins directory and name+version of plugin is displayed. If the plugin directory does not exist, it will be created. When an existing plugin is detected, the versions are compared and if the existing plugin is a lower version then it is replaced by the newer version.
```

How can customer lower the version, probably we should add --force command ?

Duffney: Thank you for the feedback and suggestion, I updated the install description and added a `--force` flag to the spec. :)

specs/commandline/plugin.md
```diff
@@ -36,15 +38,54 @@ Aliases:
...
 ## Usage
 
 ### Install a plugin
 
-Currently there is no subcommand available for plugin installation. Plugin publisher should provide instructions to download and install the plugin.
+```shell
+notation plugin install <plugin path>
+```
+
+Upon successful execution, the plugin is moved to plugins directory and name+version of plugin is displayed. If the plugin directory does not exist, it will be created. When an existing plugin is detected, the versions are compared and if the existing plugin is a lower version then it is replaced by the newer version.
```
shizhMSFT, line 80: It would be natural to do `copy` instead of `move`.
```suggestion
Upon successful execution, the plugin is copied to plugins directory and name+version of plugin is displayed. If the plugin directory does not exist, it will be created. When an existing plugin is detected, the versions are compared and if the existing plugin is a lower version then it is replaced by the newer version.
```

specs/commandline/plugin.md
shizhMSFT: @priteshbandi I can answer a few.

> What if there are more than one files(multiple plugins, unrelated files, etc) inside zip, should we copy everything?

Yes, we actually should. It may contain license files, config files for global configs, and other plugin assets.

> What if zip is a [zip bomb](https://en.wikipedia.org/wiki/Zip_bomb) ?

We can limit the max plugin size.

> What if unzipping tries to write outside the designated pugin folder ([zip slip](https://security.snyk.io/research/zip-slip-vulnerability))?

We can restrict that. Actually, `oras` does the same restriction.

> Should we support symlinks in zip, (see [symlink-vulnerability](https://effortlesssecurity.in/2020/08/05/zip-symlink-vulnerability/))?

No, we should not. Symlinks and hard links cause more problem than benefits. Actually, we should support extracting regular files within a flat folder only.

priteshbandi (approved):
LGTM

shizhMSFT (approved):
LGTM

priteshbandi (approved):

patrickzheng200 (approved):
LGTM

Issue #503, opened 2023-01-09
Improve Notation authentication experience


FeynmanZhou:
### What is the areas you would like to add the new feature to?

Notation CLI

### Is your feature request related to a problem?

Notation has a command `notation login` to enable authentication with registries. It should allow users to input a username and password from the prompted response when using `notation login sample.registry.com`.

### What solution do you propose?

The expected experience is as follows:

```
notation login sample.registry.com
Username: xxx
Password: 
Login Succeeded
```

### What alternatives have you considered?

None

FeynmanZhou:
I also detected another problem when using `notation login` with the flag `--password-stdin`, it requires users to input the password explicitly but without any correct return. This should be a bug.

JeyJeyGao:
> I also detected another problem when using `notation login` with the flag `--password-stdin`, it requires users to input the password explicitly but without any correct return. This should be a bug.
> 
> cc @JeyJeyGao

Confirmed the bug.

Duffney:
@FeynmanZhou is it possible to use `docker login` to authenticate notation to a remote registry? Being able to share that docker login credentials would safe the user from having to authenticate to the same registry twice. Once for the container push and a second time for the signing.

ningziwen:
`docker login` or the login in other tools basically store the credentials in a local file by default. For Docker, it is `~/.docker/config.json`.

I like the idea of sharing credentials to prevent users authenticating twice. Instead of sharing crednetial only for Docker, how about adding a credential path feature to let users point any config.json locally?

Users can simply put "DockerConfig" keyword in "credStore" in notation config, then notation will use $DOCKER_CONFIG/config.json as the plain test credential store. ($DOCKER_CONFIG is ~/.docker by default but can be overriden by users)

WDYT? I could try to implement this if this sounds good. @FeynmanZhou

FeynmanZhou:
@ningziwen Thanks for your interest. Yes, I agree with @Duffney and your proposal of sharing and reusing the [Docker credential store](https://docs.docker.com/engine/reference/commandline/login/#credentials-store) within Notation authentication. If users log in to a registry with `docker login`, Notation will be able to reuse `~/.docker/config.json` to authenticate with that registry without additional configuration.

As this issue is tracking an authentication UX bug in `notation login`, shall I create a new issue for implementing this proposal above? Then you can pick it up if you are interested.

ningziwen:
@FeynmanZhou Having a new issue sounds good to me. Thanks!

[...]

PR #566, opened 2023-02-23
feat: Support username and password prompt in login

ningziwen:
https://github.com/notaryproject/notation/issues/503

Added e2e tests. The current unit tests don't have RunE() testing code. Adding the unit test could lead to lots of refactoring. I'm happy to help the refactoring and start the unit tests of RunE() but I think it is better to be in a separated PR. Adding lots of unit test refactoring could mess up this PR.

TODO: Seems login requires credential helper. Need to install credential helper as a test step.

### DIFF
--- a/cmd/notation/login.go
+++ b/cmd/notation/login.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"context"
 	"errors"
 	"fmt"
@@ -11,6 +12,7 @@ import (
 	"github.com/notaryproject/notation/internal/cmd"
 	"github.com/notaryproject/notation/pkg/auth"
 	"github.com/spf13/cobra"
+	"golang.org/x/term"
 	orasauth "oras.land/oras-go/v2/registry/remote/auth"
 )
 
@@ -65,6 +67,23 @@ func runLogin(ctx context.Context, opts *loginOpts) error {
 	// initialize
 	serverAddress := opts.server
 
+	// input username and password by prompt
+	reader := bufio.NewReader(os.Stdin)
+	var err error
+	if opts.Username == "" {
+		opts.Username, err = readUsernameFromPrompt(reader)
+		if err != nil {
+			return err
+		}
+	}
+
+	if opts.Password == "" {
+		opts.Password, err = readPasswordFromPrompt(reader)
+		if err != nil {
+			return err
+		}
+	}
+
 	if err := validateAuthConfig(ctx, opts, serverAddress); err != nil {
 		return err
 	}
@@ -108,7 +127,7 @@ func newCredentialFromInput(username, password string) orasauth.Credential {
 
 func readPassword(opts *loginOpts) error {
 	if opts.passwordStdin {
-		password, err := readLine()
+		password, err := readLine(os.Stdin)
 		if err != nil {
 			return err
 		}
@@ -117,12 +136,41 @@ func readPassword(opts *loginOpts) error {
 	return nil
 }
 
-func readLine() (string, error) {
-	passwordBytes, err := io.ReadAll(os.Stdin)
+func readLine(r io.Reader) (string, error) {
+	passwordBytes, err := io.ReadAll(r)
 	if err != nil {
 		return "", err
 	}
 	password := strings.TrimSuffix(string(passwordBytes), "\n")
 	password = strings.TrimSuffix(password, "\r")
 	return password, nil
+}
+
+func readUsernameFromPrompt(reader *bufio.Reader) (string, error) {
+	fmt.Print("Username: ")
+	username, err := reader.ReadString('\n')
+	if err != nil {
+		return "", fmt.Errorf("error reading username: %w", err)
+	}
+	username = strings.TrimSpace(username)
+	return username, nil
+}
+
+func readPasswordFromPrompt(reader *bufio.Reader) (string, error) {
+	fmt.Print("Password: ")
+	if term.IsTerminal(int(os.Stdin.Fd())) {
+		bytePassword, err := term.ReadPassword(int(os.Stdin.Fd()))
+		if err != nil {
+			return "", fmt.Errorf("error reading password: %w", err)
+		}
+		fmt.Println()
+		return string(bytePassword), nil
+	} else {
+		password, err := readLine(reader)
+		if err != nil {
+			return "", fmt.Errorf("error reading password: %w", err)
+		}
+		fmt.Println()
+		return password, nil
+	}
 }
--- a/test/e2e/suite/command/login.go
+++ b/test/e2e/suite/command/login.go
@@ -1,0 +1,40 @@
+package command
+
+import (
+	"fmt"
+
+	. "github.com/notaryproject/notation/test/e2e/internal/notation"
+	"github.com/notaryproject/notation/test/e2e/internal/utils"
+	. "github.com/notaryproject/notation/test/e2e/suite/common"
+	. "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega/gbytes"
+)
+
+var _ = Describe("notation login", func() {
+	BeforeEach(func() {
+		Skip("The login tests require setting up credential helper running in host and it is not available in Github runner. Issue to remove this skip: https://github.com/notaryproject/notation/issues/587")
+	})
+	It("should sign an image after successfully logging in the registry by prompt with a correct credential", func() {
+		Host(TestLoginOptions(), func(notation *utils.ExecOpts, artifact *Artifact, vhost *utils.VirtualHost) {
+			notation.WithInput(gbytes.BufferWithBytes([]byte(fmt.Sprintf("%s\n%s\n", TestRegistry.Username, TestRegistry.Password)))).
+				Exec("login", artifact.Host).
+				MatchKeyWords(LoginSuccessfully)
+			notation.Exec("sign", artifact.ReferenceWithDigest()).
+				MatchKeyWords(SignSuccessfully)
+			notation.Exec("logout", artifact.Host).
+				MatchKeyWords(LogoutSuccessfully)
+		})
+	})
+
+	It("should fail to sign an image after failing to log in the registry with a wrong credential", func() {
+		Host(TestLoginOptions(), func(notation *utils.ExecOpts, artifact *Artifact, vhost *utils.VirtualHost) {
+			notation.WithInput(gbytes.BufferWithBytes([]byte(fmt.Sprintf("%s\n%s\n", "invalidUser", "invalidPassword")))).
+				ExpectFailure().
+				Exec("login", artifact.Host).
+				MatchErrKeyWords("unauthorized")
+			notation.ExpectFailure().
+				Exec("sign", artifact.ReferenceWithDigest()).
+				MatchErrKeyWords("credential required for basic auth")
+		})
+	})
+})
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/sirupsen/logrus v1.9.0
 	github.com/spf13/cobra v1.6.1
 	github.com/spf13/pflag v1.0.5
+	golang.org/x/term v0.5.0
 	oras.land/oras-go/v2 v2.0.2
 )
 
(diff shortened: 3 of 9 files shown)


ningziwen:
I noticed the current login tests only cover [preRun()](https://github.com/notaryproject/notation/blob/main/cmd/notation/login_test.go#L29). Is there any reason of doing this? Do we want to also cover RunE()? My change is in RunE() so it won't be covered in the test file following current tests.

priteshbandi:
> I noticed the current login tests only cover [preRun()](https://github.com/notaryproject/notation/blob/main/cmd/notation/login_test.go#L29). Is there any reason of doing this? Do we want to also cover RunE()? My change is in RunE() so it won't be covered in the test file following current tests.

Can we create new tests for this usecase?

JeyJeyGao:
We may need to add an E2E test for it because username and password are read from STDIN.  The E2E test is [here](https://github.com/notaryproject/notation/tree/main/test/e2e).

ningziwen:
I noticed the unit tests only produce the commands instead of running the commands. E2e tests run the commands. So I added the tests to e2e tests.

test/e2e/suite/command/login.go
```diff
@@ -0,0 +1,35 @@
...
+				Exec("login", artifact.ReferenceWithDigest()).
+				MatchKeyWords(LoginSuccessfully)
+
+			notation.Exec("sign", artifact.ReferenceWithDigest()).
+				MatchKeyWords(SignSuccessfully)
+		})
+	})
+
+	It("should fail to sign an image after failing to log in the registry with a wrong credential", func() {
+		Host(NoAuthOptions(gbytes.BufferWithBytes([]byte("y"))), func(notation *utils.ExecOpts, artifact *Artifact, vhost *utils.VirtualHost) {
```
JeyJeyGao, line 27: We may need to fix the build error.

ningziwen:
@JeyJeyGao I completed most of the e2e tests but the only thing left is credential helper. 

I noticed it seems notation forces users to have credential helper. If not, it will directly throw error
```
Error: could not get the credentials store: failed to load config file, error: credentials store config was not set up
```

This is different with other container tools. For example, in nerdctl, it throws `WARNING: Your password will be stored unencrypted in %s.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store` and succeeds the login.

I'm wondering is it intentionally to enforce notation users to have credential helper?

JeyJeyGao:
> @JeyJeyGao I completed most of the e2e tests but the only thing left is credential helper.
> 
> I noticed it seems notation forces users to have credential helper. If not, it will directly throw error
> 
> ```
> Error: could not get the credentials store: failed to load config file, error: credentials store config was not set up
> ```
> 
> This is different with other container tools. For example, in nerdctl, it throws `WARNING: Your password will be stored unencrypted in %s. Configure a credential helper to remove this warning. See https://docs.docker.com/engine/reference/commandline/login/#credentials-store` and succeeds the login.
> 
> I'm wondering is it intentionally to enforce notation users to have credential helper?

Yes, it depends on credential helper now, but in the next few months, login without credential helper will be supported. Currently, a `credential-go` library is under development, which will be used by notation. 

If the Github action machine doesn't have a credential helper, we cannot do `notation login`. To solve the issue, 
1. we can update `run.sh` script to install a credential helper for linux if there is no one. You can follow the example [here](https://github.com/docker/docker-credential-helpers). 
> By default, Docker looks for the native binary on each of the platforms, i.e. “osxkeychain” on macOS, “wincred” on windows, and “pass” on Linux. [link](https://docs.docker.com/engine/reference/commandline/login/)
2. wait for another few months then notation will support login without credential helper.😄

ningziwen:
@JeyJeyGao Great! Could you point me the link of credential-go? Login without credential helper is pretty valuable to my use case. Just want to track the progress and see if there is anything I can help.

JeyJeyGao:
> @JeyJeyGao Great! Could you point me the link of credential-go? Login without credential helper is pretty valuable to my use case. Just want to track the progress and see if there is anything I can help.

https://github.com/oras-project/oras-credentials-go

ningziwen:
@JeyJeyGao Pushed the changes except the credential helper in Github action. I set up credential helper in my host and current e2e tests work find.

In my investigation, to set up credential helper, there are 3 steps
1. `gpg2 --quick-generate-key notation-e2e`. This step needs users to interfact to input passphase and I don't find a way to execute quietly. This is a blocker of running it in Github action.
2. Install docker-credential-pass.
3. Set `credstore` to `pass` in ~/.config/notation/config.json

Let me know if you have any advice to deal with gpg2. Otherwise we may temporarily skip the automated e2e tests or wait for credential-go, but @FeynmanZhou may want to have this sooner than later.

ningziwen:
@JeyJeyGao I added Skip() to the new e2e tests to unblock Github actions, and created an issue to track with puting it in the comment. https://github.com/notaryproject/notation/issues/587
Let me know if you have better suggestions.

cmd/notation/login.go
```diff
@@ -126,3 +146,32 @@ func readLine() (string, error) {
...
+	if err != nil {
+		return "", fmt.Errorf("error reading username: %w", err)
+	}
+	username = strings.TrimSpace(username)
+	return username, nil
+}
+
+func readPasswordFromPrompt(reader *bufio.Reader) (string, error) {
+	fmt.Print("Password: ")
+	if term.IsTerminal(syscall.Stdin) {
```
JeyJeyGao, line 162: cmd\notation\login.go:162:21: cannot use syscall.Stdin (variable of type syscall.Handle) as int value in argument to term.IsTerminal
cmd\notation\login.go:163:42: cannot use syscall.Stdin (variable of type syscall.Handle) as int value in argument to term.ReadPassword

I tried to compile it on Windows, but it didn't work. We may need to convert `syscall.Stdin` to be an `int` type value

ningziwen: Thanks. Tried the latest change in my windows desktop and it can build now.

cmd/notation/login.go
```diff
@@ -65,6 +68,23 @@ func runLogin(ctx context.Context, opts *loginOpts) error {
...
+			return err
+		}
+	}
+
+	if opts.Password == "" {
+		opts.Password, err = readPasswordFromPrompt(reader)
+		if err != nil {
+			return err
+		}
+	}
```
shizhMSFT, line 86: Should this part be refactored with the `readPassword()` function in `PreRunE`?

cmd/notation/login.go
```diff
@@ -126,3 +146,32 @@ func readLine() (string, error) {
...
+	if err != nil {
+		return "", fmt.Errorf("error reading username: %w", err)
+	}
+	username = strings.TrimSpace(username)
+	return username, nil
+}
+
+func readPasswordFromPrompt(reader *bufio.Reader) (string, error) {
+	fmt.Print("Password: ")
+	if term.IsTerminal(int(syscall.Stdin)) {
```
shizhMSFT, line 162: Will `os.Stdin.Fd()` be better than `syscall.Stdin` for platform independency?

cmd/notation/login.go
```diff
@@ -126,3 +146,32 @@ func readLine() (string, error) {
...
+		}
+		fmt.Println()
+		return string(bytePassword), nil
+	} else {
+		password, err := reader.ReadString('\n')
+		if err != nil {
+			return "", fmt.Errorf("error reading password: %w", err)
+		}
+		password = strings.TrimSpace(password)
+		return password, nil
```
shizhMSFT, line 175: This part is duplicated with `readPassword` at line 129.

cmd/notation/login.go
```diff
@@ -126,3 +146,32 @@ func readLine() (string, error) {
...
+			return "", fmt.Errorf("error reading password: %w", err)
+		}
+		fmt.Println()
+		return string(bytePassword), nil
+	} else {
+		password, err := reader.ReadString('\n')
+		if err != nil {
+			return "", fmt.Errorf("error reading password: %w", err)
+		}
+		password = strings.TrimSpace(password)
```
shizhMSFT, line 174: Unlike the username, it is possible that spaces are part of the password.

cmd/notation/login.go
ningziwen: I think prompt happens in RunE instead of PreRunE. It is very different with readPassword() in preRunE which reads from piped input. Let me know if you have better ideas to refactor it.

cmd/notation/login.go
```diff
@@ -65,6 +67,23 @@ func runLogin(ctx context.Context, opts *loginOpts) error {
 	// initialize
 	serverAddress := opts.server
 
+	// input username and password by prompt
+	reader := bufio.NewReader(os.Stdin)
+	var err error
+	if opts.Username == "" {
+		opts.Username, err = readUsernameFromPrompt(reader)
```
qweeah, line 74: If `opts.Username` is read as an empty string with no error, `opts.Password` will be [considered as a token](https://github.com/notaryproject/notation/blob/b7b5fcf56137a0ece3fbc09c608a62860f20a050/cmd/notation/login.go#L123). In that case, should we change the prompt to `Token: ` rather than `Password: `?

priteshbandi: IMO accepting token is an enhancement and should be done when Notation starts supporting token in cli argument like `-u` and `-p`.

ningziwen: I think "Password" is consistent with the --password flag, and also has compliance with container development tools like Docker.

qweeah: > IMO accepting token is an enhancement and should be done when Notation starts supporting token in cli argument like -u and -p.

Notation already supports token if username is empty, although it's reflected in the UX. But I agree that it's an nice-to-have enhancement which is not necessarily to be included in this PR.

qweeah: > I think "Password" is consistent with the --password flag

Yes, not trying to break the UX. Just suggest a more accurate prompt for interactive login. `--password` and `--stdin-password` flags are targeting non-interactive login.

> and also has compliance with container development tools like Docker.

From compliance side, I think Notation only needs to make sure that the generated configuration file is compliant to Docker, not UX. Notation already make changes to login UX, e.g. `docker login` doesn't allow empty username. User must provide empty GUID as username if they want to use a token. Notation allows empty username.

byronchien (dismissed):

priteshbandi (approved):
LGTM

cmd/notation/login.go
```diff
@@ -117,12 +136,40 @@ func readPassword(opts *loginOpts) error {
...
+}
+
+func readPasswordFromPrompt(reader *bufio.Reader) (string, error) {
+	fmt.Print("Password: ")
+	if term.IsTerminal(int(os.Stdin.Fd())) {
+		bytePassword, err := term.ReadPassword(int(os.Stdin.Fd()))
+		if err != nil {
+			return "", fmt.Errorf("error reading password: %w", err)
+		}
+		fmt.Println()
```
patrickzheng200, line 166: nit:
do we also need `fmt.Println()` in the `else` statement below?

patrickzheng200 (dismissed):
LGTM with nit comment.

cmd/notation/login.go
ningziwen: Good call. Added.

patrickzheng200 (approved):
LGTM

cmd/notation/login.go
```diff
@@ -117,12 +136,41 @@ func readPassword(opts *loginOpts) error {
...
+	if err != nil {
+		return "", fmt.Errorf("error reading username: %w", err)
+	}
+	username = strings.TrimSpace(username)
+	return username, nil
+}
+
+func readPasswordFromPrompt(reader *bufio.Reader) (string, error) {
+	fmt.Print("Password: ")
+	if term.IsTerminal(int(os.Stdin.Fd())) {
```
shizhMSFT, line 161: nit: There is an edge case but we can fix later. It is possible that the password has been read by the `bufio.Reader` so that `os.Stdin` has nothing to read. This edge case can only be possibly triggered by the following steps

1. Open a text editor and type username and password there.
2. Open a terminal and run `notation login`.
3. Copy the username and password from the text editor to the terminal and hit enter
4. `notation` will ask the user to type the password again, which is not expected.

shizhMSFT (approved):
LGTM with suggestions.

priteshbandi (approved):

cmd/notation/login.go
ningziwen: Oh I just tried your steps with pasting the username and password from text editor, and it works for me. The login succeeded.

PR #568, opened 2023-02-25
doc: create CLI spec for managing trust policies (phase 1)

yizha1:
Update:
- Create CLI spec for managing trust policies for phase 1

### DIFF
--- a/specs/commandline/policy.md
+++ b/specs/commandline/policy.md
@@ -63,7 +63,7 @@ The `notation policy` command provides a user-friendly way to manage trust polic
               "ca:acme-rockets"
             ],                  
             "trustedIdentities": [                                    
-                "x509.subject: C=US, ST=WA, L=Seattle, O=acme-rockets.io, CN=SecureBuilder",
+                "x509.subject: C=US, ST=WA, L=Seattle, O=acme-rockets.io, CN=SecureBuilder"
             ]
         }
     ]
--- a/specs/commandline/policy.md
+++ b/specs/commandline/policy.md
@@ -9,7 +9,7 @@ The `notation policy` command provides a user-friendly way to manage trust polic
 - The Policy named "wabbit-networks-images" is for verifying images signed by Wabbit Networks and stored in two repositories `registry.acme-rockets.io/software/net-monitor` and `registry.acme-rockets.io/software/net-logger`.
 
 - Policy named "unsigned-image" is for skipping the verification on unsigned images stored in repository `registry.acme-rockets.io/software/unsigned/net-utils`.
-- Policy "allow-expired-images" is for logging instead of failing expired images stored in repository `"registry.acme-rockets.io/software/legacy/metrics`.
+- Policy "allow-expired-images" is for logging instead of failing expired images stored in repository `registry.acme-rockets.io/software/legacy/metrics`.
 - Policy "global-policy-for-all-other-images" is for verifying any other images that signed by the ACME Rockets.
   
 ```jsonc
--- a/specs/commandline/policy.md
+++ b/specs/commandline/policy.md
@@ -110,11 +110,11 @@ Usage:
   notation policy init [flags]
 
 Flags:
-  -d, --debug        debug mode
-      --force        override the existing trust polices, never prompt
-  -h, --help         help for export
-      --ts string    specify the trust store in format "<type>:<name>", e.g. "ca:my_store"
-  -v, --verbose      verbose mode
+  -d, --debug             debug mode
+      --force             override the existing trust polices, never prompt
+  -h, --help              help for export
+      --ts stringArray    specify the trust stores in format "<type>:<name>", e.g. "ca:my_store"
+  -v, --verbose           verbose mode
 ```
 
 ### notation policy show
@@ -132,16 +132,6 @@ Flags:
 ```
 
 ## Usage
-
-### Import trust policies from a JSON file
-
-```shell  
-notation policy import ./my_policy.json
-```
-
-The trust policies in the JSON file will be validated according to [trust policy properties](https://github.com/notaryproject/notaryproject/blob/v1.0.0-rc.2/specs/trust-store-trust-policy.md#trust-policy-properties). A successful message should be printed out if trust policies are imported successfully. Error logs including the reason should be printed out if the importing fails.
-
-Use `--force` flag to override existing policies without prompt.
 
 ### Initialize trust policies with default values
 
@@ -171,10 +161,10 @@ Upon successful execution, trust policies with default values are created as fol
 
 Use `--force` flag to override existing policies without prompt.
 
-### Initialize trust policies with specified trust store
+### Initialize trust policies with specified trust stores
 
 ```shell
-notation policy init --ts "ca:my_store"
+notation policy init --ts "ca:my_store" --ts "ca:my_store_2"
 ```
 
 Upon successful execution, trust policies with default values and specified trust stores are created as following:
@@ -189,7 +179,7 @@ Upon successful execution, trust policies with default values and specified trus
             "signatureVerification": {
                 "level": "strict"
             },
-            "trustStores": ["ca:my_store"],
+            "trustStores": ["ca:my_store", "ca:my_store_2"],
             "trustedIdentities": ["*"]
         }
     ]
@@ -213,6 +203,16 @@ Users can redirect the output of command `notation policy show` to a JSON file.
 ```shell
 notation policy show > ./trust_policy.json
 ```
+
+### Import trust policies from a JSON file
+
+```shell  
+notation policy import ./my_policy.json
+```
+
+The trust policies in the JSON file will be validated according to [trust policy properties](https://github.com/notaryproject/notaryproject/blob/v1.0.0-rc.2/specs/trust-store-trust-policy.md#trust-policy-properties). A successful message should be printed out if trust policies are imported successfully. Error logs including the reason should be printed out if the importing fails.
+
+Use `--force` flag to override existing policies without prompt.
 
 ### Update trust policies
 
--- a/specs/commandline/policy.md
+++ b/specs/commandline/policy.md
@@ -2,50 +2,77 @@
 
 ## Description
 
-As part of signature verification workflow, users need to configure the trust policies to specify trusted identities that sign the artifacts, and the level of signature verification to use. For more details, see [trust policy spec](https://github.com/notaryproject/notaryproject/blob/v1.0.0-rc.2/specs/trust-store-trust-policy.md#trust-policy).
+As part of signature verification workflow, users need to configure the trust policies to specify trusted identities that signed the artifacts, and the level of signature verification to use. For more details, see [trust policy specification and examples](https://github.com/notaryproject/notaryproject/blob/v1.0.0-rc.2/specs/trust-store-trust-policy.md#trust-policy).
 
-An example of trust policy configuration file:
+The `notation policy` command provides a user-friendly way to manage trust policies. It allows users to import and export trust policies from/to a JSON file. Users can export a template file with trust policy configuration to start from scratch.
+
+An example of template file:
 
 ```jsonc
 {
-    "version": "1.0",                                                   // version info
-    "trustPolicies": [                                                  // list of trust policy statements
+    "version": "1.0",
+    "trustPolicies": [
         {
-            "name": "wabbit-networks-dev",                              // Name of the first policy statement
-            "registryScopes": [ "dev.wabbitnetworks.io/net-monitor" ],  // The registry artifacts to which the policy applies
-            "signatureVerification": {                                  // The level of verification - strict, permissive, audit, skip
+            // Policy for set of artifacts signed by Wabbit Networks
+            // that are pulled from ACME Rockets repository
+            "name": "wabbit-networks-images",
+            "registryScopes": [ 
+              "registry.acme-rockets.io/software/net-monitor",
+              "registry.acme-rockets.io/software/net-logger" 
+            ],
+            "signatureVerification": {
                 "level": "strict"
-                "override" : {
-                     "expiry" : "log",
-                     "authenticity": "log"
-                }
             },
-            "trustStores": [ "ca:wabbit-networks-dev" ],                // The trust stores that contains the X.509 certificates
-            "trustedIdentities": [                                      // Identities that are trusted to sign the artifact.
-                "x509.subject: C=US, ST=WA, L=Seattle, O=Example, OU=Dev, CN=wabbit-networks.io"
+            "trustStores": [ 
+              "ca:wabbit-networks",
+              "ca:wabbit-networks-ca2"
+            ],
+            "trustedIdentities": [
+                "x509.subject: C=US, ST=WA, L=Seattle, O=wabbit-networks.io, OU=Security Tools"
             ]
         },
         {
-            "name": "wabbit-networks-prod",                             // Name of the second policy statement.
-            "registryScopes": [ "prod.wabbitnetworks.io/net-monitor" ],       
-            "signatureVerification": {                                
-                "level": "permissive"
+            // Exception policy for a single unsigned artifact pulled from
+            // Wabbit Networks repository
+            "name": "unsigned-image",
+            "registryScopes": [ "registry.wabbit-networks.io/software/unsigned/net-utils" ],
+            "signatureVerification": {
+              "level" : "skip" 
+            }
+        },
+        {
+            // Policy that uses custom verification level to relax the strict verification.
+            // It logs expiry and skips revocation check for a specific artifact.
+            "name": "allow-expired-images",
+            "registryScopes": [ "registry.acme-rockets.io/software/legacy/metrics" ],
+            "signatureVerification": {
+              "level" : "strict",
+              "override" : {
+                "expiry" : "log",
+                "revocation" : "skip"
+              }
             },
-            "trustStores": [ "ca:wabbit-networks-prod" ],                  
+            "trustStores": ["ca:acme-rockets"],
+            "trustedIdentities": ["*"]
+        },
+        {
+            // Policy for all other artifacts signed by ACME Rockets
+            "name": "global-policy-for-all-other-images",
+            "registryScopes": [ "*" ],       
+            "signatureVerification": {                                
+                "level": "strict"
+            },
+            "trustStores": [ 
+              "ca:acme-rockets-others"
+            ],                  
             "trustedIdentities": [                                    
-                "x509.subject: C=US, ST=WA, L=Seattle, O=Example, OU=Prod, CN=wabbit-networks.io"
+                "x509.subject: C=US, ST=WA, L=Seattle, O=acme-rockets.io, OU=Finance, CN=SecureBuilder",
+                "x509.subject: C=US, ST=WA, L=Seattle, O=acme-rockets.io, OU=Hr, CN=SecureBuilder"
             ]
         }
     ]
 }
 ```
-
-The goal of `notation policy` command is to provide a user-friendly CLI for users to manage trust policies without the knowledge of trust policy configuration name, directory path and property names. Two high level use cases as follows:
-
-1. Users can import trust policies from a JSON file, and export trust policies into a JSON file.
-2. Users can add/view/update/delete trust policies and properties without editing the trust policy configuration file.
-
-A phased approach is adopted to achieve the goal. The first use case will be implemented in phase-1. A trust policy template file will be provided for users to get started. This specification only covers phase-1.
 
 ## Outline
 
@@ -75,7 +102,10 @@ Usage:
   notation policy export [flags] <file_path>
 
 Flags:
-  -h, --help    help for export
+  -d, --debug     debug mode
+  -h, --help      help for export
+      --template  export a template of trust policies
+  -v, --verbose   verbose mode
 ```
 
 ### notation policy import
@@ -87,7 +117,9 @@ Usage:
   notation policy import [flags] <file_path>
 
 Flags:
-  -h, --help    help for import
+  -d, --debug     debug mode
+  -h, --help      help for import
+  -v, --verbose   verbose mode
 ```
 
 ### notation policy show
@@ -99,7 +131,9 @@ Usage:
   notation policy [flags] show
 
 Flags:
-  -h, --help    help for show
+  -d, --debug     debug mode
+  -h, --help      help for show
+  -v, --verbose   verbose mode
 ```
 
 ## Usage
@@ -110,7 +144,15 @@ Flags:
 notation policy import ./my_policy.json
 ```
 
-The trust policies in the JSON file will be validated according to [trust policy properties](https://github.com/notaryproject/notaryproject/blob/v1.0.0-rc.2/specs/trust-store-trust-policy.md#trust-policy-properties). Upon successful import, the trust policies are printed out.
+The trust policies in the JSON file will be validated according to [trust policy properties](https://github.com/notaryproject/notaryproject/blob/v1.0.0-rc.2/specs/trust-store-trust-policy.md#trust-policy-properties). A successful message should be printed out if trust policies are imported successfully. Error logs including the reason should be printed out if the importing fails.  
+
+### Export a template file of trust policies
+
+The template file is for users to create trust policies from scratch. Users should update the trust policies according to own requirements before importing the template file.
+
+```shell
+notation policy export ./trustpolicy_template.json
+```
 
 ### Export trust policies into a JSON file
 
@@ -118,7 +160,7 @@ The trust policies in the JSON file will be validated according to [trust policy
 notation policy export ./policy_exported.json
 ```
 
-For phase-1, to update trust policies, users need to export the trust policies to a file first, update the file, and import the file again.
+Upon successful execution, the trust policies are exported into a json file.
 
 ### Show trust policies
 
@@ -126,4 +168,4 @@ For phase-1, to update trust policies, users need to export the trust policies t
 notation policy show
 ```
 
-Upon successful execution, the trust policies are printed out. If trust policies are not configured, users should receive a warning message.
+Upon successful execution, the trust policies are printed out. If trust policies are not configured, users should receive an error message.
--- a/specs/commandline/policy.md
+++ b/specs/commandline/policy.md
@@ -2,9 +2,9 @@
 
 ## Description
 
-As part of signature verification workflow, users need to configure the trust policies to specify trusted identities that signed the artifacts, and the level of signature verification to use. For more details, see [trust policy specification and examples](https://github.com/notaryproject/notaryproject/blob/v1.0.0-rc.2/specs/trust-store-trust-policy.md#trust-policy).
+As part of signature verification workflow, users need to configure the trust policy configuration file to specify trusted identities that signed the artifacts, the level of signature verification to use and other settings. For more details, see [trust policy specification and examples](https://github.com/notaryproject/notaryproject/blob/v1.0.0-rc.2/specs/trust-store-trust-policy.md#trust-policy).
 
-The `notation policy` command provides a user-friendly way to manage trust policies. It allows users to import/export trust policies from/to a JSON file, and show trust policies. To get started user can use following sample trust policy. In this sample, there are four policies configured for different requirements:
+The `notation policy` command provides a user-friendly way to manage trust policies. It allows users to show trust policy configuration, import/export a trust policy configuration file from/to a JSON file. To get started user can refer to the following trust policy configuration sample. In this sample, there are four policies configured for different requirements:
 
 - The Policy named "wabbit-networks-images" is for verifying images signed by Wabbit Networks and stored in two repositories `registry.acme-rockets.io/software/net-monitor` and `registry.acme-rockets.io/software/net-logger`.
 - Policy named "unsigned-image" is for skipping the verification on unsigned images stored in repository `registry.acme-rockets.io/software/unsigned/net-utils`.
@@ -72,14 +72,14 @@ The `notation policy` command provides a user-friendly way to manage trust polic
 ### notation policy command
 
 ```text
-Manage trust policies for signature verification.
+Manage trust policy configuration for signature verification.
 
 Usage:
   notation policy [command]
 
 Available Commands:
-  import    import trust policies from a JSON file
-  show      show trust policies
+  import    import trust policy configuration from a JSON file
+  show      show trust policy configuration
 
 Flags:
   -h, --help   help for policy
@@ -88,14 +88,14 @@ Flags:
 ### notation policy import
 
 ```text
-Import trust policies from a JSON file
+Import trust policy configuration from a JSON file
 
 Usage:
   notation policy import [flags] <file_path>
 
 Flags:
   -d, --debug     debug mode
-      --force     override the existing trust policies, never prompt
+      --force     override the existing trust policy configuration, never prompt
   -h, --help      help for import
   -v, --verbose   verbose mode
 ```
@@ -103,7 +103,7 @@ Flags:
 ### notation policy show
 
 ```text
-Show trust policies
+Show trust policy configuration
 
 Usage:
   notation policy show [flags]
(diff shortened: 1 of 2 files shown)


specs/commandline/policy.md
```diff
@@ -0,0 +1,129 @@
...
+    ]
+}
+```
+
+The goal of `notation policy` command is to provide a user-friendly CLI for users to manage trust policies without the knowledge of trust policy configuration name, directory path and property names. Two high level use cases as follows:
+
+1. Users can import trust policies from a JSON file, and export trust policies into a JSON file.
+2. Users can add/view/update/delete trust policies and properties without editing the trust policy configuration file.
+
+A phased approach is adopted to achieve the goal. The first use case will be implemented in phase-1. A trust policy template file will be provided for users to get started. This specification only covers phase-1.
```
priteshbandi, line 48:
```suggestion
The `notation policy` command provides a user-friendly way to manage trust policies. It allows users to import, export trust policies from/to a JSON file.
```

As we are not writing a spec for add/view/update/delete, we should omit its reference from this spec , that way we won't have to say that we are only implementing 1st phase.

Also, we can/will update trust policy spec again, when we support 2nd phase.

specs/commandline/policy.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Usage
+
+### Import trust policies from a JSON file
+
+```shell  
+notation policy import ./my_policy.json
+```
+
+The trust policies in the JSON file will be validated according to [trust policy properties](https://github.com/notaryproject/notaryproject/blob/v1.0.0-rc.2/specs/trust-store-trust-policy.md#trust-policy-properties). Upon successful import, the trust policies are printed out.
```
priteshbandi, line 113: Do we really need to print trust policy after successful import as trust policies can be big?

specs/commandline/policy.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Export trust policies into a JSON file
+
+```shell
+notation policy export ./policy_exported.json
+```
+
+For phase-1, to update trust policies, users need to export the trust policies to a file first, update the file, and import the file again.
```
priteshbandi, line 121: There is a usecase where users have stored trust policy in some other location and they just want to import it.

Also, as mentioned earlier we should remove the `phase-1` string because the user doesn't know which phase we are in and we might forget to remove phase-1 when we implement phase-2.

specs/commandline/policy.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Show trust policies
+
+```shell
+notation policy show
+```
+
+Upon successful execution, the trust policies are printed out. If trust policies are not configured, users should receive a warning message.
```
priteshbandi, line 129:
```suggestion
Upon successful execution, the trust policies are printed out. If trust policies are not configured, users should receive a error message.
```

yizha1:
@priteshbandi A new option was added for `notation policy export` to export a template file of trust policy for users to start from scratch.

JeyJeyGao:
How to update a trust policy? Does the `import` command needs to merge the new trust policy with the old one? Or just replace the old one?

yizha1:
> How to update a trust policy? Does the `import` command needs to merge the new trust policy with the old one? Or just replace the old one?

The import command will replace the old one. To update a trust policy, users can export the existing policies to a file, update the file and import it again.  The second phase for managing trust policy is to introduce new commands for add/update/delete policies.

specs/commandline/policy.md
```diff
@@ -0,0 +1,171 @@
...
+
+An example of template file:
+
+```jsonc
+{
+    "version": "1.0",
+    "trustPolicies": [
+        {
+            // Policy for set of artifacts signed by Wabbit Networks
+            // that are pulled from ACME Rockets repository
```
shizhMSFT, line 17: I know this is for documentation. In the real implementation, do we support _JSON with comments_ as input?

specs/commandline/policy.md
```diff
@@ -0,0 +1,171 @@
...
+
+The trust policies in the JSON file will be validated according to [trust policy properties](https://github.com/notaryproject/notaryproject/blob/v1.0.0-rc.2/specs/trust-store-trust-policy.md#trust-policy-properties). A successful message should be printed out if trust policies are imported successfully. Error logs including the reason should be printed out if the importing fails.  
+
+### Export a template file of trust policies
+
+The template file is for users to create trust policies from scratch. Users should update the trust policies according to own requirements before importing the template file.
+
+```shell
+notation policy export --template ./trustpolicy_template.json
+```
```
shizhMSFT, line 155: How about `notation policy init`?

yizha1: Does the `init` command initialize the trust policies configurations? If yes, users still need to export and update the template file.

specs/commandline/policy.md
yizha1: Maybe we can use comments to indicate which fields should be updated for a template file. I am not sure whether there are other ways.

specs/commandline/policy.md
```diff
@@ -0,0 +1,171 @@
+# notation policy
+
+## Description
+
+As part of signature verification workflow, users need to configure the trust policies to specify trusted identities that signed the artifacts, and the level of signature verification to use. For more details, see [trust policy specification and examples](https://github.com/notaryproject/notaryproject/blob/v1.0.0-rc.2/specs/trust-store-trust-policy.md#trust-policy).
+
+The `notation policy` command provides a user-friendly way to manage trust policies. It allows users to import and export trust policies from/to a JSON file. Users can export a template file with trust policy configuration to start from scratch.
+
+An example of template file:
```
priteshbandi, line 9:
```suggestion
The `notation policy` command provides a user-friendly way to manage trust policies. It allows users to import and export trust policies from/to a JSON file. To get started user can use following sample trust policy (when using sample trust policy please remove all comments):
```

specs/commandline/policy.md
```diff
@@ -0,0 +1,171 @@
...
+```text
+Export trust policies to a JSON file
+
+Usage:
+  notation policy export [flags] <file_path>
+
+Flags:
+  -d, --debug     debug mode
+  -h, --help      help for export
+      --template  export a template of trust policies
```
priteshbandi, line 107: There is no trustpolicy template and a single trust policy template wont work for user. IMO we should start small and drop this flag.

specs/commandline/policy.md
```diff
@@ -0,0 +1,171 @@
...
+  -h, --help      help for show
+  -v, --verbose   verbose mode
+```
+
+## Usage
+
+### Import trust policies from a JSON file
+
+```shell  
+notation policy import ./my_policy.json
```
priteshbandi, line 144: When an trust policy exists this command will override it? If yes, we should add `-f` aka `--force` flag to capture the intent.

specs/commandline/policy.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+The trust policies in the JSON file will be validated according to [trust policy properties](https://github.com/notaryproject/notaryproject/blob/v1.0.0-rc.2/specs/trust-store-trust-policy.md#trust-policy-properties). A successful message should be printed out if trust policies are imported successfully. Error logs including the reason should be printed out if the importing fails.  
+
+### Export a template file of trust policies
+
+The template file is for users to create trust policies from scratch. Users should update the trust policies according to own requirements before importing the template file.
+
+```shell
+notation policy export --template ./trustpolicy_template.json
```
priteshbandi, line 154: If destination file exists will this command will override it? If yes, we should add -f aka --force flag to capture the intent.

specs/commandline/policy.md
yizha1: @shizhMSFT @JeyJeyGao any comments on Pritesh's suggestion? We drop the `--template` flag for now, only considering `import` and `export` first?

specs/commandline/policy.md
```diff
@@ -0,0 +1,164 @@
...
+
+The `notation policy` command provides a user-friendly way to manage trust policies. It allows users to import and export trust policies from/to a JSON file. To get started user can use following sample trust policy. In this sample, there are four policies configured for different requirements:
+
+- The Policy named "wabbit-networks-images" is for verifying images signed by Wabbit Networks and stored in two repositories `registry.acme-rockets.io/software/net-monitor` and `registry.acme-rockets.io/software/net-logger`.
+
+- Policy named "unsigned-image" is for skipping the verification on unsigned images stored in repository `registry.acme-rockets.io/software/unsigned/net-utils`.
+
+- Policy "allow-expired-images" is for logging instead of failing expired images stored in repository `"registry.acme-rockets.io/software/legacy/metrics`.
+  
+- Policy "global-policy-for-all-other-images" is for verifying any other images that signed by the ACME Rockets.
```
priteshbandi, line 15: Can you please remove line-break between bullet points, this will improve rich mode rendering?

priteshbandi (approved):
LGTM with a nitpick

specs/commandline/policy.md
```diff
@@ -0,0 +1,162 @@
...
+### notation policy command
+
+```text
+Manage trust policies for signature verification.
+
+Usage:
+  notation policy [command]
+
+Available Commands:
+  export    export trust policies to a JSON file
```
patrickzheng200, line 82: I'm trying to understand the use case here. Are we trying to export the trustpolicy.json to another JSON file?

yizha1: Yes. The purpose for `import` and `export` commands is users don't need to know where the trust policies are stored for different OS, and what the name of the trust policy configuration file should be.

specs/commandline/policy.md
```diff
@@ -0,0 +1,162 @@
...
+
+The `notation policy` command provides a user-friendly way to manage trust policies. It allows users to import and export trust policies from/to a JSON file. To get started user can use following sample trust policy. In this sample, there are four policies configured for different requirements:
+
+- The Policy named "wabbit-networks-images" is for verifying images signed by Wabbit Networks and stored in two repositories `registry.acme-rockets.io/software/net-monitor` and `registry.acme-rockets.io/software/net-logger`.
+
+- Policy named "unsigned-image" is for skipping the verification on unsigned images stored in repository `registry.acme-rockets.io/software/unsigned/net-utils`.
+- Policy "allow-expired-images" is for logging instead of failing expired images stored in repository `"registry.acme-rockets.io/software/legacy/metrics`.
```
shizhMSFT, line 12: nit: typo
```suggestion
- Policy "allow-expired-images" is for logging instead of failing expired images stored in repository `registry.acme-rockets.io/software/legacy/metrics`.
```

specs/commandline/policy.md
```diff
@@ -0,0 +1,162 @@
...
+
+```text
+Export trust policies to a JSON file
+
+Usage:
+  notation policy export [flags] <file_path>
+
+Flags:
+  -d, --debug     debug mode
+      --force     override the supplied file, never prompt
```
shizhMSFT, line 100: What's the design purpose of `--force` for export?

specs/commandline/policy.md
priteshbandi: Also, user can generate and test trust policy locally and export it, then import it in there CI/CD or production environment.

specs/commandline/policy.md
```diff
@@ -0,0 +1,162 @@
...
+
+```text
+Import trust policies from a JSON file
+
+Usage:
+  notation policy import [flags] <file_path>
+
+Flags:
+  -d, --debug     debug mode
+      --force     override the existing policies, never prompt
```
shizhMSFT, line 115: What's the design purpose of `--force` for import?

specs/commandline/policy.md
```diff
@@ -0,0 +1,162 @@
...
+  -v, --verbose   verbose mode
+```
+
+### notation policy show
+
+```text
+Show trust policies
+
+Usage:
+  notation policy [flags] show
```
shizhMSFT, line 126: Does `notation policy show` prints the raw file or prettified JSON?

specs/commandline/policy.md
patrickzheng200: I see. Then what if there isn't an existing trustpolicy.json under the user's config dir, for example, when the user just downloaded the notation binary. How would the user get started in this case? IMO, a `notation policy init` command is required here.

specs/commandline/policy.md
```diff
@@ -0,0 +1,162 @@
...
+  -h, --help   help for policy
+```
+
+### notation policy export
+
+```text
+Export trust policies to a JSON file
+
+Usage:
+  notation policy export [flags] <file_path>
```
shizhMSFT, line 96: What's the difference between
```sh
notation policy export policy.json
```
and
```sh
notation policy show > policy.json
```

specs/commandline/policy.md
shizhMSFT: The policy file exists most time, doesn't it?

specs/commandline/policy.md
yizha1: This is the first iteration focusing on `import` and `export`. To get started, users can copy the template from the spec into a file. We can improve it on the next iteration.

specs/commandline/policy.md
yizha1: With `--force` flag, users don't need to confirm whether overriding the existing trust policies.

specs/commandline/policy.md
yizha1: With `--force` flag, users don't need to confirm whether overriding the existing file.

specs/commandline/policy.md
yizha1: IMS, the prettified JSON for `show` is better.

specs/commandline/policy.md
yizha1: This is a good point. Users can redirect the output of `show` command to a file.

specs/commandline/policy.md
patrickzheng200: > 

I'm not sure, "users can copy the template from the spec into a file" sounds not a very good user experience because in this way, the user still needs to know where they should put this file OR still call `notation policy import` as an extra command. IMO, we should include this `notation policy init` method in phase-1, once this command is called, a default trust policy should be generated and put at the correct place for the user automatically.

yizha1: The problem is that when we generate the default policy, users still need to export it, update it, and then import it. It will take more time to get an aligned UX, so the template will be put to next iteration.

patrickzheng200: > The problem is that when we generate the default policy, users still need to export it, update it, and then import it. It will take more time to get an aligned UX, so the template will be put to next iteration.

Then the question becomes can we generate a good enough default policy that satisfy most beginners/new users? For example, a default trust policy with all levels `strict` as global registry scope. Because for a 0-experience starter, going to the specs to find a trust policy template is very hard. A workaround is that we explicitly print out a template for users to use when notation detects the trustpolicy.json is missing.
Btw, correct me if I'm wrong, phase-1 targets RC.4, and phase-2 targets later than RC.4 right? Will we have both phases ready before the Kubecon?

yizha1: Let's find a time to discuss the default policy to see whether we can make it by RC-4.

yizha1:
@shizhMSFT @priteshbandi @priteshbandi your comments are resolved, and I also marked `notation policy` as a command in `Preview` status for RC-4.

specs/commandline/policy.md
```diff
@@ -0,0 +1,232 @@
...
+Initialize trust policies with default values
+
+Usage:
+  notation policy init [flags]
+
+Flags:
+  -d, --debug             debug mode
+      --force             override the existing trust polices, never prompt
+  -h, --help              help for export
+      --ts stringArray    specify the trust stores in format "<type>:<name>", e.g. "ca:my_store"
```
patrickzheng200, line 116: Should we use the full name here? something like:
```suggestion
      --trust-store stringArray    specify the trust stores in format "<type>:<name>", e.g. "ca:my_store"
```

specs/commandline/policy.md
```diff
@@ -0,0 +1,232 @@
...
+
+```text
+Initialize trust policies with default values
+
+Usage:
+  notation policy init [flags]
+
+Flags:
+  -d, --debug             debug mode
+      --force             override the existing trust polices, never prompt
```
patrickzheng200, line 114: Should a user even be allowed to use `notation policy init` with a `--force` flag? If there's already an existing trust policy file, `notation policy init` could just return an error. 
Using `--force` with `init` command sounds like a "restore to factory setting" behavior to me. If we do want to keep it, then its description should reflect this, how about:
```suggestion
      --force             restore the trust policy document to default. Any existing trust polices will be discarded, never prompt
```

specs/commandline/policy.md
```diff
@@ -0,0 +1,232 @@
...
+  -v, --verbose           verbose mode
+```
+
+### notation policy show
+
+```text
+Show trust policies
+
+Usage:
+  notation policy [flags] show
```
patrickzheng200, line 126: nit:
```suggestion
  notation policy show [flags]
```

specs/commandline/policy.md
```diff
@@ -0,0 +1,232 @@
...
+
+## Usage
+
+### Initialize trust policies with default values
+
+```shell
+notation policy init
+```
+
+Upon successful execution, trust policies with default values are created as following. Make sure to update the `trustStores` property with correct values.
```
patrickzheng200, line 142: I'm trying to understand the logic here. In this case, the user chooses to ignore the `--trust-store` flag, thus, the default `ca:default` trust store is used. If the user decided to go for the default, do they still need to "Make sure to update the `trustStores` property with correct values." ? Ideally, they don't need to modify the trust policy statement to get start.
```suggestion
Upon successful execution, trust policies with default values are created as following. Since the `--trust-store` flag is not specified, a default trust store `ca:default` is used.
```
Btw, we should print out the trust policy statement generated by `notation policy init` command.

specs/commandline/policy.md
```diff
@@ -0,0 +1,232 @@
...
+
+Use `--force` flag to override existing policies without prompt.
+
+### Initialize trust policies with specified trust stores
+
+```shell
+notation policy init --ts "ca:my_store" --ts "ca:my_store_2"
+```
+
+Upon successful execution, trust policies with default values and specified trust stores are created as following:
```
patrickzheng200, line 170: Again, we should print out the generated trust policy statement as the command's output.

specs/commandline/policy.md
yizha1: The default trust store `ca:default` may not exist, so need to inform users to create one or use other trust stores.

specs/commandline/policy.md
```diff
@@ -0,0 +1,232 @@
...
+Initialize trust policies with default values
+
+Usage:
+  notation policy init [flags]
+
+Flags:
+  -d, --debug                      debug mode
+      --force                      restore the trust policies to default values. Any existing trust polices will be discarded, never prompt
+  -h, --help                       help for export
+      --trust-store stringArray    specify the trust stores in format "<type>:<name>", e.g. "ca:my_store". If this flag is ignored, a default trust store "ca:default" is used
```
shizhMSFT, line 116: nit:
```suggestion
      --trust-store <type>:<name>  specify the trust stores in format "<type>:<name>", e.g. "ca:my_store". If this flag is ignored, a default trust store "ca:default" is used
```

specs/commandline/policy.md
```diff
@@ -0,0 +1,232 @@
...
+}
+
+```
+
+If there are existing trust policies configured and users still run `notation policy init` command, A prompt should be displayed asking for confirmation on whether restoring to default values or not. Use `--force` flag to discard any existing trust policies without prompt.
+
+### Initialize trust policies with specified trust stores
+
+```shell
+notation policy init --ts "ca:my_store" --ts "ca:my_store_2"
```
shizhMSFT, line 167: `--ts` is not specified or supported. Full flag name `--trust-store` should be used.

specs/commandline/policy.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Import trust policies from a JSON file
+
+```shell  
+notation policy import ./my_policy.json
+```
+
+The trust policies in the JSON file will be validated according to [trust policy properties](https://github.com/notaryproject/notaryproject/blob/v1.0.0-rc.2/specs/trust-store-trust-policy.md#trust-policy-properties). A successful message should be printed out if trust policies are imported successfully. Error logs including the reason should be printed out if the importing fails.
+
+Use `--force` flag to override existing policies without prompt.
```
shizhMSFT, line 215: Question: What if the imported policy file is invalid? Will the `--force` flag let it import?

specs/commandline/policy.md
```diff
@@ -0,0 +1,232 @@
...
+
+   ```shell
+   notation policy show > ./trust_policy.json
+   ```
+
+2. Edit the exported JSON file "trust_policy.json", update trust policies and save the file.
+3. Import trust policies from the file.
+
+   ```shell
+   notation policy import ./trust_policy.json
```
shizhMSFT, line 231: This command will fail since the policy file exists.

specs/commandline/policy.md
patrickzheng200: > The default trust store `ca:default` may not exist, so need to inform users to create one or use other trust stores.

Actually, that's not true. After the user called `notation policy init`, the next step would be adding their certificates into the trust store, for example, `notation cert add -t ca -s default ./cert1.crt ./cert2.crt`, this command will automatically create directory `truststore/x509/ca/default` and add cert1.crt and cert2.crt into the trust store. (Since the default trust policy is printed out as the output of `notation policy init` command, the user would explicitly see `ca:default` as their trust store.)

specs/commandline/policy.md
```diff
@@ -0,0 +1,232 @@
+# notation policy
+
+## Description
+
+As part of signature verification workflow, users need to configure the trust policies to specify trusted identities that signed the artifacts, and the level of signature verification to use. For more details, see [trust policy specification and examples](https://github.com/notaryproject/notaryproject/blob/v1.0.0-rc.2/specs/trust-store-trust-policy.md#trust-policy).
+
+The `notation policy` command provides a user-friendly way to manage trust policies. It allows users to initialize trust policies with default values, import trust policies from a JSON file, and show trust policies. To get started user can use following sample trust policy. In this sample, there are four policies configured for different requirements:
```
patrickzheng200, line 7: I think we should update this part since the first step is no longer manually copying the examples. Instead, `notation policy init` is used.
```suggestion
The `notation policy` command provides a user-friendly way to manage trust policies. It allows users to initialize trust policies with default values, import trust policies from a JSON file, and show trust policies. To get started user can call `notaiton policy init` command to generate a default trust policy.
```

specs/commandline/policy.md
```diff
@@ -0,0 +1,232 @@
...
+            },
+            "trustStores": [ 
+              "ca:acme-rockets"
+            ],                  
+            "trustedIdentities": [                                    
+                "x509.subject: C=US, ST=WA, L=Seattle, O=acme-rockets.io, CN=SecureBuilder"
+            ]
+        }
+    ]
+}
```
patrickzheng200, line 67: We should move these examples under the `update` session. Only users who don't like the default trust policy should consider updating it. Hence, they will need some examples at that point.

specs/commandline/policy.md
```diff
@@ -0,0 +1,232 @@
...
+  -d, --debug     debug mode
+      --force     override the existing trust policies, never prompt
+  -h, --help      help for import
+  -v, --verbose   verbose mode
+```
+
+### notation policy init
+
+```text
+Initialize trust policies with default values
```
patrickzheng200, line 107: We should mention that this command is the RECOMMENDED first step for a user who wants to verify a signature using Notation. How about:
```suggestion
Initialize trust policies with default values. This command serves as the entry point of Notation trust policy configuration.
```

specs/commandline/policy.md
```diff
@@ -0,0 +1,232 @@
...
+
+```shell  
+notation policy import ./my_policy.json
+```
+
+The trust policies in the JSON file will be validated according to [trust policy properties](https://github.com/notaryproject/notaryproject/blob/v1.0.0-rc.2/specs/trust-store-trust-policy.md#trust-policy-properties). A successful message should be printed out if trust policies are imported successfully. Error logs including the reason should be printed out if the importing fails.
+
+Use `--force` flag to override existing policies without prompt.
+
+### Update trust policies
```
patrickzheng200, line 217: As mentioned above, we should move those examples under this session.

specs/commandline/policy.md
```diff
@@ -0,0 +1,232 @@
...
+  -v, --verbose   verbose mode
+```
+
+### notation policy init
+
+```text
+Initialize trust policies with default values
+
+Usage:
+  notation policy init [flags]
```
priteshbandi, line 110: The issue with init command is that users will just use it without knowing its implication. 
In init command we are making following assumptions
*  `"registryScopes": ["*"]`  Looks fine 
*  `"signatureVerification": {"level": "strict"}`  If user are not aware of this then there verification might start failing in production because of revoked/expired signature. IMO we shouldn't default here.
*  `"trustStores": ["ca:default"]` In most of the usecases this value will be different as user will create there own trust-store. IMO we shouldn't default here.
*  `"trustedIdentities": ["*"]`  <-- If users are using publicly trusted certificates then it gives false sense of security. IMO we shouldn't default here.

Given that we can safely default only for 1 option out of 3, IMO we shouldn't include init option.

yizha1: @shizhMSFT @patrickzheng200 do you have any comments on Pritesh's comments? There are some security concerns here. One way to improve it is to introduce more flags for users to specify the value for each property, but it is the `notation policy add` command we want to do initially, but considering the scope and the time of release, it's better to do it in an iterate way. Let's agree on the way forward and move on.

shizhMSFT: @priteshbandi The purpose of the `notation policy init` is to improve the overall out of the box experience (OOBE) of the `notation` CLI. If an advanced customer does not think it is secure, that customer can compose their own policy file and use `notation policy import` to import it.

shizhMSFT: Again, writing a valid trust policy is not an easy job for a new user. They need to learn from scratch.

patrickzheng200: For a 0-experience starter, find our specs -> find a trust policy template -> learn how to use it -> change it to user wanted values could be a very hard workflow. The learning cost for a beginner is a bit too high here. The concern is that some new users just want to give it a try, but found it's hard to use, then they might just give up at this point.

`notation policy init` serves as an entry point for notation's trust policy configuration. We should print out the generated default trust policy statement for users to evaluate if it's good enough. For users who don't like this policy or think this policy is too naive, then they can use the default policy as the template (so they don't need to do the search on their own), then use `notation policy import` command to overwrite the default.

yizha1: Here is my proposal, please be aware that the `notation policy` command is new and marked as `preview` for RC-4 release. We can collect feedback from users and continuously improve it. 
iteration 1: introduce `notation policy import` and `notation policy show` commands （RC-4 and 1.0 experience)
iteration 2: introduce `notation policy init` for OOBE experience (post 1.0)
iteration 3: introduce `notation policy` CRUD commands for fine tuning (post 1.0)

@priteshbandi @patrickzheng200 @shizhMSFT Could you vote? Then I can start updating the spec.

shizhMSFT: I'd suggest including `notation policy init` in `1.0`.

yizha1: @shizhMSFT Let's get `iteration 1` done in RC-4, and in parallel, create a new issue to discuss the experience for `notation policy init`. I will create one for tracking.

patrickzheng200: > Here is the proposal, please be aware that the `notation policy` command is new and marked as `preview` for RC-4 release. We can collect feedback from users and continuously improve it. iteration 1: introduce `notation policy import` and `notation policy show` commands （RC-4 and 1.0 experience) iteration 2: introduce `notation policy init` for OOBE experience (post 1.0) iteration 3: introduce `notation policy` CRUD commands for fine tuning (post 1.0)
> 
> @priteshbandi @patrickzheng200 @shizhMSFT Could you vote? Then I can start updating the spec.

+1 for having `notaiton policy init` in `1.0`.

priteshbandi: > @priteshbandi The purpose of the notation policy init is to improve the overall out of the box experience (OOBE) of the notation CLI. If an advanced customer does not think it is secure, that customer can compose their own policy file and use notation policy import to import it.

How are we going to communicate that the trust policy can be insecure and unsage? As an end user I would assume the out of box defaults for a security products should be safe and secure.

> For a 0-experience starter, find our specs -> find a trust policy template -> learn how to use it -> change it to user wanted values could be a very hard workflow. The learning cost for a beginner is a bit too high here. The concern is that some new users just want to give it a try, but found it's hard to use, then they might just give up at this point.
notation policy init serves as an entry point for notation's trust policy configuration. We should print out the generated default trust policy statement for users to evaluate if it's good enough. For users who don't like this policy or think it's too basic, then they can use the default policy as the template (so they don't need to do the search on their own), then use notation policy import command to overwrite the default.

How will we communicate to user that they need to look beyond init? I am worried that if something things works, user might not look into its implications. Should we have command `notation policy generate-sample` which will print out policy, if user wants they can pipe it to notation import ?

I do recognize the need of out of the box experience but being security product we should be cautious to not provide insecure and unsafe defaults.

yizha1: @priteshbandi @shizhMSFT @patrickzheng200 Thanks for the great discussions. I suggest we implement `import` and `show` commands in RC-4, and continue the discussion of OOBE tracked at issue #590.

specs/commandline/policy.md
yizha1: The import will fail if the JSON file is invalid. it should be validated before overwriting the existing one.

specs/commandline/policy.md
yizha1: Need to prompt users whether to overwrite existing trust policies.

specs/commandline/policy.md
vaninrao10: > * ost of the usecases this value will be different as user will create there own trust-store. IMO we shouldn't default here.

+1. I think @priteshbandi is right here, do we have customer data to support the defaults ?

specs/commandline/policy.md
```diff
@@ -0,0 +1,164 @@
...
+
+```text
+Import trust policies from a JSON file
+
+Usage:
+  notation policy import [flags] <file_path>
+
+Flags:
+  -d, --debug     debug mode
+      --force     override the existing trust policies, never prompt
```
priteshbandi, line 98: If the trust policy exits and user tries uses import command; the command should fail
```suggestion
      --force     override the existing trust policies
```

priteshbandi (commented):
LGTM with a nit

priteshbandi (dismissed):

specs/commandline/policy.md
```diff
@@ -0,0 +1,164 @@
...
+  show      show trust policies
+
+Flags:
+  -h, --help   help for policy
+```
+
+### notation policy import
+
+```text
+Import trust policies from a JSON file
```
patrickzheng200, line 91: Just to confirm, what content in the JSON file does `notation policy import` expect? Is it the whole trust policy document file OR a single trust policy statement OR we accept both?
Maybe we should make it more clear in the description here.

specs/commandline/policy.md
patrickzheng200, line 98: If we only accept a whole trust policy document as the input JSON file, then `--force` flag should be used very carefully:
```suggestion
      --force     override the existing trust policy document file, any existing trust policies are discarded
```

specs/commandline/policy.md
```diff
@@ -0,0 +1,164 @@
...
+  -d, --debug     debug mode
+      --force     override the existing trust policies, never prompt
+  -h, --help      help for import
+  -v, --verbose   verbose mode
+```
+
+### notation policy show
+
+```text
+Show trust policies
```
patrickzheng200, line 106: similarly, what is the output of `notation policy show` command? Is it the whole trust policy document or just a list of trust policy statements. If the `import` command requires a whole document, then the `show` command should output a whole document accordingly.
Maybe we should make it more clear in the description here.

specs/commandline/policy.md
yizha1: The whole trust policy document. I will improve the wording.

specs/commandline/policy.md
yizha1: Here is the logic:
- If the trust policy exists, and users try to import without `--force`, the default behavior is to popup a prompt asking users whether overwrite existing one or not.
- Importing with `--force` flag, the existing trust policy will be overwritten. 

if users cannot overwrite it, users cannot update trust policies any more

yizha1: The suggested change doesn't indicate there is no prompt

specs/commandline/policy.md
```diff
@@ -0,0 +1,164 @@
...
+
+```text
+Import trust policy configuration from a JSON file
+
+Usage:
+  notation policy import [flags] <file_path>
+
+Flags:
+  -d, --debug     debug mode
+      --force     override the existing trust policy configuration, never prompt
```
patrickzheng200, line 98: On a second thought, do we even want to keep this `--force` flag? Should we always prompt if there is an existing trust policy configuration, and let user choose whether to proceed or not?

specs/commandline/policy.md
```diff
@@ -0,0 +1,164 @@
...
+
+An example of import trust policy configuration from a JSON file:
+
+```shell  
+notation policy import ./my_policy.json
+```
+
+The trust policy configuration in the JSON file should be validated according to [trust policy properties](https://github.com/notaryproject/notaryproject/blob/v1.0.0-rc.2/specs/trust-store-trust-policy.md#trust-policy-properties). A successful message should be printed out if trust policy configuration are imported successfully. Error logs including the reason should be printed out if the importing fails.
+
+If there is an existing trust policy configuration, prompt for users to confirm whether discarding existing configuration or not. Users can use `--force` flag to discard existing trust policy configuration without prompt.
```
patrickzheng200, line 129: Same here, what is the benefit of `Users can use --force flag to discard existing trust policy configuration without prompt.` ? IMO, to be safe, we should always let the user choose whether to proceed or not when there is already an existing trust policy configuration.

specs/commandline/policy.md
yizha1: It makes sense. @priteshbandi  any thoughts from you?

priteshbandi: Scripting around CLI prompts is non-intuitive, user will either need to pipe characters or use other tools/scrits like `expect`. IMO we should try to reduce usage of prompt and if we use prompt we support their bypass.

patrickzheng200: > Scripting around CLI prompts is non-intuitive, user will either need to pipe characters or use other tools/scripts like `expect`. IMO we should try to reduce usage of prompt and if we use prompt we support their bypass.

I see. Make sense to me.

patrickzheng200 (approved):
LGTM

priteshbandi (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #573, opened 2023-02-28
update: changed Sign to use OCI image manifest as default

### LINKED ISSUE #574: Change the default signature manifest

Notation CLI

Currently only a few registries support artifact manifests, but all the registries support image manifests. Notation CLI stores signatures using artifact manifest by default, so there is a portability issue when copying signatures from registries with artifact manifest support to registries do not support it. The experience is broken. Since rc.2, a flag named `--signature-manifest` was introduced for users to select the default manifest, but users need to know what manifest type that registries support. 

The proposal is for Notation CLI to store signatures using image manifest by default. The UX is improved since It works for most registries without additional configuration, and users don't need to learn what manifest type supported by the target registry.

No alternatives.

### Any additional context?

_No response_

patrickzheng200:
This PR resolves #574 .

### DIFF
--- a/cmd/notation/sign.go
+++ b/cmd/notation/sign.go
@@ -46,7 +46,7 @@ func signCommand(opts *signOpts) *cobra.Command {
 
 Prerequisite: a signing key needs to be configured using the command "notation key".
 
-Example - Sign an OCI artifact using the default signing key, with the default JWS envelope:
+Example - Sign an OCI artifact using the default signing key, with the default JWS envelope, and use OCI image manifest to store the signature:
   notation sign <registry>/<repository>@<digest>
 
 Example - Sign an OCI artifact using the default signing key, with the COSE envelope:
@@ -61,8 +61,8 @@ Example - Sign an OCI artifact identified by a tag (Notation will resolve tag to
 Example - Sign an OCI artifact stored in a registry and specify the signature expiry duration, for example 24 hours
   notation sign --expiry 24h <registry>/<repository>@<digest>
 
-Example - Sign an OCI artifact and use OCI image manifest to store the signature, with the default JWS envelope:
-  notation sign --signature-manifest image <registry>/<repository>@<digest>
+Example - Sign an OCI artifact and use OCI artifact manifest to store the signature:
+  notation sign --signature-manifest artifact <registry>/<repository>@<digest>
 `,
 		Args: func(cmd *cobra.Command, args []string) error {
 			if len(args) == 0 {
@@ -84,7 +84,7 @@ Example - Sign an OCI artifact and use OCI image manifest to store the signature
 	opts.SecureFlagOpts.ApplyFlags(command.Flags())
 	cmd.SetPflagExpiry(command.Flags(), &opts.expiry)
 	cmd.SetPflagPluginConfig(command.Flags(), &opts.pluginConfig)
-	command.Flags().StringVar(&opts.signatureManifest, "signature-manifest", signatureManifestArtifact, "manifest type for signature. options: \"artifact\", \"image\"")
+	command.Flags().StringVar(&opts.signatureManifest, "signature-manifest", signatureManifestImage, "manifest type for signature. options: \"image\", \"artifact\"")
 	cmd.SetPflagUserMetadata(command.Flags(), &opts.userMetadata, cmd.PflagUserMetadataSignUsage)
 	return command
 }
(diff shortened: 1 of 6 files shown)


specs/commandline/sign.md
```diff
@@ -154,10 +152,10 @@ Warning: Always sign the artifact using digest(`@sha256:...`) rather than a tag(
 Successfully signed localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
 ```
 
-### Sign an artifact and store the signature using OCI image manifest
+### Sign an artifact and store the signature using OCI artifact manifest
```
yizha1, line 155: I think we can remove this usage.

cmd/notation/sign.go
```diff
@@ -61,8 +61,8 @@ Example - Sign an OCI artifact identified by a tag (Notation will resolve tag to
 Example - Sign an OCI artifact stored in a registry and specify the signature expiry duration, for example 24 hours
   notation sign --expiry 24h <registry>/<repository>@<digest>
 
-Example - Sign an OCI artifact and use OCI image manifest to store the signature, with the default JWS envelope:
-  notation sign --signature-manifest image <registry>/<repository>@<digest>
+Example - Sign an OCI artifact and use OCI artifact manifest to store the signature:
+  notation sign --signature-manifest artifact <registry>/<repository>@<digest>
```
yizha1, line 65: I suggest we remove this example

specs/commandline/sign.md
shizhMSFT: I think we can keep it for now and evaluate later since we still support this case.

specs/commandline/sign.md
```diff
@@ -36,25 +36,23 @@ Flags:
...
+       --signature-manifest string  manifest type for signature, options: "image", "artifact" (default "image")
   -u,  --username string            username for registry operations (default to $NOTATION_USERNAME if not specified)
   -m,  --user-metadata stringArray  {key}={value} pairs that are added to the signature payload
   -v,  --verbose                    verbose mode
 ```
 
 ## Use OCI image manifest to store signatures
 
-By default, Notation uses [OCI artifact manifest][oci-artifact-manifest] to store signatures in registries. For registries that don't support `OCI artifact` or [Referrers API][oci-referers-api] is not enabled, users SHOULD use flag `--signature-manifest image` to force Notation to store the signatures using [OCI image manifest][oci-image-spec].
+By default, Notation uses [OCI image manifest][oci-image-spec] to store signatures in registries. Users could instead choose [OCI artifact manifest][oci-artifact-manifest] by enabling the `--signature-manifest artifact` flag. In this case, the registry is REQUIRED to support `OCI artifact` and with [Referrers API][oci-referers-api] enabled.
```
priteshbandi, line 47:
```suggestion
By default, Notation uses [OCI image manifest][oci-image-spec] to store signatures in registries. Users can use [OCI artifact manifest][oci-artifact-manifest] by enabling the `--signature-manifest artifact` flag. When using OCI artifact manifest to store the signature, the registry is REQUIRED to support both `OCI artifact` and [Referrers API][oci-referers-api].
```

specs/commandline/sign.md
```diff
@@ -36,25 +36,23 @@ Flags:
...
 Note that there is no deterministic way to determine whether a registry supports `OCI artifact` or not. The following response status contained in error messages MAY indicate that the registry doesn't support `OCI artifact`.
 
 - Response status `400 BAD Request` with error code `MANIFEST_INVALID` or `UNSUPPORTED`
 
 ### Set config property for OCI image manifest
 
-OCI image manifest requires additional property `config` of type `descriptor`, which is not required by OCI artifact manifest. Notation creates a default config descriptor for the user if flag `--signature-manifest image` is used.
-
-Notation uses empty JSON object `{}` as the default configuration content, and thus the default `config` property is fixed, as following:
+OCI image manifest requires additional property `config` of type `descriptor`, which is not required by OCI artifact manifest. On default signing with OCI image manifest, Notation uses empty JSON object `{}` as the default configuration content, and thus the `config` property is fixed, as following:
```
priteshbandi, line 55:
```suggestion
OCI image manifest requires additional property `config` of type `descriptor`, which is not required by OCI artifact manifest. When signing with OCI image manifest, Notation uses empty JSON object `{}` as the default configuration content, and thus the `config` property is fixed, as following:
```

cmd/notation/sign.go
yizha1: Agree to keep the example here

JeyJeyGao (approved):
LGTM

shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

Issue #584, opened 2023-03-07
Add an example to CLI help info for notation signing


FeynmanZhou:
Hi @kody-kimberl @priteshbandi @JeyJeyGao @patrickzheng200  ,

It seems the CLI example was not added to the [CLI help info](https://github.com/notaryproject/notation/blob/main/cmd/notation/sign.go#L49). Just a kindly reminder, when we change CLI commands and flags, we need to add or update the CLI examples in [notation/cmd](https://github.com/notaryproject/notation/blob/main/cmd/notation/) accordingly. 

I will raise another PR to add it.

_Originally posted by @FeynmanZhou in https://github.com/notaryproject/notation/issues/579#issuecomment-1457876374_

PR #585, opened 2023-03-07
doc: add an example to CLI help info for notation sign

FeynmanZhou:
Fix https://github.com/notaryproject/notation/issues/584

### DIFF
--- a/cmd/notation/sign.go
+++ b/cmd/notation/sign.go
@@ -51,13 +51,16 @@ func signCommand(opts *signOpts) *cobra.Command {
 		Short: "Sign artifacts",
 		Long: `Sign artifacts
 
-Prerequisite: a signing key needs to be configured using the command "notation key".
+Note: a signing key must be specified. This can be done temporarily by specifying a key ID, or a new key can be configured using the command "notation key add"
 
 Example - Sign an OCI artifact using the default signing key, with the default JWS envelope, and use OCI image manifest to store the signature:
   notation sign <registry>/<repository>@<digest>
 
 Example - Sign an OCI artifact using the default signing key, with the COSE envelope:
   notation sign --signature-format cose <registry>/<repository>@<digest> 
+
+Example - Sign an OCI artifact with a specified plugin and signing key stored in KMS 
+  notation sign --plugin <plugin_name> --id <remote_key_id> <registry>/<repository>@<digest>
 
 Example - Sign an OCI artifact using a specified key
   notation sign --key <key_name> <registry>/<repository>@<digest>


cmd/notation/sign.go
```diff
@@ -52,6 +52,9 @@ Example - Sign an OCI artifact using the default signing key, with the default J
 Example - Sign an OCI artifact using the default signing key, with the COSE envelope:
   notation sign --signature-format cose <registry>/<repository>@<digest> 
 
+Example - Sign an OCI artifact with a specified plugin and on-demand remote key
+  notation sign --plugin <plugin_name> --id <remote_key_id> <registry>/<repository>@<digest>
+
```
kody-kimberl, line 57: Should this go above the Prerequisite since it doesn't require a signing key to be configured first?

FeynmanZhou: @kody-kimberl Good catch. I removed the prerequisite from the example section. It makes the examples look neat. We can explain this prerequisite in Notation documentation.

kody-kimberl (dismissed):
LGTM

cmd/notation/sign.go
```diff
@@ -47,14 +47,16 @@ func signCommand(opts *signOpts) *cobra.Command {
 		Short: "Sign artifacts",
 		Long: `Sign artifacts
 
-Prerequisite: a signing key needs to be configured using the command "notation key".
```
patrickzheng200, line 50: We recommend users to use `notation key` to configure keys before signing. In other words, the original Signing process should still be the recommended way for Signing. Should we rephrase this line to reflect this?

FeynmanZhou: @patrickzheng200  Good suggestion. I rephrased this note to `Note: a signing key needs to be configured using the command "notation key" or with specifying a key ID`.

cmd/notation/sign.go
```diff
@@ -47,14 +47,17 @@ func signCommand(opts *signOpts) *cobra.Command {
...
+Note: a signing key needs to be configured using the command "notation key" or with specifying a key ID
```
patrickzheng200, line 50: How about:
```suggestion
Recommend: configure signing key using the command "notation key add".
```

FeynmanZhou: @patrickzheng200 I would suggest `Note: a signing key needs to be configured using the command "notation key add" or with specifying a key ID`. There is an example to sign with a specified key ID. It doesn't require `notation key add`.

kody-kimberl: I'm not sure if this is the best phrasing, since using a key ID is not configuring a new key. What about `Note: a signing key must be specified. This can be done temporarily by specifying a key ID, or a new key can be configured using the command "notation key add"`

patrickzheng200: > I'm not sure if this is the best phrasing, since using a key ID is not configuring a new key. What about `Note: a signing key must be specified. This can be done temporarily by specifying a key ID, or a new key can be configured using the command "notation key add"`

Yeah, I think Kody's suggestion is a bit better here. @FeynmanZhou

FeynmanZhou: Agree with @kody-kimberl . Updated.

patrickzheng200 (approved):
LGTM

priteshbandi (approved):
LGTM

shizhMSFT (approved):
LGTM

patrickzheng200 (approved):
LGTM

priteshbandi (approved):
LGTM

PR #593, opened 2023-03-22
feat: add `show` and `import`  for trust policy management

### LINKED ISSUE #548: CLI Cmds for trust policy management - phase 1

Notation CLI

Today the UX of trust policy management is cumbersome and error prone. 
- Users need to manually create and edit a JSON file, and there are several parameters user need to properly configure. See [trust policy spec](https://github.com/notaryproject/notaryproject/blob/main/specs/trust-store-trust-policy.md).
- The name for the JSON file MUST be `trustpolicy.json`, and this file needs to be stored in Notation configuration directory, which is different for different OS, see [directory structure](https://notaryproject.dev/docs/how-to/directory-structure/). It's hard to remember the directory path for different OS.
- The JSON file encoding MUST be utf-8 without BOM. However, for some environments it is not by default. It is error prone.

To improve the UX of managing trust policy and meet Notation v1.0 release goal, a phased approach is proposed. This issue will address the solution for the first phase. The scope for first phase:
- `Import trust policy`: User can import a trust policy configuration file with arbitrary name into notation, so that users don't need to consider where this file should go to, and what is the name of trust policy file. 
- `Export trust policy`: User can export the trust policy configuration file to local disk, so that user can update it and import it afterwards.
- `View trust policy`: User can easily view the content of trust policy file.
- `template file`: Provide a template file of trust policy configuration for users to get started. The template file could be packaged with notation binary together.

No

### Any additional context?

_No response_

qweeah:
This PR implements what's described in [policy.md](https://github.com/notaryproject/notation/blob/18efc3de233ef5eb00e2f32009475a2b8a3eae47/specs/commandline/policy.md).

Resolves #548

### DIFF
--- a/cmd/notation/policy/import.go
+++ b/cmd/notation/policy/import.go
@@ -1,0 +1,83 @@
+package policy
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/notaryproject/notation-go/dir"
+	"github.com/notaryproject/notation-go/verifier/trustpolicy"
+	"github.com/notaryproject/notation/cmd/notation/internal/cmdutil"
+	"github.com/notaryproject/notation/internal/osutil"
+	"github.com/spf13/cobra"
+)
+
+type importOpts struct {
+	filePath string
+	force    bool
+}
+
+func importCmd() *cobra.Command {
+	var opts importOpts
+	command := &cobra.Command{
+		Use:   "import [flags] <file_path>",
+		Short: "[Preview] Import trust policy configuration from a JSON file",
+		Long: `[Preview] Import trust policy configuration from a JSON file.
+
+** This command is in preview and under development. **
+
+Example - Import trust policy configuration from a file:
+  notation policy import my_policy.json
+`,
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			opts.filePath = args[0]
+			return runImport(cmd, opts)
+		},
+	}
+	command.Flags().BoolVar(&opts.force, "force", false, "override the existing trust policy configuration, never prompt")
+	return command
+}
+
+func runImport(command *cobra.Command, opts importOpts) error {
+	// optional confirmation
+	if !opts.force {
+		if _, err := trustpolicy.LoadDocument(); err == nil {
+			confirmed, err := cmdutil.AskForConfirmation(os.Stdin, "Existing trust policy configuration found, do you want to overwrite it?", opts.force)
+			if err != nil {
+				return err
+			}
+			if !confirmed {
+				return nil
+			}
+		}
+	} else {
+		fmt.Fprintf(os.Stderr, "Warning: existing trust policy configuration file will be overwritten")
+	}
+
+	// read configuration
+	policyJSON, err := os.ReadFile(opts.filePath)
+	if err != nil {
+		return fmt.Errorf("failed to read trust policy file: %w", err)
+	}
+
+	// parse and validate
+	var doc trustpolicy.Document
+	if err = json.Unmarshal(policyJSON, &doc); err != nil {
+		return fmt.Errorf("failed to parse trust policy configuration: %w", err)
+	}
+	if err = doc.Validate(); err != nil {
+		return fmt.Errorf("failed to validate trust policy: %w", err)
+	}
+
+	// write
+	policyPath, err := dir.ConfigFS().SysPath(dir.PathTrustPolicy)
+	if err != nil {
+		return fmt.Errorf("failed to obtain path of trust policy file: %w", err)
+	}
+	if err = osutil.WriteFile(policyPath, policyJSON); err != nil {
+		return fmt.Errorf("failed to write trust policy file: %w", err)
+	}
+	_, err = fmt.Fprintln(os.Stdout, "Trust policy configuration imported successfully.")
+	return err
+}
--- a/cmd/notation/policy/show.go
+++ b/cmd/notation/policy/show.go
@@ -1,0 +1,64 @@
+package policy
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/notaryproject/notation-go/dir"
+	"github.com/notaryproject/notation-go/verifier/trustpolicy"
+	"github.com/spf13/cobra"
+)
+
+type showOpts struct {
+}
+
+func showCmd() *cobra.Command {
+	var opts showOpts
+	command := &cobra.Command{
+		Use:   "show [flags]",
+		Short: "[Preview] Show trust policy configuration",
+		Long: `[Preview] Show trust policy configuration.
+
+** This command is in preview and under development. **
+
+Example - Show current trust policy configuration:
+  notation policy show
+
+Example - Save current trust policy configuration to a file:
+  notation policy show > my_policy.json
+`,
+		Args: cobra.ExactArgs(0),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runShow(cmd, opts)
+		},
+	}
+	return command
+}
+
+func runShow(command *cobra.Command, opts showOpts) error {
+	// get policy file path
+	policyPath, err := dir.ConfigFS().SysPath(dir.PathTrustPolicy)
+	if err != nil {
+		return fmt.Errorf("failed to obtain path of trust policy configuration file: %w", err)
+	}
+
+	// core process
+	policyJSON, err := os.ReadFile(policyPath)
+	if err != nil {
+		return fmt.Errorf("failed to load trust policy configuration, you may import one via `notation policy import <path-to-policy.json>`: %w", err)
+	}
+	var doc trustpolicy.Document
+	if err = json.Unmarshal(policyJSON, &doc); err == nil {
+		err = doc.Validate()
+	}
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %s\n", err.Error())
+		fmt.Fprintf(os.Stderr, "Existing trust policy configuration is invalid, you may update or create a new one via `notation policy import <path-to-policy.json>`\n")
+		// not returning to show the invalid policy configuration
+	}
+
+	// show policy content
+	_, err = os.Stdout.Write(policyJSON)
+	return err
+}
--- a/specs/commandline/policy.md
+++ b/specs/commandline/policy.md
@@ -94,10 +94,8 @@ Usage:
   notation policy import [flags] <file_path>
 
 Flags:
-  -d, --debug     debug mode
       --force     override the existing trust policy configuration, never prompt
   -h, --help      help for import
-  -v, --verbose   verbose mode
 ```
 
 ### notation policy show
@@ -109,9 +107,7 @@ Usage:
   notation policy show [flags]
 
 Flags:
-  -d, --debug     debug mode
   -h, --help      help for show
-  -v, --verbose   verbose mode
 ```
 
 ## Usage
@@ -136,7 +132,7 @@ Use the following command to show trust policy configuration:
 notation policy show
 ```
 
-Upon successful execution, the trust policy configuration are printed out in a pretty JSON format. If trust policy is not configured, users should receive an error message, and a tip to import trust policy configuration from a JSON file.
+Upon successful execution, the trust policy configuration are printed out to standard output. If trust policy is not configured or is malformed, users should receive an error message via standard error output, and a tip to import trust policy configuration from a JSON file.
 
 ### Export trust policy configuration into a JSON file
 
--- a/cmd/notation/main.go
+++ b/cmd/notation/main.go
@@ -4,6 +4,7 @@ import (
 	"os"
 
 	"github.com/notaryproject/notation/cmd/notation/cert"
+	"github.com/notaryproject/notation/cmd/notation/policy"
 	"github.com/spf13/cobra"
 )
 
@@ -18,6 +19,7 @@ func main() {
 		verifyCommand(nil),
 		listCommand(nil),
 		cert.Cmd(),
+		policy.Cmd(),
 		keyCommand(),
 		pluginCommand(),
 		loginCommand(nil),
--- a/cmd/notation/policy/cmd.go
+++ b/cmd/notation/policy/cmd.go
@@ -1,0 +1,18 @@
+package policy
+
+import "github.com/spf13/cobra"
+
+func Cmd() *cobra.Command {
+	command := &cobra.Command{
+		Use:   "policy [command]",
+		Short: "[Preview] Manage trust policy configuration",
+		Long:  "[Preview] Manage trust policy configuration for signature verification.",
+	}
+
+	command.AddCommand(
+		showCmd(),
+		importCmd(),
+	)
+
+	return command
+}
--- a/test/e2e/suite/command/policy.go
+++ b/test/e2e/suite/command/policy.go
@@ -1,0 +1,188 @@
+package command
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+
+	. "github.com/notaryproject/notation/test/e2e/internal/notation"
+	"github.com/notaryproject/notation/test/e2e/internal/utils"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("trust policy maintainer", func() {
+	When("showing configuration", func() {
+		It("should show error and hint if policy doesn't exist", func() {
+			Host(Opts(), func(notation *utils.ExecOpts, artifact *Artifact, vhost *utils.VirtualHost) {
+				notation.ExpectFailure().
+					Exec("policy", "show").
+					MatchErrKeyWords("failed to load trust policy configuration", "notation policy import")
+			})
+		})
+
+		It("should show exist policy", func() {
+			content, err := os.ReadFile(filepath.Join(NotationE2ETrustPolicyDir, TrustPolicyName))
+			Expect(err).NotTo(HaveOccurred())
+			Host(Opts(AddTrustPolicyOption(TrustPolicyName)), func(notation *utils.ExecOpts, artifact *Artifact, vhost *utils.VirtualHost) {
+				notation.Exec("policy", "show").
+					MatchContent(string(content))
+			})
+		})
+
+		It("should display error hint when showing invalid policy", func() {
+			policyName := "invalid_format_trustpolicy.json"
+			content, err := os.ReadFile(filepath.Join(NotationE2ETrustPolicyDir, policyName))
+			Expect(err).NotTo(HaveOccurred())
+			Host(Opts(AddTrustPolicyOption(policyName)), func(notation *utils.ExecOpts, artifact *Artifact, vhost *utils.VirtualHost) {
+				notation.Exec("policy", "show").
+					MatchErrKeyWords("existing trust policy configuration is invalid").
+					MatchContent(string(content))
+			})
+		})
+	})
+
+	When("importing configuration without existing trust policy configuration", func() {
+		opts := Opts()
+		It("should fail if no file path is provided", func() {
+			Host(opts, func(notation *utils.ExecOpts, artifact *Artifact, vhost *utils.VirtualHost) {
+				notation.ExpectFailure().
+					Exec("policy", "import")
+			})
+		})
+
+		It("should fail if provided file doesn't exist", func() {
+			Host(opts, func(notation *utils.ExecOpts, artifact *Artifact, vhost *utils.VirtualHost) {
+				notation.ExpectFailure().
+					Exec("policy", "import", "/??/???")
+			})
+		})
+
+		It("should fail if registry scope is malformed", func() {
+			Host(opts, func(notation *utils.ExecOpts, artifact *Artifact, vhost *utils.VirtualHost) {
+				notation.ExpectFailure().
+					Exec("policy", "import", filepath.Join(NotationE2ETrustPolicyDir, "malformed_registry_scope_trustpolicy.json"))
+			})
+		})
+
+		It("should fail if store is malformed", func() {
+			Host(opts, func(notation *utils.ExecOpts, artifact *Artifact, vhost *utils.VirtualHost) {
+				notation.ExpectFailure().
+					Exec("policy", "import", filepath.Join(NotationE2ETrustPolicyDir, "malformed_trust_store_trustpolicy.json"))
+			})
+		})
+
+		It("should fail if identity is malformed", func() {
+			Host(opts, func(notation *utils.ExecOpts, artifact *Artifact, vhost *utils.VirtualHost) {
+				notation.ExpectFailure().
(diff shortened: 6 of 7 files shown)


cmd/notation/policy/show.go
```diff
@@ -0,0 +1,55 @@
...
+	var opts showOpts
+	command := &cobra.Command{
+		Use:   "show [flags]",
+		Short: "Show trust policy configuration",
+		Long: `Show trust policy configuration.
+Example - Show current trust policy configuration:
+  notation policy show
+
+Example - Save current trust policy configuration to a file:
+    notation policy show > my_policy.json
```
shizhMSFT, line 29: nit: inconsistent indentation

cmd/notation/policy/show.go
```diff
@@ -0,0 +1,55 @@
...
+		return fmt.Errorf("failed to read trust policy configuration: %w", err)
+	}
+
+	buf := bytes.NewBuffer(nil)
+	if err := json.Indent(buf, policyJson, "", "  "); err != nil {
+		return fmt.Errorf("failed to prettify: %w", err)
+	}
+	buf.WriteByte('\n')
+	_, err = os.Stdout.Write(buf.Bytes())
+	return err
```
shizhMSFT, line 54: We should dump the original json file directly.

cmd/notation/policy/import.go
```diff
@@ -0,0 +1,106 @@
...
+	var doc trustpolicy.Document
+	if err = json.Unmarshal(policyJson, &doc); err != nil {
+		return fmt.Errorf("failed to parse trust policy configuration: %w", err)
+	}
+	if err = doc.Validate(); err != nil {
+		return fmt.Errorf("failed to validate trust policy configuration: %w", err)
+	}
+
+	// write
+	if err = os.WriteFile(policyPath, policyJson, 0666); err != nil {
```
shizhMSFT, line 74: What's the expected permission? `0600` or `0644`?

qweeah: Let's do it this way:
1) if the original policy file exists, clone the file mode from it
2) otherwise set it to `0644`

yizha1:
Suggest marking the new policy command as `Preview` in the output of help command.

cmd/notation/policy/import.go
```diff
@@ -0,0 +1,107 @@
...
+	return command
+}
+
+func runImport(command *cobra.Command, opts importOpts) error {
+	// optional confirmation
+	var mode os.FileMode = 0644
+	policyPath := filepath.Join(dir.UserConfigDir, dir.PathTrustPolicy)
+	if fi, err := os.Stat(policyPath); err == nil {
+		mode = fi.Mode()
+		confirmed, err := opts.askForConfirmation(os.Stdin, "Existed trust policy configuration found, do you want to overwrite it?")
```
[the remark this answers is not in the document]
qweeah: Good catch! Resolved.

qweeah:
> Suggest marking the new policy command as `Preview` in the output of help command.

Added `Preview` feature label. @yizha1 Do notation have or need to have any doc on the classification of work-in-progress feature labels?

yizha1:
> > Suggest marking the new policy command as `Preview` in the output of help command.
> 
> Added `Preview` feature label. @yizha1 Do notation have or need to have any doc on the classification of work-in-progress feature labels?

No. We need this kind of doc.

cmd/notation/policy/show.go
```diff
@@ -0,0 +1,50 @@
...
+	return command
+}
+
+func runShow(command *cobra.Command, opts showOpts) error {
+	// read policy json contents
+	policyJson, err := os.ReadFile(filepath.Join(dir.UserConfigDir, dir.PathTrustPolicy))
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "no trust policy configuration found, you may import one via `notation policy import <path-to-policy.json>`")
+		return fmt.Errorf("failed to read trust policy configuration: %w", err)
+	}
```
patrickzheng200, line 47:
```suggestion
	policyDoc, err := trustpolicy.LoadDocument()
	if err != nil {
		fmt.Fprintln(os.Stderr, "failed to load trust policy configuration, you may import one via `notation policy import <path-to-policy.json>`")
		return err
	}
	if err := policyDoc.Validate(); err != nil {
		fmt.Fprintln(os.Stderr, "existing trust policy configuration is invalid, you may import one via `notation policy import <path-to-policy.json>`")
		return err
	}
```
we need to call `Validate()` according to our spec: https://github.com/notaryproject/notation/blob/main/specs/commandline/policy.md#usage, i.e., only valid trust policy document shall be returned to the user.

cmd/notation/policy/import.go
```diff
@@ -0,0 +1,110 @@
...
+	"strings"
+
+	"github.com/notaryproject/notation-go/dir"
+	"github.com/notaryproject/notation-go/verifier/trustpolicy"
+	"github.com/notaryproject/notation/internal/cmd"
+	"github.com/spf13/cobra"
+)
+
+type importOpts struct {
+	cmd.LoggingFlagOpts
```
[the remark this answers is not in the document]
qweeah: Adding this because `-d` and `-v` are [in the spec](https://github.com/notaryproject/notation/blob/main/specs/commandline/policy.md#notation-policy-show)

cmd/notation/policy/show.go
```diff
@@ -0,0 +1,50 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/notaryproject/notation-go/dir"
+	"github.com/notaryproject/notation/internal/cmd"
+	"github.com/spf13/cobra"
+)
+
+type showOpts struct {
+	cmd.LoggingFlagOpts
```
patrickzheng200, line 14: looks like this is not used?

cmd/notation/policy/show.go
qweeah: I do see the validation for `import` but doesn't find it in `show` @yizha1 Can you help confirm the UX?

cmd/notation/policy/import.go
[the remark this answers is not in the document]
patrickzheng200: if we are not logging anything, and there's no http round trips, we could just drop these flags from the spec.

cmd/notation/policy/show.go
patrickzheng200: If that's not in the spec, I think that's something we want to add into it. Because returning an invalid trust policy document to the user might be confusing and not useful.

qweeah: Yes, I agree that validation need to be done. But how to deal with invalid policy is a tricky work and the UX need to be clarified:

Option 1: show validation error only.
Option 2: show validation error and the current configuration.
Option 3: show validation error and the current configuration with invalid part highlighted.

patrickzheng200: IMO,
Option 3 might be the best option, it tells the user which exact part needs to be fixed.
Option 1 is also good enough, as it tells the user the existing trust policy is malformed, import a new one to replace it. (Note, we do not have a `notation policy init` command, this means that we can fairly assume at this point user knows all the required specs of a notation trust policy document.)
Option 2 is not what we want. Since it's an invalid configuration, merely showing it does not add any value to the user.

yizha1: > I do see the validation for `import` but doesn't find it in `show` @yizha1 Can you help confirm the UX?

Why we need validation for show command? Show command just displays the current trust policy configuration which should be valid

patrickzheng200: > > I do see the validation for `import` but doesn't find it in `show` @yizha1 Can you help confirm the UX?
> 
> Why we need validation for show command? Show command just displays the current trust policy configuration which should be valid

trustpolicy.json is not guaranteed to be valid at any point of time, since it's just a file, it can still be manually removed/added by the user without using notation's command. (Btw, that's one of the reasons that we should have a `notation policy init` command.) Think about a new user, they might just compose their own trustpolicy.json then drag and drop it into notation's directory without using `notation policy import`.

yizha1: I chose option 3, but the validation result only shows the invalid part not the whole configuration. @qweeah @patrickzheng200

patrickzheng200: > I chose option 3, but the validation result only shows the invalid part not the whole configuration. @qweeah @patrickzheng200

Yeah, agree. Our `validate()` process fails at the first error it encounters, so it won't show all the failures in the doc. In this case, we could just `print out the error + the certain line of content that triggered this error`.

cmd/notation/policy/import.go
```diff
@@ -0,0 +1,110 @@
...
+	opts.LoggingFlagOpts.ApplyFlags(command.Flags())
+	return command
+}
+
+func runImport(command *cobra.Command, opts importOpts) error {
+	// optional confirmation
+	var mode os.FileMode = 0644
+	policyPath := filepath.Join(dir.UserConfigDir, dir.PathTrustPolicy)
+	if fi, err := os.Stat(policyPath); err == nil {
+		mode = fi.Mode()
```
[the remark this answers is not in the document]
qweeah: `LoadDocument` seems too heavy for detecting whether a file exists or not, also the permission information is not viable on caller side.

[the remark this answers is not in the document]
qweeah: It would be great if notation-go have a `LoadDocumentInPath` which can be used to read in the to-be-imported policy file

cmd/notation/policy/show.go
qweeah: I had an implementation, but the output information may not be accurate, e.g. if the version number doesn't exist, the error message will still mention the first policy.  So to fully support option3, Changes need to be made to [Document.Validate](https://pkg.go.dev/github.com/notaryproject/notation-go@v1.0.0-rc.3/verifier/trustpolicy#Document.Validate) to provide more contextual info.

1. Version validation failure returns certain type of exported errors
2. Separate version and policy validation into two exported functions

cmd/notation/policy/import.go
```diff
@@ -0,0 +1,96 @@
...
+			tmp := doc
+			tmp.TrustPolicies = []trustpolicy.TrustPolicy{p}
+			if err = tmp.Validate(); err != nil {
+				logger.Errorf("failed to validate trust policy %q:\n", p.Name)
+				if content, err := json.Marshal(&p); err == nil {
+					logger.Errorf("\n%s\n", string(content))
+				}
+				return err
+			}
+		}
```
patrickzheng200, line 87: We could just return the err directly. The error messages of notation-go `Validate()` https://github.com/notaryproject/notation-go/blob/fb79e6df1aa8cacc8c9f4e917e36ec6dc0403f37/verifier/trustpolicy/trustpolicy.go#L157 ought to provide the details of the failure. If not, please file an issue in notation-go, and we need to improve those error messages there.
```suggestion
		return err
```

cmd/notation/policy/import.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	// write
+	if err = osutil.WriteFileWithPermission(policyPath, policyJson, mode, true); err != nil {
```
patrickzheng200, line 91: As mentioned above, could just use `WriteFile` here.

cmd/notation/policy/show.go
patrickzheng200: @qweeah Based on what I discussed with @yizha1, we could just show the error for now, since the error message contains which trust policy statement triggered the error on which field due to what reason. We could improve the logic to show all the errors in the file after RC.4.

cmd/notation/policy/import.go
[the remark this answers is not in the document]
qweeah: Will make imported config permission to 600.

But still it's wrong to use `LoadDocument()` here on the to-be-replaced config. Consider this scenario: if user have a malformed trust policy file `A` in the notation config path and want to import a valid policy `B` in. Use `LoadDocument()` here will 1) open `A`, (which is unnecessary since checking the existence of a file only requires open its containing folder); and 2) always fail because `A` is malformed. So the result is **user will not be able to import a valid policy since current policy is malformed**.

cmd/notation/policy/import.go
qweeah: Will change to minimal user access (https://github.com/notaryproject/notation/pull/453/files) as is suggested by @patrickzheng200

cmd/notation/policy/import.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	if err = doc.Validate(); err != nil {
+		return fmt.Errorf("failed to validate trust policy configuration: %w", err)
+	}
+
+	// write
+	if err = os.WriteFile(policyPath, policyJson, mode); err != nil {
+		return fmt.Errorf("failed to write trust policy configuration: %w", err)
+	}
+	_, err = fmt.Fprintln(os.Stdout, "Trust policy configuration imported successfully.")
```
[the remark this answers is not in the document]
qweeah: Shouldn't the the final trustpolicy.json be identical to what user has provided? Can you give a use case to explain why use need to view the content of the input?

cmd/notation/policy/show.go
qweeah: sure, implemented as suggested

cmd/notation/policy/import.go
qweeah: implemented as suggested

cmd/notation/policy/import.go
qweeah: changed accordingly

cmd/notation/policy/import.go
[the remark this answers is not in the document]
patrickzheng200: > Will make imported config permission to 600.
> 
> But still it's wrong to use `LoadDocument()` here on the to-be-replaced config. Consider this scenario: if user have a malformed trust policy file `A` in the notation config path and want to import a valid policy `B` in. Use `LoadDocument()` here will 1) open `A`, (which is unnecessary since checking the existence of a file only requires opennig its containing folder); 

As I mentioned above, we are not just checking the existence here. If the original trust policy cannot even pass `LoadDocument()` then we should allow user to override it directly. Because otherwise, the user might think there's already a working trust policy doc in the configuration and stop the import at confirmation.

> and 2) always fail because `A` is malformed. So the result is **user will not be able to import a valid policy since current policy is malformed**.

my code suggestion was: `if _, err := LoadDocument(); err == nil {`, if the original doc can pass `LoadDocument()`, then we need to prompt to ask for confirmation. If `err != nil`, we just go ahead and import the new policy.

cmd/notation/policy/import.go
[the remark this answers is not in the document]
patrickzheng200: > Shouldn't the the final trustpolicy.json be identical to what user has provided? Can you give a use case to explain why use need to view the content of the input?

Correct. It should be identical, but printing it out in the end gives user feedback that Notation interpreted the file correctly and did not manipulate the file during the process.

cmd/notation/policy/import.go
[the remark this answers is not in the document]
qweeah: > If the original trust policy cannot even pass LoadDocument() then we should allow user to override it directly

Sounds workable. Will change accordingly. Also the scope of `--force` need to be expanded and if it's set, policy loading will be bypassed.

cmd/notation/policy/import.go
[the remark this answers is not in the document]
qweeah: > Notation interpreted the file correctly and did not manipulate the file during the process.

`imported successfully` should done providing the feedback clearly.

cmd/notation/policy/import.go
```diff
@@ -0,0 +1,82 @@
...
+		},
+	}
+	opts.LoggingFlagOpts.ApplyFlags(command.Flags())
+	command.Flags().BoolVar(&opts.force, "force", false, "override the existing trust policy configuration, never prompt")
+	return command
+}
+
+func runImport(command *cobra.Command, opts importOpts) error {
+	// optional confirmation
+	policyPath := filepath.Join(dir.UserConfigDir, dir.PathTrustPolicy)
```
patrickzheng200, line 45: `policyPath` is used in line 74, we could define it right above its usage.
Also, we should use `dir.ConfigFS()`:
```suggestion
	policyPath, err := dir.ConfigFS().SysPath(dir.PathTrustPolicy)
```

cmd/notation/policy/import.go
```diff
@@ -0,0 +1,79 @@
...
+		return fmt.Errorf("failed to read trust policy configuration: %w", err)
+	}
+
+	// parse and validate
+	var doc trustpolicy.Document
+	if err = json.Unmarshal(policyJson, &doc); err != nil {
+		return fmt.Errorf("failed to parse trust policy configuration: %w", err)
+	}
+	if err := doc.Validate(); err != nil {
+		return fmt.Errorf("failed to validate trust policy:%w", err)
```
patrickzheng200, line 70: nit:
```suggestion
		return fmt.Errorf("failed to validate trust policy: %w", err)
```

cmd/notation/policy/show.go
```diff
@@ -0,0 +1,47 @@
...
+	return command
+}
+
+func runShow(command *cobra.Command, opts showOpts) error {
+	// read policy json contents
+	policyJson, err := os.ReadFile(filepath.Join(dir.UserConfigDir, dir.PathTrustPolicy))
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "no trust policy configuration found, you may import one via `notation policy import <path-to-policy.json>`")
+		return fmt.Errorf("failed to read trust policy configuration: %w", err)
+	}
```
patrickzheng200, line 44: I thought this is what we discussed to do?
```suggestion
	policyDoc, err := trustpolicy.LoadDocument()
	if err != nil {
		fmt.Fprintln(os.Stderr, "failed to load trust policy configuration, you may import one via `notation policy import <path-to-policy.json>`")
		return err
	}
	if err := policyDoc.Validate(); err != nil {
		fmt.Fprintln(os.Stderr, "existing trust policy configuration is invalid, you may import one via `notation policy import <path-to-policy.json>`")
		return err
	}
```

patrickzheng200: Here's one of the reasons that we need `Validate()` here:
Let's say the user wants to add a new trust policy statement to the existing trust policy document. If the existing trust policy doc is not valid, and `notation policy show` does not report it, then the user would just add their new statement into the original invalid document and call `notation policy import` to apply the change. And since `notation policy import` calls `Validate()` on the new document, it would not surprisingly fail. Only until this point, the user will realize the original policy doc is invalid and all their previous work needs to be redone. Thus, explicitly displaying the existing trust policy is malformed before the user make any change on it is necessary.

qweeah: `fmt.Fprintf` does not support error-wrapping directive `%w`, changed it to `%v`

patrickzheng200 (approved):
LGTM

patrickzheng200:
Lookes like the e2e has failed.

patrickzheng200 (approved):
LGTM

byronchien (dismissed):

specs/commandline/policy.md
qweeah, line 1: @yizha1 There are some UX changes reflected in this file, can you help confirm, thanks.

cmd/notation/policy/show.go
```diff
@@ -0,0 +1,47 @@
...
+	return command
+}
+
+func runShow(command *cobra.Command, opts showOpts) error {
+	// read policy json contents
+	policyJson, err := os.ReadFile(filepath.Join(dir.UserConfigDir, dir.PathTrustPolicy))
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "no trust policy configuration found, you may import one via `notation policy import <path-to-policy.json>`")
+		return fmt.Errorf("failed to read trust policy configuration: %w", err)
```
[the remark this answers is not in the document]
qweeah: > Is it possible that os.ReadFile throws error even when file exists?

Yes, it's possible. Are you suggesting to use [os.IsNotExist](https://pkg.go.dev/os#IsNotExist) to print a different error message? like

```golang
policyJson, err := os.ReadFile(policyPath)
if os.IsNotExist(err, os.ErrNotExist) {
  // suggest `policyJson, err := os.ReadFile(policyPath)olicy import`
} else {
  // suggest modify file with path or `policy import`
}
```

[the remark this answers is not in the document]
qweeah: Yes we did. @priteshbandi commented based on an [old version](https://github.com/notaryproject/notation/pull/593/files/9000315d9b9a4b574cca43d7ade7320ad2193cf7#diff-db059f4e4c4081192cb101741279bd491c9bdb0a34c5ff403bb2665afff8ee03) of code.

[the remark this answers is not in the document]
patrickzheng200: > > I'm a bit confused, I remember we already changed this to:
> 
> @patrickzheng200 And I am asking if we need to specific error handling based on **latest code**

okay. If based on the older version, then yes, we need to have different handlings for different errors. For the latest version, the error message should be good enough.

cmd/notation/policy/show.go
```diff
@@ -0,0 +1,63 @@
...
+		Args: cobra.ExactArgs(0),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runShow(cmd, opts)
+		},
+	}
+	return command
+}
+
+func runShow(command *cobra.Command, opts showOpts) error {
+	// read policy json contents
```
patrickzheng200, line 40: this comment needs to be updated.
```suggestion
	// get policy file path
```

cmd/notation/policy/show.go
```diff
@@ -0,0 +1,63 @@
...
+	policyPath, err := dir.ConfigFS().SysPath(dir.PathTrustPolicy)
+	if err != nil {
+		return fmt.Errorf("failed to obtain path of trust policy configuration file: %w", err)
+	}
+
+	// parse and validate
```
patrickzheng200, line 46:
```suggestion
	// core process
```

cmd/notation/policy/import.go
```diff
@@ -0,0 +1,83 @@
...
+			if !confirmed {
+				return nil
+			}
+		}
+	} else {
+		fmt.Fprintf(os.Stderr, "Warning: existing trust policy configuration file will be overwritten")
+	}
+
+	// read configuration
+	policyJson, err := os.ReadFile(opts.filePath)
```
shizhMSFT, line 59: nit: 
```suggestion
	policyJSON, err := os.ReadFile(opts.filePath)
```

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

byronchien (approved):

priteshbandi (dismissed):
LGTM

patrickzheng200 (dismissed):
LGTM

qweeah:
Hi @priteshbandi @shizhMSFT @patrickzheng200 I forget to sign the commits. 

I resigned with my GPG and rebased, can you help review again, thanks in advance.

patrickzheng200 (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #595, opened 2023-03-23
feat: add local sign/list/verification for OCI layout directory

patrickzheng200:
This PR adds local sign/list/verification for OCI image layout directory. 
For RC.4:
1. It only supports storing the generated signature into the target OCI layout directory.
2. It supports listing signatures within the OCI layout directory.
3. It only supports verifying signatures within the target OCI layout directory. 

This PR is based on spec PR: https://github.com/notaryproject/notation/pull/601 (Merged).

This PR is dependent on the corresponding notation-go PR: https://github.com/notaryproject/notation-go/pull/288. 
Please review the notation-go PR first.

Resolves #283.

Both remote registry and oci-layout scenario are tested. E2E tests are also included.

PR #601, opened 2023-03-25
doc: update spec for feature sign/verify local images

yizha1:
Update specs for feature sign/verify local images

### DIFF
--- a/specs/commandline/list.md
+++ b/specs/commandline/list.md
@@ -29,7 +29,7 @@ Aliases:
 Flags:
   -d, --debug             debug mode
   -h, --help              help for list
-      --local-artifact    [Preview] if set, list signatures associated with the image in OCI layout directory
+      --local-artifact    [Preview] list signatures associated with the artifact in OCI layout directory
   -p, --password string   password for registry operations (default to $NOTATION_PASSWORD if not specified)
       --plain-http        registry access via plain HTTP
   -u, --username string   username for registry operations (default to $NOTATION_USERNAME if not specified)
--- a/specs/commandline/list.md
+++ b/specs/commandline/list.md
@@ -72,7 +72,7 @@ notation list --oci-layout hello-world@sha256:xxx
 An example output:
 
 ```shell
-./hello-world@sha256:a08753c0c7bcdaaf5c2fdb375f68e860c34bffb146368982c201d41769e1763c
+hello-world@sha256:a08753c0c7bcdaaf5c2fdb375f68e860c34bffb146368982c201d41769e1763c
 └── application/vnd.cncf.notary.signature
     ├── sha256:647039638efb22a021f59675c9449dd09956c981a44b82c1ff074513c2c9f273
     └── sha256:6bfb3c4fd485d6810f9656ddd4fb603f0c414c5f0b175ef90eeb4090ebd9bfa1
--- a/specs/commandline/list.md
+++ b/specs/commandline/list.md
@@ -29,7 +29,7 @@ Aliases:
 Flags:
   -d, --debug             debug mode
   -h, --help              help for list
-      --oci-layout        [Preview] whether the artifact is OCI image layout
+      --oci-layout        [Preview] list signatures stored in OCI image layout
   -p, --password string   password for registry operations (default to $NOTATION_PASSWORD if not specified)
       --plain-http        registry access via plain HTTP
   -u, --username string   username for registry operations (default to $NOTATION_USERNAME if not specified)
@@ -49,24 +49,24 @@ An example output:
 ```shell
 localhost:5000/net-monitor:v1
 └── application/vnd.cncf.notary.signature
-    ├── sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
-    └── sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
+    ├── sha256:647039638efb22a021f59675c9449dd09956c981a44b82c1ff074513c2c9f273
+    └── sha256:6bfb3c4fd485d6810f9656ddd4fb603f0c414c5f0b175ef90eeb4090ebd9bfa1
 ```
 
-### [Preview] List all the signatures associated with the image in OCI layout directory on disk
+### [Preview] List all the signatures associated with the image in OCI layout directory
 
 The following example lists the signatures associated with the image in OCI layout directory named `hello-world`.
 
 Reference an image in OCI layout directory using tags:
 
 ```shell
-notation list --oci-layout ./hello-world:v1
+notation list --oci-layout hello-world:v1
 ```
 
 Reference an image in OCI layout directory using exact digest:
 
 ```shell
-notation list --oci-layout ./hello-world@sha256:xxx
+notation list --oci-layout hello-world@sha256:xxx
 ```
 
 An example output:
--- a/specs/commandline/list.md
+++ b/specs/commandline/list.md
@@ -29,6 +29,7 @@ Aliases:
 Flags:
   -d, --debug             debug mode
   -h, --help              help for list
+      --local-artifact    [Preview] if set, list signatures associated with the image in OCI layout directory
   -p, --password string   password for registry operations (default to $NOTATION_PASSWORD if not specified)
       --plain-http        registry access via plain HTTP
   -u, --username string   username for registry operations (default to $NOTATION_USERNAME if not specified)
@@ -39,7 +40,7 @@ Flags:
 
 ### List all the signatures of the signed container image
 
-```text
+```shell
 notation list <registry>/<repository>:<tag>
 ```
 
@@ -50,4 +51,29 @@ localhost:5000/net-monitor:v1
 └── application/vnd.cncf.notary.signature
     ├── sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
     └── sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
+```
+
+### [Preview] List all the signatures associated with the image in OCI layout directory
+
+The following example lists the signatures associated with the image in OCI layout directory named `hello-world`.
+
+Reference an image in OCI layout directory using tags:
+
+```shell
+notation list --local-artifact ./hello-world:v1
+```
+
+Reference an image in OCI layout directory using exact digest:
+
+```shell
+notation list --local-artifact ./hello-world@sha256:xxx
+```
+
+An example output:
+
+```shell
+./hello-world@sha256:a08753c0c7bcdaaf5c2fdb375f68e860c34bffb146368982c201d41769e1763c
+└── application/vnd.cncf.notary.signature
+    ├── sha256:647039638efb22a021f59675c9449dd09956c981a44b82c1ff074513c2c9f273
+    └── sha256:6bfb3c4fd485d6810f9656ddd4fb603f0c414c5f0b175ef90eeb4090ebd9bfa1
 ```
--- a/specs/commandline/list.md
+++ b/specs/commandline/list.md
@@ -55,7 +55,7 @@ localhost:5000/net-monitor:v1
 
 ### [Experimental] List all the signatures associated with the image in OCI layout directory
 
-The following example lists the signatures associated with the image in OCI layout directory named `hello-world`. To access this flag `--oci-layout` set the environment variable `NOTATION_EXPERIMENTAL`.
+The following example lists the signatures associated with the image in OCI layout directory named `hello-world`. To access this flag `--oci-layout` , set the environment variable `NOTATION_EXPERIMENTAL=1`.
 
 Reference an image in OCI layout directory using tags:
 
--- a/specs/commandline/list.md
+++ b/specs/commandline/list.md
@@ -60,13 +60,17 @@ The following example lists the signatures associated with the image in OCI layo
 Reference an image in OCI layout directory using tags:
 
 ```shell
-NOTATION_EXPERIMENTAL=1 notation list --oci-layout hello-world:v1
+export NOTATION_EXPERIMENTAL=1
+# Assume OCI layout directory hello-world is under current path
+notation list --oci-layout hello-world:v1
 ```
 
 Reference an image in OCI layout directory using exact digest:
 
 ```shell
-NOTATION_EXPERIMENTAL=1 notation list --oci-layout hello-world@sha256:xxx
+export NOTATION_EXPERIMENTAL=1
+# Assume OCI layout directory hello-world is under current path
+notation list --oci-layout hello-world@sha256:xxx
 ```
 
 An example output:
--- a/specs/commandline/list.md
+++ b/specs/commandline/list.md
@@ -29,7 +29,7 @@ Aliases:
 Flags:
   -d, --debug             debug mode
   -h, --help              help for list
-      --oci-layout        [Preview] list signatures stored in OCI image layout
+      --oci-layout        [Experimental] list signatures stored in OCI image layout
   -p, --password string   password for registry operations (default to $NOTATION_PASSWORD if not specified)
       --plain-http        registry access via plain HTTP
   -u, --username string   username for registry operations (default to $NOTATION_USERNAME if not specified)
@@ -53,20 +53,20 @@ localhost:5000/net-monitor:v1
     └── sha256:6bfb3c4fd485d6810f9656ddd4fb603f0c414c5f0b175ef90eeb4090ebd9bfa1
 ```
 
-### [Preview] List all the signatures associated with the image in OCI layout directory
+### [Experimental] List all the signatures associated with the image in OCI layout directory
 
-The following example lists the signatures associated with the image in OCI layout directory named `hello-world`.
+The following example lists the signatures associated with the image in OCI layout directory named `hello-world`. To access this flag `--oci-layout` set the environment variable `NOTATION_EXPERIMENTAL`.
 
 Reference an image in OCI layout directory using tags:
 
 ```shell
-notation list --oci-layout hello-world:v1
+NOTATION_EXPERIMENTAL=1 notation list --oci-layout hello-world:v1
 ```
 
 Reference an image in OCI layout directory using exact digest:
 
 ```shell
-notation list --oci-layout hello-world@sha256:xxx
+NOTATION_EXPERIMENTAL=1 notation list --oci-layout hello-world@sha256:xxx
 ```
 
 An example output:
--- a/specs/commandline/list.md
+++ b/specs/commandline/list.md
@@ -29,7 +29,7 @@ Aliases:
 Flags:
   -d, --debug             debug mode
   -h, --help              help for list
-      --local-artifact    [Preview] list signatures associated with the artifact in OCI layout directory
+      --oci-layout        [Preview] whether the artifact is OCI image layout
   -p, --password string   password for registry operations (default to $NOTATION_PASSWORD if not specified)
       --plain-http        registry access via plain HTTP
   -u, --username string   username for registry operations (default to $NOTATION_USERNAME if not specified)
@@ -53,20 +53,20 @@ localhost:5000/net-monitor:v1
     └── sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
 ```
 
-### [Preview] List all the signatures associated with the image in OCI layout directory
+### [Preview] List all the signatures associated with the image in OCI layout directory on disk
 
 The following example lists the signatures associated with the image in OCI layout directory named `hello-world`.
 
 Reference an image in OCI layout directory using tags:
 
 ```shell
-notation list --local-artifact ./hello-world:v1
+notation list --oci-layout ./hello-world:v1
 ```
 
 Reference an image in OCI layout directory using exact digest:
 
 ```shell
-notation list --local-artifact ./hello-world@sha256:xxx
+notation list --oci-layout ./hello-world@sha256:xxx
 ```
 
 An example output:
--- a/specs/commandline/sign.md
+++ b/specs/commandline/sign.md
@@ -169,29 +169,29 @@ notation sign --signature-manifest artifact <registry>/<repository>@<digest>
 
 ### [Preview] Sign local OCI image
 
-A local OCI image means an OCI image Layout on local disk. OCI image layout is defined by spec [OCI image layout][oci-image-layout]. It is a directory structure that contains files and folders. Users can reference an image in the layout using either tags, or the exact digest. The OCI image layout could be a tarball or a directory. Notation only supports signing OCI layout directory.
+A local OCI image means an OCI image Layout on local disk. OCI image layout is defined by spec [OCI image layout][oci-image-layout]. It is a directory structure that contains files and folders. Users can reference an image in the layout using either tags, or the exact digest. The OCI image layout could be a tarball or a directory. Notation only supports signing OCI layout directory for now.
 
-Tools like `docker buildx` support building an OCI image layout on local disk. The following example creates a tarball named `hello-world.tar` with tag `v1`, which is an OCI layout tarball on local disk. Please note that the digest can be found in the output messages.
+Tools like `docker buildx` support building an OCI image layout on local disk. The following example creates a tarball named `hello-world.tar` with tag `v1`, which is an OCI layout tarball on local disk. Please note that the digest can be found in the output messages of `docker buildx build`.
 
 ```shell
 docker buildx create --use
 docker buildx build . -f Dockerfile -o type=oci,dest=hello-world.tar -t hello-world:v1
 ```
 
-Users need to extract the tarball into a directory first. The following command creates the OCI layout directory, and the image can be referred by `./hello-world:v1` or `./hello-world@sha256:xxx`.
+Users need to extract the tarball into a directory first. The following command creates the OCI layout directory, and the image can be referenced by `./hello-world:v1` or `./hello-world@sha256:xxx`.
 
 ```shell
 mkdir hello-world
 tar -xf ./hello-world.tar -C hello-world
 ```
 
-Use flag `--local-artifact` for signing local OCI images. The signatures are stored in the layout directory and associated with the image. For example:
+Use flag `--local-artifact` for signing local OCI images. For example:
 
 ```shell
 notation sign --local-artifact ./hello-world@sha256:xxx
 ```
 
-Upon successful signing, use the `notation list` command to list the signatures, for example:
+Upon successful signing, the signature is stored in the same layout directory and associated with the image. Use `notation list` command to list the signatures, for example:
 
 ```shell
 notation list ./hello-world@sha256:xxx
--- a/specs/commandline/sign.md
+++ b/specs/commandline/sign.md
@@ -182,19 +182,19 @@ Users need to extract the tarball into a directory first, since Notation only su
 
 ```shell
 mkdir hello-world
-tar -xf ./hello-world.tar -C hello-world
+tar -xf hello-world.tar -C hello-world
 ```
 
 Use flag `--oci-layout` to sign the image stored in OCI layout directory referenced by `hello-world@sha256xxx`. For example:
 
 ```shell
-notation sign --oci-layout ./hello-world@sha256:xxx
+notation sign --oci-layout hello-world@sha256:xxx
 ```
 
 Upon successful signing, the signature is stored in the same layout directory and associated with the image. Use `notation list` command to list the signatures, for example:
 
 ```shell
-notation list --oci-layout ./hello-world@sha256:xxx
+notation list --oci-layout hello-world@sha256:xxx
 ```
 
 [oci-artifact-manifest]: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/artifact.md
--- a/specs/commandline/sign.md
+++ b/specs/commandline/sign.md
@@ -163,7 +163,7 @@ Successfully signed localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da
 
 ### [Experimental] Sign an artifact and store the signature using OCI artifact manifest
 
-To access this flag `--signature-manifest` set the environment variable `NOTATION_EXPERIMENTAL`.
+To access this flag `--signature-manifest`, set the environment variable `NOTATION_EXPERIMENTAL=1`.
 
 ```shell
 NOTATION_EXPERIMENTAL=1 notation sign --signature-manifest artifact <registry>/<repository>@<digest>
@@ -187,7 +187,7 @@ mkdir hello-world
 tar -xf hello-world.tar -C hello-world
 ```
 
-Use flag `--oci-layout` to sign the image stored in OCI layout directory referenced by `hello-world@sha256xxx`. To access this flag `--oci-layout` set the environment variable `NOTATION_EXPERIMENTAL`. For example:
+Use flag `--oci-layout` to sign the image stored in OCI layout directory referenced by `hello-world@sha256xxx`. To access this flag `--oci-layout` , set the environment variable `NOTATION_EXPERIMENTAL=1`. For example:
 
 ```shell
 NOTATION_EXPERIMENTAL=1 notation sign --oci-layout hello-world@sha256:xxx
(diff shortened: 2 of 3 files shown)


specs/commandline/list.md
```diff
@@ -29,6 +29,7 @@ Aliases:
 Flags:
   -d, --debug             debug mode
   -h, --help              help for list
+      --local-artifact    [Preview] if set, list signatures associated with the image in OCI layout directory
```
qweeah, line 32: I saw the flag naming is different from the [doc](https://hackmd.io/@patrickzheng200/B1F82vi6o) and [PR](https://github.com/notaryproject/notation/pull/595), is it intended?

specs/commandline/sign.md
```diff
@@ -33,6 +33,7 @@ Flags:
   -h,  --help                       help for sign
        --id string                  key id (required if --plugin is set). This is mutually exclusive with the --key flag
   -k,  --key string                 signing key name, for a key previously added to notation's key list. This is mutually exclusive with the --id and --plugin flags
+       --local-artifact             [Preview] if set, sign artifacts on local disk
```
patrickzheng200, line 36:
```suggestion
       --local-artifact             [Preview] sign artifact on local disk
```

specs/commandline/verify.md
```diff
@@ -40,6 +40,7 @@ Flags:
   -p,  --password string             password for registry operations (default to $NOTATION_PASSWORD if not specified)
        --plain-http                  registry access via plain HTTP
        --plugin-config stringArray   {key}={value} pairs that are passed as it is to a plugin, if the verification is associated with a verification plugin, refer plugin documentation to set appropriate values
+       --scope string                [Preview] trust policy scope for local artifact verification.
```
patrickzheng200, line 43: missing the `--local-artifact` flag here?

specs/commandline/list.md
patrickzheng200: > I found the flag naming is different from the [doc](https://hackmd.io/@patrickzheng200/B1F82vi6o) and [PR](https://github.com/notaryproject/notation/pull/595), is it intended?

Yeah, I think `--local-artifact` is better. I'll change to `--local-artifact` in the PR.

byronchien:
should `notation inspect` also support local signatures/images?

specs/commandline/verify.md
```diff
@@ -168,3 +170,25 @@ An example of output messages for a successful verification:
...
 ```
+
+### [Preview] Verify local OCI image
+
+Users should configure trust policy properly before verifying local OCI image. According to trust policy specification, `registryScopes` property determines which trust policy is applicable for the given artifact. The value of `registryScopes` should be a fully qualified URL for both local and remote images. For example, a remote image is referenced by "localhost:5000/net-monitor:v1", thus the value of `registryScopes` should contain "localhost:5000/net-monitor", which is the repository URL of the image. Local OCI image is in the form of OCI layout directory, the reference is different from a remote image, for example "./hello-world:v1". Besides flag `--local-artifact`, users need to specify which `registryScopes` is used to verify the local image using flag `--scope`. Here is an example of trust policy configured for local image `./hello-world:v1`:
+
+```jsonc
+{
+    "name": "local-images",
+    "registryScopes": [ "local/hello-world" ],
```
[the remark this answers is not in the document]
FeynmanZhou: A user may have multiple OCI image layout folders in their local environment. He may want to have fine-grained permission control on those registries.

specs/commandline/list.md
```diff
@@ -29,6 +29,7 @@ Aliases:
 Flags:
   -d, --debug             debug mode
   -h, --help              help for list
+      --local-artifact    [Preview] list signatures associated with the artifact in OCI layout directory
```
sajayantony, line 32: 1. Can we considering inferring that its a file path? @priteshbandi also brought this up. 
2. Secondly if the flag is needed can we call it something like --file or --path etc. --local is quite subjective.

kody-kimberl: For the option, what if we considered something more flexible than just a boolean. If we use something like --type (maybe this isn't the best name, but I think the concept is interesting) we could specify through the option that it is a directory. That would allow us to easily incorporate new modes for tarballs, SBOMs, or other files that could be added in the future.

---
Example:
```
notation sign $IMAGE
notation sign --type dir /path/to/registry.acme-rockets.io/net-monitor:v2
notation sign --type tar hello-world.tar
...
```

yizha1:
> should `notation inspect` also support local signatures/images?

`notation inspect` will come in later release. The plan is to implement sign/verify local images in a phased approach. There are other scenarios and commands not covered yet.

specs/commandline/list.md
yizha1: @shizhMSFT @patrickzheng200 @qweeah @priteshbandi @vaninrao10 @FeynmanZhou @toddysm @sajayantony please take a look at Kody's comment above. Now we have these options:

1. `--local-content`
2. `--local-artifact`
3. `--path`
4. `--file`
5. `--use-oci-layout`
6. `--oci-dir`
7. `--offline`
8. `file://xxx/`
9. `--dirpath`
10. `--target-path`
11. `--type`

patrickzheng200: In fact, if @yizha1 could remember, the very first version of my design was pretty similar to Kody's suggestion, i.e. specifying type of the user input. So personally speaking, I think it's a good idea as long as there's no usability concern (users need to know what options are available in Notation, and what type their input belong to).

yizha1: I vote for option 3: `--path <dir path | file path>`

qweeah: option 3 +1

patrickzheng200: @yizha1 Question for option 3: we have certificate files, trust policy file and probably local key files in the future, just calling it `--path` might bring in confusion to the user?

toddysm: Looking at the options, can we have a clear understanding what the behavior will be. I think I am confused as a user what will happen if I use one or another option. Here is an example with the `--file` option. If I specify `--file` do I sign the _file_ itself or does Notation considers the _file_ an OCI layout and signs the artifact that is represented by this layout?

The same is the confusion with the `--type tar` - do I sign the actual _tarball_ or does Notation consider the _tar_ an OCI layout and signs the artifact represented by this _tarball_?

Note that there is strong desire to sign individual blobs (and not OCI layouts) and the options above may be quite confusing for users who have that expectation.

shizhMSFT: `--path` and other generic terms are ambiguous if we want to sign a single file in the future. For instance, `--path a.tar` can mean that we are processing the `a.tar` as it is or the content in the `a.tar`.

In fact, I think `--type <type>` makes more sense. `cobra` also allows use to type `--type=<type>` For examples,
- `--type=oci` or `--type=dir` for OCI-Layout in a folder
- `--type=oci` or `--type=tar` for OCI-layout in a tarball
- `--type=file` is reserved for future file signing
- `--type=desc` is reserved for future descriptor signing

However, `--type=<type>` is still not straightforward.

As a reference, [oras](https://github.com/oras-project/oras/blob/v1.0.0/cmd/oras/internal/option/target.go#L65-L72) introduces the following flag:
>     --target type=<type>[[,<key>=<value>][...]]
>  For better UX, the boolean flag `--oci-layout` is introduced as an alias of `--target type=oci-layout`.

Similarly, we could also follow the practice of `oras`.
```sh
notation sign registry.acme-rockets.io/net-monitor:v2                     # remote registry
notation sign --oci-layout registry.acme-rockets.io/net-monitor:v2        # oci-layout folder
notation sign --oci-layout net-monitor.tar:v2 --output net-monitor_v2.sig # oci-layout tarball
notation sign --descriptor desc.json --output desc.sig                    # descriptor
notation sign --file file.bin --output file.sig                           # standalone file

notation list registry.acme-rockets.io/net-monitor:v2              # remote registry
notation list --oci-layout registry.acme-rockets.io/net-monitor:v2 # oci-layout folder
notation list --oci-layout net-monitor.tar:v2                      # oci-layout tarball

notation verify registry.acme-rockets.io/net-monitor:v2                                                           # remote registry
notation verify --scope registry.acme-rockets.io/net-monitor --oci-layout registry.acme-rockets.io/net-monitor:v2 # oci-layout folder
notation verify --scope registry.acme-rockets.io/net-monitor --oci-layout net-monitor.tar:v2                      # oci-layout tarball
notation verify --scope registry.acme-rockets.io/net-monitor --descriptor desc.json --signature desc.sig          # descriptor
notation verify --scope registry.acme-rockets.io/net-monitor --file file.bin --signature file.sig                 # standalone file
```

specs/commandline/verify.md
```diff
@@ -168,3 +170,25 @@ An example of output messages for a successful verification:
...
 Successfully verified signature for localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
 ```
+
+### [Preview] Verify local OCI image
+
+Users should configure trust policy properly before verifying local OCI image. According to trust policy specification, `registryScopes` property determines which trust policy is applicable for the given artifact. The value of `registryScopes` should be a fully qualified URL for both local and remote images. For example, a remote image is referenced by "localhost:5000/net-monitor:v1", thus the value of `registryScopes` should contain "localhost:5000/net-monitor", which is the repository URL of the image. Local OCI image is in the form of OCI layout directory, the reference is different from a remote image, for example "./hello-world:v1". Besides flag `--local-artifact`, users need to specify which `registryScopes` is used to verify the local image using flag `--scope`. Here is an example of trust policy configured for local image `./hello-world:v1`:
```
shizhMSFT, line 176: `--scope` specifies the reference `<registry>/<repository>` that the artifact is intended to live. It can be viewed that `--scope` overrides the intended scope for any artifact. 

If the artifact is in a remote registry, the scope can be extracted from the reference, and can be override by `--scope`.
For example,
```
notation verify --scope foo.io/net-monitor bar.io/net-monitor@sha256:xxxx
```
The above command tries to verify `bar.io/net-monitor@sha256:xxxx` as if it is in `foo.io/net-monitor`. It is useful for mirrors.

Following the same logic,
```
notation verify --scope foo.io/net-monitor --oci-layout net-monitor.tar@sha256:xxxx
```
The above command tries to verify `sha256:xxxx` in `net-monitor.tar` as if it is in `foo.io/net-monitor`. Since we cannot extract scope from `net-monitor.tar`, the flag `--scope` becomes a required flag.

specs/commandline/list.md
toddysm: Reading what @shizhMSFT wrote above I think we should look deeper in what we can do with `--type`. Maybe we should rename it to `--artifact-type` and have a predefined list of _types_ that can be used. I would prefer a single switch with multiple pre-defined values instead of multiple switches. 

Looking at what @shizhMSFT wrote above, we currently plan to have the following _types_
- `oci-layout`
- `oci-archive` (does it need to be tar?)
- `descriptor`

In the future, we will have:
- `file` or `blob` although `blob` may be confusing because everything is a `blob`

yizha1: @priteshbandi @vaninrao10  @patrickzheng200 @iamsamirzon @kody-kimberl please take a look at @shizhMSFT and @toddysm's proposal above. Your input is highly appreciated. We'd better finalize it ASAP.

specs/commandline/verify.md
vaninrao10: I and Samir had a discussion and agree with what Shiwei has proposed.

specs/commandline/list.md
```diff
@@ -29,6 +29,7 @@ Aliases:
 Flags:
   -d, --debug             debug mode
   -h, --help              help for list
+      --oci-layout        [Preview] whether the artifact is OCI image layout
```
shizhMSFT, line 32:
```suggestion
      --oci-layout        [Preview] use OCI image layout for storage
```

specs/commandline/list.md
```diff
@@ -51,3 +52,28 @@ localhost:5000/net-monitor:v1
...
 ```
+
+### [Preview] List all the signatures associated with the image in OCI layout directory on disk
+
+The following example lists the signatures associated with the image in OCI layout directory named `hello-world`.
+
+Reference an image in OCI layout directory using tags:
+
+```shell
+notation list --oci-layout ./hello-world:v1
```
shizhMSFT, line 63:
```suggestion
notation list --oci-layout hello-world:v1
```

or

```suggestion
notation list --oci-layout path/to/hello-world:v1
```

shizhMSFT: The prefix `./` is redundant.

specs/commandline/sign.md
```diff
@@ -166,6 +167,37 @@ Successfully signed localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da
 notation sign --signature-manifest artifact <registry>/<repository>@<digest>
 ```
 
+### [Preview] Sign artifacts in OCI layout directory on disk
+
+Artifacts can be stored on disk in the form of OCI image Layout defined in spec [OCI image layout][oci-image-layout]. It is a directory structure that contains files and folders. The OCI image layout could be a tarball or a directory in the filesystem. For example, a file named `hello-world.tar` or a directory named `hello-world`. Notation only supports signing OCI layout directory for now. Users can reference an artifact in the layout using either tags, or the exact digest. For example, use `hello-world:v1` or `hello-world@sha256xxx` to reference the container image in OCI layout directory named `hello-world`.
```
shizhMSFT, line 172: Currently, we don't support tarball since we don't support `--output <signature_path>`. We can introduce this in the next iteration.

specs/commandline/list.md
toddysm: I think for usability and compatibility purposes it is better to start with the `--target type=...`. We will need to depecate the `--oci-layout` in the future and customers will also need to change their scripts when we do that. This will make the customer experience more cumbersome.

specs/commandline/sign.md
```diff
@@ -33,6 +33,7 @@ Flags:
   -h,  --help                       help for sign
        --id string                  key id (required if --plugin is set). This is mutually exclusive with the --key flag
   -k,  --key string                 signing key name, for a key previously added to notation's key list. This is mutually exclusive with the --id and --plugin flags
+       --oci-layout                 [Preview] whether the artifact is OCI image layout
```
patrickzheng200, line 36:
```suggestion
       --oci-layout                 [Preview] use OCI image layout for storage
```

specs/commandline/verify.md
```diff
@@ -37,9 +37,11 @@ Usage:
 Flags:
   -d,  --debug                       debug mode
   -h,  --help                        help for verify
+       --oci-layout                  [Preview] whether artifact is OCI image layout
```
patrickzheng200, line 40:
```suggestion
       --oci-layout                 [Preview] use OCI image layout for storage
```

specs/commandline/sign.md
yizha1: > Currently, we don't support tarball since we don't support `--output <signature_path>`. We can introduce this in the next iteration.

@shizhMSFT I understand this. Do you expect me to add this explanation to the spec? I don't think we need to mention this in spec.

specs/commandline/list.md
yizha1: This flag is to specify the signing artifact type, but this sentence `use OCI image layout for storage` sounds like storing the artifact or signature in OCI image layout.

specs/commandline/sign.md
FeynmanZhou: Suggestion: Sign the artifact stored in the OCI image layout

specs/commandline/verify.md
FeynmanZhou: Suggestion: --oci-layout   [Preview] Verify the artifact stored in the OCI image layout

specs/commandline/list.md
yizha1: @toddysm the `--oci-layout` will not be deprecated, it is the shorthand for `--target type=oci-layout`. Later when `--target type=<type>` is introduced, users can use either `--target type=oci-layout` or `--oci-layout`. Another reason is, there is not much time left for RC-4, adding `--target type=...` requiring more effort, we will see whether there is a chance between RC-4 and v1. Does it make sense to you? See an example from `docker build buildx -h`
image

yizha1: @toddysm the `--oci-layout` will not be deprecated, it is the shorthand for `--target type=oci-layout`. Later when `--target type=<type>` is introduced, users can use either `--target type=oci-layout` or `--oci-layout`. Another reason is, there is not much time left for RC-4, adding `--target type=...` requiring more effort, we will see whether there is a chance between RC-4 and v1. Does it make sense to you? See an example from `docker build buildx -h`
image

specs/commandline/list.md
```diff
@@ -48,6 +49,31 @@ An example output:
...
+Reference an image in OCI layout directory using exact digest:
+
+```shell
+notation list --oci-layout hello-world@sha256:xxx
+```
+
+An example output:
+
+```shell
+./hello-world@sha256:a08753c0c7bcdaaf5c2fdb375f68e860c34bffb146368982c201d41769e1763c
```
shizhMSFT, line 75:
```suggestion
hello-world@sha256:a08753c0c7bcdaaf5c2fdb375f68e860c34bffb146368982c201d41769e1763c
```

specs/commandline/sign.md
```diff
@@ -166,6 +167,37 @@ Successfully signed localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da
...
+Use flag `--oci-layout` to sign the image stored in OCI layout directory referenced by `hello-world@sha256xxx`. For example:
+
+```shell
+notation sign --oci-layout ./hello-world@sha256:xxx
+```
+
+Upon successful signing, the signature is stored in the same layout directory and associated with the image. Use `notation list` command to list the signatures, for example:
+
+```shell
+notation list --oci-layout ./hello-world@sha256:xxx
```
shizhMSFT, line 197:
```suggestion
notation list --oci-layout hello-world@sha256:xxx
```

specs/commandline/sign.md
```diff
@@ -166,6 +167,37 @@ Successfully signed localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da
...
+
+```shell
+mkdir hello-world
+tar -xf ./hello-world.tar -C hello-world
+```
+
+Use flag `--oci-layout` to sign the image stored in OCI layout directory referenced by `hello-world@sha256xxx`. For example:
+
+```shell
+notation sign --oci-layout ./hello-world@sha256:xxx
```
shizhMSFT, line 191:
```suggestion
notation sign --oci-layout hello-world@sha256:xxx
```

specs/commandline/sign.md
```diff
@@ -166,6 +167,37 @@ Successfully signed localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da
...
+```shell
+docker buildx create --use
+docker buildx build . -f Dockerfile -o type=oci,dest=hello-world.tar -t hello-world:v1
+```
+
+Users need to extract the tarball into a directory first, since Notation only support OCI layout directory for now. The following command creates the OCI layout directory.
+
+```shell
+mkdir hello-world
+tar -xf ./hello-world.tar -C hello-world
```
shizhMSFT, line 185:
```suggestion
tar -xf hello-world.tar -C hello-world
```

specs/commandline/verify.md
```diff
@@ -168,3 +170,25 @@ An example of output messages for a successful verification:
...
+    "registryScopes": [ "localhost:5000/hello-world" ],
+    "signatureVerification": {
+        "level" : "strict"
+    },
+    "trustStores": [ "ca:hello-world" ],
+    "trustedIdentities": ["*"]
+}
+```
+
+To verify image `./hello-world:v1`, user should use flag `--oci-layout` and `--scope` together, for example:
```
shizhMSFT, line 190:
```suggestion
To verify image `hello-world:v1`, user should use flag `--oci-layout` and `--scope` together, for example:
```

specs/commandline/verify.md
```diff
@@ -37,9 +37,11 @@ Usage:
 Flags:
   -d,  --debug                       debug mode
   -h,  --help                        help for verify
+       --oci-layout                  [Preview] verify the artifact stored in OCI image layout
```
patrickzheng200, line 40:
```suggestion
       --oci-layout                 verify artifact stored in OCI image layout
```

specs/commandline/sign.md
```diff
@@ -33,6 +33,7 @@ Flags:
   -h,  --help                       help for sign
        --id string                  key id (required if --plugin is set). This is mutually exclusive with the --key flag
   -k,  --key string                 signing key name, for a key previously added to notation's key list. This is mutually exclusive with the --id and --plugin flags
+       --oci-layout                 [Preview] sign the artifact stored in OCI image layout
```
patrickzheng200, line 36:
```suggestion
       --oci-layout                 sign artifact stored in OCI image layout
```

specs/commandline/verify.md
```diff
@@ -37,9 +37,11 @@ Usage:
...
   -p,  --password string             password for registry operations (default to $NOTATION_PASSWORD if not specified)
        --plain-http                  registry access via plain HTTP
        --plugin-config stringArray   {key}={value} pairs that are passed as it is to a plugin, if the verification is associated with a verification plugin, refer plugin documentation to set appropriate values
+       --scope string                [Preview] set trust policy scope for artifact verification, only use it with flag "--oci-layout"
```
patrickzheng200, line 44:
```suggestion
       --scope string                [Preview] set trust policy scope for verification, required to be used with flag "--oci-layout"
```

specs/commandline/list.md
FeynmanZhou: Agree with adding `--target type=...` flag in v1.0.0. There is only one target type `oci-layout` in RC.4, it's reasonable to use `--oci-layout` as a shorthand for `--target type=oci-layout`.

specs/commandline/sign.md
FeynmanZhou: The [Preview] tag should be reserved in RC.4.

specs/commandline/verify.md
FeynmanZhou: The [Preview] label should be reserved in RC.4.

specs/commandline/sign.md
patrickzheng200: > The [Preview] label should be reserved in RC.4.

Could you share any reason behind this? IMO, if we decide a flag won't change in the future, we should not tag it as `[Preview]`.

specs/commandline/list.md
```diff
@@ -29,6 +29,7 @@ Aliases:
 Flags:
   -d, --debug             debug mode
   -h, --help              help for list
+      --oci-layout        [Preview] list signatures stored in OCI image layout
```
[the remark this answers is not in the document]
FeynmanZhou: In general, [Preview] means this is a new feature for an early preview and is on track to become a permanent part of the product. It may be turned off only if this feature runs into a problem in the future.

[Experimental] means this new feature can radically change or completely disappear at any time. It's not recommended for production use.

So [Preview] may be still necessary for `--oci-layout` in case this flag is renamed or changed in the future.

specs/commandline/list.md
toddysm: I am not sure having shorthand is a good idea. Having too many switches pollutes the CLI and makes its use hard to use.

specs/commandline/list.md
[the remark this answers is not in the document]
yizha1: @FeynmanZhou Thanks Feynman for the explanation. We have a work item in Ga for feature maturity. I think at least one job we need to do is to document the meaning of the labels for feature maturity. cc @toddysm for comments.

specs/commandline/list.md
yizha1: @toddysm We will consider adding `--target type=` if the time allows for 1.0. For RC.4 it is a good start, since we only have one type now, compared to `--target type=oci-layout`, `--oci-layout` is short and convenient.  Shorthand could be a good practice for some scenarios, if we have multiple choice, and we know what option 80% users will choose, then a shorthand will be convenient for them, since it is short. In our cases, OCI image layout could be the most choice, since it is the only standard now OCI images and other OCI artifacts can be stored in a directory structure under filesystem (could be completely offline), and also easy to copy to remote registry.

specs/commandline/sign.md
yizha1: Same reason as Feynman explained above.

specs/commandline/list.md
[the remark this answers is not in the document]
toddysm: The definition of [Preview] vs [Experimental] from @FeynmanZhou makes sense to me. In this case I would say the switch `--oci-layout` should be [Experimental].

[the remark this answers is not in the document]
yizha1: @toddysm as explained here https://github.com/notaryproject/notation/pull/601#discussion_r1155038676, I think `Preview` is more proper. This flag also has the most votes from the community. Both `--target type=oci-layout` (coming in later release) and `--oci-layout` will be supported.

patrickzheng200 (dismissed):
LGTM

specs/commandline/sign.md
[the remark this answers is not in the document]
sajayantony: Follow up conversion was to introduce a type and consider avoiding more flags if possible /cc @toddysm

specs/commandline/list.md
yizha1: As discussed, `--oci-layout` and `--scope` will be experimental feature which requires system variable `NOTATION_EXPERIMENTAL` to be set.

specs/commandline/sign.md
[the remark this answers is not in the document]
yizha1: As discussed, `--oci-layout` and `--scope` will be experimental features which require system variable `NOTATION_EXPERIMENTAL` to be set.
sign a file and output the signature blob/manifest will be considered post v1 release, and CLI spec should be created and review first before implementation.

specs/commandline/list.md
```diff
@@ -48,6 +49,31 @@ An example output:
...
 └── application/vnd.cncf.notary.signature
-    ├── sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
-    └── sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
+    ├── sha256:647039638efb22a021f59675c9449dd09956c981a44b82c1ff074513c2c9f273
+    └── sha256:6bfb3c4fd485d6810f9656ddd4fb603f0c414c5f0b175ef90eeb4090ebd9bfa1
+```
+
+### [Experimental] List all the signatures associated with the image in OCI layout directory
+
+The following example lists the signatures associated with the image in OCI layout directory named `hello-world`. To access this flag `--oci-layout` set the environment variable `NOTATION_EXPERIMENTAL`.
```
patrickzheng200, line 58:
```suggestion
The following example lists the signatures associated with the image in OCI layout directory named `hello-world`. To access this flag, `--oci-layout`, set the environment variable `NOTATION_EXPERIMENTAL=1`.
```

specs/commandline/sign.md
```diff
@@ -162,10 +163,43 @@ Successfully signed localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da
 
 ### [Experimental] Sign an artifact and store the signature using OCI artifact manifest
 
+To access this flag `--signature-manifest` set the environment variable `NOTATION_EXPERIMENTAL`.
```
patrickzheng200, line 166:
```suggestion
To access this flag, `--signature-manifest`, set the environment variable `NOTATION_EXPERIMENTAL=1`.
```

specs/commandline/sign.md
```diff
@@ -162,10 +163,43 @@ Successfully signed localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da
...
+```
+
+Users need to extract the tarball into a directory first, since Notation only support OCI layout directory for now. The following command creates the OCI layout directory.
+
+```shell
+mkdir hello-world
+tar -xf hello-world.tar -C hello-world
+```
+
+Use flag `--oci-layout` to sign the image stored in OCI layout directory referenced by `hello-world@sha256xxx`. To access this flag `--oci-layout` set the environment variable `NOTATION_EXPERIMENTAL`. For example:
```
patrickzheng200, line 190:
```suggestion
Use flag `--oci-layout` to sign the image stored in OCI layout directory referenced by `hello-world@sha256xxx`. To access this flag, `--oci-layout`, set the environment variable `NOTATION_EXPERIMENTAL=1`. For example:
```

patrickzheng200 (dismissed):
LGTM

specs/commandline/verify.md
```diff
@@ -37,9 +37,11 @@ Usage:
 Flags:
   -d,  --debug                       debug mode
   -h,  --help                        help for verify
+       --oci-layout                  [Experimental] verify the artifact stored in OCI image layout
   -p,  --password string             password for registry operations (default to $NOTATION_PASSWORD if not specified)
        --plain-http                  registry access via plain HTTP
        --plugin-config stringArray   {key}={value} pairs that are passed as it is to a plugin, if the verification is associated with a verification plugin, refer plugin documentation to set appropriate values
+       --scope string                [Experimental] set trust policy scope for artifact verification, required to be used with flag "--oci-layout"
```
priteshbandi, line 44: Don't know how to phrase it better but we should mention that `scope` can only be used when `--oci-layout` is specified.

specs/commandline/verify.md
```diff
@@ -168,3 +170,25 @@ An example of output messages for a successful verification:
...
+    },
+    "trustStores": [ "ca:hello-world" ],
+    "trustedIdentities": ["*"]
+}
+```
+
+To verify image `hello-world:v1`, user should set the environment variable `NOTATION_EXPERIMENTAL` and use flags `--oci-layout` and `--scope` together. for example:
+
+```shell
+NOTATION_EXPERIMENTAL=1 notation verify --oci-layout --scope "localhost:5000/hello-world" hello-world:v1
```
priteshbandi, line 193: Shouldn't `localhost:5000/hello-world" hello-world:v1` be disk path?

specs/commandline/sign.md
```diff
@@ -162,10 +163,43 @@ Successfully signed localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da
 
 ### [Experimental] Sign an artifact and store the signature using OCI artifact manifest
```
priteshbandi, line 164: Are we putting `--signature-manifest` behind experimental flag?

specs/commandline/sign.md
```diff
@@ -162,10 +163,43 @@ Successfully signed localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da
...
+
+```shell
+mkdir hello-world
+tar -xf hello-world.tar -C hello-world
+```
+
+Use flag `--oci-layout` to sign the image stored in OCI layout directory referenced by `hello-world@sha256xxx`. To access this flag `--oci-layout` , set the environment variable `NOTATION_EXPERIMENTAL=1`. For example:
+
+```shell
+NOTATION_EXPERIMENTAL=1 notation sign --oci-layout hello-world@sha256:xxx
```
priteshbandi, line 193: How do we know where is the directory/file to be signed?

specs/commandline/sign.md
yizha1: @priteshbandi  Yes, it is experimental indeed, as you may know that artifact manifest is removed from OCI 1.1 scope.

specs/commandline/verify.md
yizha1: @priteshbandi Currently the trust policy spec doesn't specify what will be the registryScope for a local content on filesystem, so the scope value here is an example, that it could be a URL that the local content will be stored in the registry later phase, if it is configured in trust policy. The disk absolute path could be very long, and relative path could be duplicated with others. Welcome any suggestions. Since it is experimental, we could iterate it gradually.

specs/commandline/sign.md
```diff
@@ -33,6 +33,7 @@ Flags:
   -h,  --help                       help for sign
        --id string                  key id (required if --plugin is set). This is mutually exclusive with the --key flag
   -k,  --key string                 signing key name, for a key previously added to notation's key list. This is mutually exclusive with the --id and --plugin flags
+       --oci-layout                 [Experimental] sign the artifact stored in OCI image layout
```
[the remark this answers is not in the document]
yizha1: I try to avoid using locally, since it could be done on a filesystem mounted from remote disk.

specs/commandline/sign.md
yizha1: It is users build the oci layout tarball or extract it to a directory, so users should be aware of the path to the directory or file

priteshbandi (approved):
LGTM

patrickzheng200 (approved):
LGTM

PR #602, opened 2023-03-27
feat: use Referrers API per OCI v1.1 spec 

### LINKED ISSUE #575: Verify referrers API when pushing image manifest

Notation CLI

Since RC-2 release, Notation supports storing signatures using image manifest. But [referrers-tag-schema](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#referrers-tag-schema) is used when pushing image manifest, which is not compliant with [OCI 1.1 spec](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#unavailable-referrers-api).

Store the image manifest by verifying [Referrers API](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers) and fallback according to [Backwards Compatibility](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#backwards-compatibility)

### How can we reproduce it?

- Install Notation CLI v1.0.0-rc.2
- Use `notation sign` command to sign artifacts with flag `--signature-manifest` set to `image`
- Check that the image index was pushed even the referers API is enabled in the target registry

### Describe your environment

Linux

Notation v1.0.0-rc.2

patrickzheng200:
This PR: when storing signature using oci image manifest, the Referrers API will be used. If Referrers API is not supported, automatically fallback to use Referrers Tag Schema.

Resolves #575.

### DIFF
--- a/cmd/notation/registry.go
+++ b/cmd/notation/registry.go
@@ -57,22 +57,18 @@ func getSignatureRepositoryForSign(ctx context.Context, opts *SecureFlagOpts, re
 	}
 
 	// Notation enforces the following two paths during Sign process:
-	// 1. OCI artifact manifest uses the Referrers API
+	// 1. OCI artifact manifest uses the Referrers API.
 	// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers
-	// 2. OCI image manifest uses the Referrers Tag Schema
+	// 2. OCI image manifest uses the Referrers API and automatically fallback
+	// 	  to Referrers Tag Schema if Referrers API is not supported.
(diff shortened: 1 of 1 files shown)


byronchien (approved):

priteshbandi (approved):
LGTM

patrickzheng200:
> LGTM but IANAM. Just a question: why notation need to detect the support status of referrer API? Shouldn't it be automatically taken care of from oras-go side?

Correct. The fallback logic is indeed in oras-go.

shizhMSFT (approved):
LGTM

JeyJeyGao (approved):
LGTM

Issue #604, opened 2023-03-29
Notation delete returns confusing message


toddysm:
### What is the areas you experience the issue in?

Notation CLI

### What is not working as expected?

I am trying to delete a key from the keystore but after issuing the `notation key delete` command, I see this:
```
~ notation key delete $TEST_KEY_NAME
wabbit-networks.io: unmarked as default
```

This is quite confusing :( Did the key get deleted or just "unmarked as default"? There is no other command that I can use to "remove" the key and this is my intention - the key is expired and not needed anymore.  

### What did you expect to happen?

I expect the key to be removed from the trust store and the message to be clear that the key is removed.

### How can we reproduce it?

Use the command to delete the key

### Describe your environment

N/A

### What is the version of your Notation CLI or Notation Library?

Notation: Notary v2, A tool to sign, store, and verify artifacts.

Version:     1.0.0-rc.3
Go version:  go1.20.1
Git commit:  233c0ea71edfb68b951eb54a739a101e2a05cd26

FeynmanZhou:
For `notation key delete`, I think it should have the option to delete the local key from the Trust Store especially when it is invalid or for testing purpose only. Meanwhile, we could refine the error message to avoid confusion when deleting a signing key.

For `notation cert delete`, if you want to delete the certificate from the Trust Store, try `notation cert delete --type ca --store <trust_storeName> (--all | <cert_fileName>)`. The error message should be refined and those two tags should be considered as required options.

PR #606, opened 2023-03-30
doc: improve error output in notation key and notation cert

FeynmanZhou:
Update CLI spec to improve error output in `notation key` and `notation cert` and resolve the UX problem in #604

### DIFF
--- a/specs/commandline/certificate.md
+++ b/specs/commandline/certificate.md
@@ -208,7 +208,7 @@ A prompt is showed asking user to confirm the deletion. Upon successful deletion
 notation certificate delete --type <type> --store <name> <cert_fileName>
 ```
 
-A prompt is showed asking user to confirm the deletion. Upon successful deletion, the specific certificate is deleted from the trust store named `<name>` of type `<type>`. The output message is printed out as following:
+A prompt is displayed, asking the user to confirm the deletion. Upon successful deletion, the specific certificate is deleted from the trust store named `<name>` of type `<type>`. The output message is printed out as following:
 
 ```text
 Successfully deleted <cert_fileName> from the trust store. 
--- a/specs/commandline/certificate.md
+++ b/specs/commandline/certificate.md
@@ -208,7 +208,17 @@ A prompt is showed asking user to confirm the deletion. Upon successful deletion
 notation certificate delete --type <type> --store <name> <cert_fileName>
 ```
 
-A prompt is showed asking user to confirm the deletion. Upon successful deletion, the specific certificate is deleted in trust store named `<name>` of type `<type>`. If deletion fails, an error message with specific reasons is printed out.
+A prompt is showed asking user to confirm the deletion. Upon successful deletion, the specific certificate is deleted from the trust store named `<name>` of type `<type>`. The output message is printed out as following:
+
+```text
+Successfully deleted <cert_fileName> from the trust store. 
+```
+
+If users execute the deletion without specifying required flags using `notation cert delete <cert_fileName>`, the deletion fails and the error output message is printed out as follows:
+
+```text
+Error: You need to specify the required flag "--type" and "--store". Try "notation cert delete -h" to view examples.
+```
 
 ### Generate a local RSA key and a corresponding self-generated certificate for testing purpose and add the certificate into trust store
 
--- a/specs/commandline/certificate.md
+++ b/specs/commandline/certificate.md
@@ -217,7 +217,7 @@ Successfully deleted <cert_fileName> from the trust store.
 If users execute the deletion without specifying required flags using `notation cert delete <cert_fileName>`, the deletion fails and the error output message is printed out as follows:
 
 ```text
-Error: You need to specify the required flag "--type" and "--store". Try "notation cert delete -h" to view examples.
+Error: required flag(s) "store", "type" not set
 ```
 
 ### Generate a local RSA key and a corresponding self-generated certificate for testing purpose and add the certificate into trust store
--- a/specs/commandline/key.md
+++ b/specs/commandline/key.md
@@ -124,12 +124,12 @@ notation key remove <key_name>
 - Upon successful removal of a local testing key created by notation, the output message is printed out as follows:
 
 ```text
-Removed <key_name> from Notation signing key list. The source key file still exists locally.
+Removed <key_name> from Notation signing key list. The source key file still exists.
 ```
 - Upon successful removal of a key associated with a KMS, the output message is printed out as follows:
 
 ```text
-Removed <key_name> from Notation signing key list.
+Removed <key_name> from Notation signing key list. The source key files still exist.
 ```
 
 ### Remove two keys from Notation signing key list
@@ -141,7 +141,7 @@ notation key remove <key_name_1> <key_name_2>
 Upon successful execution, the output message is printed out as follows. Please be noted if default signing key is removed, Notation will not automatically assign a new default signing key. User needs to update the default signing key explicitly.
 
 ```text
-Removed the following keys from Notation signing key list. The source key files still exist locally.
+Removed the following keys from Notation signing key list. The source key files still exist.
 <key_name_1>
 <key_name_2>
 ```
--- a/specs/commandline/key.md
+++ b/specs/commandline/key.md
@@ -121,10 +121,15 @@ Upon successful execution, a list of keys is printed out with information of nam
 notation key remove <key_name>
 ```
 
-Upon successful execution, the output message is printed out as following:
+- Upon successful execution, the output message is printed out as below when removed a key created by notation for local testing purpose:
 
 ```text
 Removed <key_name> from signing key list. You still need to delete the key file from <key_path> and delete the certificate from <cert_path>.
+```
+- Upon successful execution, the output message is printed out as below when removed a key associates with a KMS:
+
+```text
+Removed <key_name> from signing key list.
 ```
 
 ### Remove two keys from signing key list
--- a/specs/commandline/key.md
+++ b/specs/commandline/key.md
@@ -121,12 +121,12 @@ Upon successful execution, a list of keys is printed out with information of nam
 notation key remove <key_name>
 ```
 
-- Upon successful execution, the output message is printed out as below when removed a key created by notation for local testing purpose:
+- Upon successful removal of a local testing key created by notation, the output message is printed out as follows:
 
 ```text
-Removed <key_name> from signing key list. You still need to delete the key file from <key_path> and delete the certificate from <cert_path>.
+Removed <key_name> from signing key list. The source key file is still existed in <key_path>.
 ```
-- Upon successful execution, the output message is printed out as below when removed a key associates with a KMS:
+- Upon successful removal of a key associated with a KMS, the output message is printed out as follows:
 
 ```text
 Removed <key_name> from signing key list.
@@ -138,4 +138,10 @@ Removed <key_name> from signing key list.
 notation key remove <key_name_1> <key_name_2>
 ```
 
-Upon successful execution, the names of removed signing keys are printed out. Please be noted if default signing key is removed, Notation will not automatically assign a new default signing key. User needs to update the default signing key explicitly.
+Upon successful execution, the output message is printed out as follows. Please be noted if default signing key is removed, Notation will not automatically assign a new default signing key. User needs to update the default signing key explicitly.
+
+```text
+Removed the following keys from signing key list. The source key files are still existed in <key_path>.
+<key_name_1>
+<key_name_2>
+```
--- a/specs/commandline/key.md
+++ b/specs/commandline/key.md
@@ -16,7 +16,7 @@ Usage:
 
 Available Commands:
   add         Add key to Notation signing key list
-  remove      Remove key from Notation signing key list
+  delete      Remove key from Notation signing key list
   list        List keys used for signing
   update      Update key in Notation signing key list
 
--- a/specs/commandline/key.md
+++ b/specs/commandline/key.md
@@ -124,7 +124,7 @@ notation key remove <key_name>
 - Upon successful removal of a local testing key created by notation, the output message is printed out as follows:
 
 ```text
-Removed <key_name> from Notation signing key list. The source key file is still existed in <key_path>.
+Removed <key_name> from Notation signing key list. The source key file still exists locally.
 ```
 - Upon successful removal of a key associated with a KMS, the output message is printed out as follows:
 
@@ -141,7 +141,7 @@ notation key remove <key_name_1> <key_name_2>
 Upon successful execution, the output message is printed out as follows. Please be noted if default signing key is removed, Notation will not automatically assign a new default signing key. User needs to update the default signing key explicitly.
 
 ```text
-Removed the following keys from Notation signing key list. The source key files are still existed in <key_path>.
+Removed the following keys from Notation signing key list. The source key files still exist locally.
 <key_name_1>
 <key_name_2>
 ```
--- a/specs/commandline/key.md
+++ b/specs/commandline/key.md
@@ -48,7 +48,7 @@ Flags:
 Remove key from Notation signing key list
 
 Usage:
-  notation key remove [flags] <key_name>...
+  notation key delete [flags] <key_name>...
 
 Flags:
   -d, --debug     debug mode
@@ -118,7 +118,7 @@ Upon successful execution, a list of keys is printed out with information of nam
 ### Remove a specified key from Notation signing key list
 
 ```shell
-notation key remove <key_name>
+notation key delete <key_name>
 ```
 
 - Upon successful removal of a local testing key created by notation, the output message is printed out as follows:
@@ -135,7 +135,7 @@ Removed <key_name> from Notation signing key list. The source key files still ex
 ### Remove two keys from Notation signing key list
 
 ```shell
-notation key remove <key_name_1> <key_name_2>
+notation key delete <key_name_1> <key_name_2>
 ```
 
 Upon successful execution, the output message is printed out as follows. Please be noted if default signing key is removed, Notation will not automatically assign a new default signing key. User needs to update the default signing key explicitly.
--- a/specs/commandline/key.md
+++ b/specs/commandline/key.md
@@ -16,7 +16,7 @@ Usage:
 
 Available Commands:
   add         Add key to signing key list
-  delete      Delete key from signing key list
+  remove      Remove key from signing key list
   list        List keys used for signing
   update      Update key in signing key list
 
@@ -45,10 +45,10 @@ Flags:
 ### notation key delete
 
 ```text
-Delete key from signing key list
+Remove key from signing key list
 
 Usage:
-  notation key delete [flags] <key_name>...
+  notation key remove [flags] <key_name>...
 
 Flags:
   -d, --debug     debug mode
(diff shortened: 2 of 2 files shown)


specs/commandline/key.md
```diff
@@ -45,10 +45,10 @@ Flags:
 ### notation key delete
 
 ```text
-Delete key from signing key list
+Remove key from signing key list
 
 Usage:
-  notation key delete [flags] <key_name>...
+  notation key remove [flags] <key_name>...
```
shizhMSFT, line 51: Do we want to have an alias such as
```
Usage:
  notation key remove [flags] <key_name>...

Aliases:
  remove, delete
```

specs/commandline/key.md
```diff
@@ -115,10 +115,22 @@ notation key list
...
 ```shell
-notation key delete <key_name_1> <key_name_2>
+notation key remove <key_name>
 ```
 
-Upon successful execution, the names of deleted signing keys are printed out. Please be noted if default signing key is deleted, Notation will not automatically assign a new default signing key. User needs to update the default signing key explicitly.
+Upon successful execution, the output message is printed out as following:
+
+```text
+Removed <key_name> from signing key list. You still need to delete the key file from <key_path> and delete the certificate from <cert_path>.
```
shizhMSFT, line 127: What's the output if the key is associated with a plugin?

specs/commandline/certificate.md
```diff
@@ -208,7 +208,17 @@ A prompt is showed asking user to confirm the deletion. Upon successful deletion
...
+A prompt is showed asking user to confirm the deletion. Upon successful deletion, the specific certificate is deleted from the trust store named `<name>` of type `<type>`. The output message is printed out as following:
+
+```text
+Successfully deleted <cert_fileName> from the trust store. 
+```
+
+If users execute the deletion without specifying required flags using `notation cert delete <cert_fileName>`, the deletion fails and the error output message is printed out as follows:
+
+```text
+Error: You need to specify the required flag "--type" and "--store". Try "notation cert delete -h" to view examples.
```
patrickzheng200, line 220: This Error message is triggered/created automatically by Cobra:
`Error: required flag(s) "store", "type" not set`

specs/commandline/key.md
patrickzheng200, line 127: 1. Since we do not support local key, the key created by `generate-test` becomes the only use case of this error message.
2. `You still need to delete the certificate from <cert_path>.` We can show this message only when the key is created by `generate-test`. Because in production, `decide not to Sign with a key anymore` is different from `decide not to trust a certificate in Verification anymore`.

specs/commandline/certificate.md
FeynmanZhou: Is the error output msg customizable?

specs/commandline/key.md
FeynmanZhou: Agree 1 and 2. This is against locally creating the test key/cert by `generate-test`.
I will add a prerequisite to explain this scenario in Spec. We also need to differentiate scenarios and explicitly tell users that the remote key/cert can not be deleted by Notation.

specs/commandline/key.md
FeynmanZhou: I would not suggest adding an alias here since `Delete` has a different meaning from `Remove`. `Remove` is more accurate in the context of `notation key`. 

I am open to this question. @toddysm @iamsamirzon @yizha1 @vaninrao10

yizha1: My opinion is No. So there are no aliases for `notation key remove` command

specs/commandline/certificate.md
patrickzheng200: > Is the error output msg customizable?

As far as I know, no, it cannot be customized.

specs/commandline/key.md
```diff
@@ -115,10 +115,22 @@ notation key list
 
 Upon successful execution, a list of keys is printed out with information of name, key path, certificate path, key id and plugin name. The default signing key name is preceded by an asterisk. The key id and plugin name are used together to provide the information of the key identifier for the remote key and the plugin associated with it.
 
-### Delete two keys from signing key list
+### Remove a specified key from signing key list
```
yizha1, line 118: Is `specific` better?

specs/commandline/key.md
patrickzheng200: Yes, and as we discussed, we should introduce `notation generate-test` command to decouple `generate-test` with `cert`. This can be done between RC.4 and v1.0.

yizha1: I agree with Patrick

yizha1: There is a case that if the key to be removed is marked as default key. The output message should be improved as well.

yizha1 (commented):
Some comments left.

specs/commandline/key.md
FeynmanZhou: @patrickzheng200 I will create another issue to track this new feature request.

FeynmanZhou: @yizha1 @patrickzheng200  If a default key is removed from the signing key list, naturally, it means there is no default key anymore. So it's not necessary to tell users the key `xxx` is unmarked as default in the output msg.

specs/commandline/key.md
FeynmanZhou: No. I think `specified` points to a particular thing here. See https://www.askdifference.com/specific-vs-specified/

specs/commandline/key.md
patrickzheng200: @FeynmanZhou On a second thought, this actually is not a simple change. When deleting keys, we are deleting them by key name. From key name itself, we cannot tell if that's the key that been generated by `notation cert generate-test` command. In order to do it, we need to also make changes in notation-go library. Since we have agreed to introduce `notation generate-test` command in next release, I suggest we just make a minor change on the error message here in RC.4, something like:
`Removed test key <key_name> from Notation signing key list. The source key file is not deleted.`

yizha1: @patrickzheng200 When using `notation key list`, `key path` is also printed out, maybe we can print out the `The source <key path> is not deleted. `  If it also takes more time, I agree with your proposal. There is also `cert path` in the output message of `notation key list` and seems only useful for local testing keys. We can improve this later with the new `notation generate-test` command.

patrickzheng200: > @patrickzheng200 When using `notation key list`, `key path` is also printed out, maybe we can print out the `The source <key path> is not deleted. ` If it also takes more time, I agree with your proposal. There is also `cert path` in the output message of `notation key list` and seems only useful for local testing keys. We can improve this later with the new `notation generate-test` command.

@yizha1 Good idea, but I think it won't work on Deletion. Here's the reason: `notation key list` reads in the `signingkeys.json` file and print out the keys one by one. When deleting one or more keys, we delete by the key names (the core deletion logic is in notation-go). At the point of printing out, those keys have already been deleted from `signingkeys.json`. In other words, we cannot find the deleted key names anymore by just going through the file again. A workaround would be saving a copy of the whole key list before doing any Deletion. When printing out, we go through that pre-load list. But IMO, this isn't something that we should do just for some print outs. Therefore, I still suggest we use the message: `Removed test key <key_name> from Notation signing key list. The source key file is not deleted.` and introduce `notation generate-test` command in the next release to solve the issue completely.

FeynmanZhou: @patrickzheng200 Agree. We could improve it iteratively and aim to fix it by introducing a new command in v1.0.0 completely.

specs/commandline/key.md
toddysm: I believe, we had this discussion with either @iamsamirzon or @priteshbandi in the past. We need to be clear what is the meaning of `delete` and `remove`. It seems different people have different expectation. As a non-English speaker my expectation is similar to @shizhMSFT, i.e. I expect the key (and associated assets) to be gone completely. Though, the implementation right now is that the key is only "deleted" from the list of keys but not removed from the file system and associated certs are also not removed. As an experience, I would expect the command to be intelligent enough to give choices to do all of the above and not have separate command to "remove".

*Note:* We are talking about `key delete` here as far as I can see from @shizhMSFT comment.

specs/commandline/key.md
toddysm: Are we renaming the command from `delete` to `remove`? This will be a breaking change - isn't it?

specs/commandline/key.md
shizhMSFT: `remove` v.s.` delete`:
- `remove` is about removing something from a place and **can** be easily recovered. For instance, remove a song from a playlist.
- `delete` is about removing something from a place and **cannot** be easily recovered. For instance, delete a song from a music library.

In `notation key`, all keys are referenced. Therefore, `key remove` is more suitable. Removing the key from list will not delete the key from the storage.
In `notation cert`, all certs are stored as files. Therefore, `cert delete` is more suitable. Deleting the cert will delete the cert file from the disk and cannot be easily recovered.

However, as @toddysm mentioned, non-English speakers mix the term `remove` and `delete`. That's why I initially proposed to add `delete` as an alias to `remove`. Of course, we need to clearly state the key is **removed** in the response.

On the other hand, completely removing the key using `notation` is technically not possible unless a `delete key` protocol is introduced for plugins.

specs/commandline/key.md
```diff
@@ -115,10 +115,27 @@ notation key list
...
 
 ```shell
-notation key delete <key_name_1> <key_name_2>
+notation key remove <key_name>
 ```
 
-Upon successful execution, the names of deleted signing keys are printed out. Please be noted if default signing key is deleted, Notation will not automatically assign a new default signing key. User needs to update the default signing key explicitly.
+- Upon successful execution, the output message is printed out as below when removed a key created by notation for local testing purpose:
```
byronchien, line 124: grammar nit:
```suggestion
- Upon successful execution, the output message is printed out as below when removing a key created by notation for local testing purposes:
```

or can be reworded

```suggestion
- Upon successful removal of a local testing key created by notation, the output message is printed out as follows:
```

specs/commandline/key.md
```diff
@@ -115,10 +115,27 @@ notation key list
...
+
+```text
+Removed <key_name> from signing key list. You still need to delete the key file from <key_path> and delete the certificate from <cert_path>.
+```
+- Upon successful execution, the output message is printed out as below when removed a key associates with a KMS:
```
byronchien, line 129: grammar nit:
```suggestion
- Upon successful execution, the output message is printed out as below when removing a key associated with a KMS:
```

or can be reworded

```suggestion
- Upon successful removal of a key associated with a KMS, the output message is printed out as follows:
```

specs/commandline/key.md
```diff
@@ -115,10 +115,27 @@ notation key list
...
+Removed <key_name> from signing key list.
+```
+
+### Remove two keys from signing key list
+
+```shell
+notation key remove <key_name_1> <key_name_2>
+```
+
+Upon successful execution, the names of removed signing keys are printed out. Please be noted if default signing key is removed, Notation will not automatically assign a new default signing key. User needs to update the default signing key explicitly.
```
byronchien, line 141: do we need an example output for removing multiple keys? even if it's just
```shell
key_name_1
key_name_2
```

specs/commandline/key.md
```diff
@@ -115,10 +115,27 @@ notation key list
...
 ```shell
-notation key delete <key_name_1> <key_name_2>
+notation key remove <key_name>
 ```
 
-Upon successful execution, the names of deleted signing keys are printed out. Please be noted if default signing key is deleted, Notation will not automatically assign a new default signing key. User needs to update the default signing key explicitly.
+- Upon successful execution, the output message is printed out as below when removed a key created by notation for local testing purpose:
+
+```text
+Removed <key_name> from signing key list. You still need to delete the key file from <key_path> and delete the certificate from <cert_path>.
```
patrickzheng200, line 127:
```suggestion
Removed <key_name> from Notation signing key list. The source key file is not deleted.
```

specs/commandline/key.md
yizha1: My two cents:

It seems a practice that `add` and `remove` are a pair; `create` and `delete` are a pair. This is also another reason `notation key` and `notation cert` commands cause confusion.

Since we don't create any resources (key and certificate) except the `notation cert generate-test`. Using `add` and `remove` are more appropriate in most scenarios for `notation key` and `notation cert` commands. 

We could consider a different command for local key or certificate later.

yizha1: We can fix the error and output messages first to make current usage clear enough to our users.

FeynmanZhou: Per the discuss in the community meeting on Apr 3, we agreed that change `notation key delete` to `notation key remove` and use `delete` as an alias of `remove`.

specs/commandline/key.md
FeynmanZhou: Accepted.

specs/commandline/key.md
FeynmanZhou: Accepted and added a new output example for removing multiple keys.

specs/commandline/key.md
FeynmanZhou: Updated

specs/commandline/key.md
patrickzheng200: @FeynmanZhou I thought we agreed on this 2 weeks ago: https://github.com/notaryproject/notation/pull/606#discussion_r1154298416.

FeynmanZhou: @patrickzheng200 It's fine that we do not print the `<key_path>` in the output for this release. Currently, there is a [guide](https://github.com/notaryproject/notaryproject.dev/pull/164/files#diff-ebf28445e02ae412163e2fde43e76bee51e880abab8449ac8b4b863fec8d4354R29 ) to tell users to clean up those source key and cert files. We can determine if we should design a feature like `notation generate-test` or support local signing key scenario.

specs/commandline/key.md
```diff
@@ -115,10 +115,33 @@ notation key list
...
-Upon successful execution, the names of deleted signing keys are printed out. Please be noted if default signing key is deleted, Notation will not automatically assign a new default signing key. User needs to update the default signing key explicitly.
+- Upon successful removal of a local testing key created by notation, the output message is printed out as follows:
+
+```text
+Removed <key_name> from Notation signing key list. The source key file still exists locally.
+```
+- Upon successful removal of a key associated with a KMS, the output message is printed out as follows:
+
+```text
+Removed <key_name> from Notation signing key list.
```
patrickzheng200, line 132:
```suggestion
Removed <key_name> from Notation signing key list. The source key files still exist.
```
We should mention the source key files still exist for KMS scenario as well. Otherwise, users might misunderstand that the key is also removed from their KMS.

specs/commandline/key.md
```diff
@@ -115,10 +115,33 @@ notation key list
...
+### Remove two keys from Notation signing key list
+
+```shell
+notation key remove <key_name_1> <key_name_2>
+```
+
+Upon successful execution, the output message is printed out as follows. Please be noted if default signing key is removed, Notation will not automatically assign a new default signing key. User needs to update the default signing key explicitly.
+
+```text
+Removed the following keys from Notation signing key list. The source key files still exist locally.
```
patrickzheng200, line 144:
```suggestion
Removed the following keys from Notation signing key list. The source key files still exist.
```
When deleting a key, telling it's a local key or a key stored in a KMS is as tricky as printing out the local key path. Actually, if we could tell a key is a local key, then we are able to print its path out. Both cases need changes on notation CLI + notation-go library.

specs/commandline/key.md
```diff
@@ -115,10 +115,33 @@ notation key list
...
 ```shell
-notation key delete <key_name_1> <key_name_2>
+notation key remove <key_name>
 ```
 
-Upon successful execution, the names of deleted signing keys are printed out. Please be noted if default signing key is deleted, Notation will not automatically assign a new default signing key. User needs to update the default signing key explicitly.
+- Upon successful removal of a local testing key created by notation, the output message is printed out as follows:
+
+```text
+Removed <key_name> from Notation signing key list. The source key file still exists locally.
```
patrickzheng200, line 127:
```suggestion
Removed <key_name> from Notation signing key list. The source key file still exists.
```
same here.

specs/commandline/key.md
FeynmanZhou: Got it. Keeping minimum changes on the CLI outputs is acceptable without adding additional logic in notation-go lib this time. We could improve this experience thoroughly after v1.0.0 by proposing a feature to clean up local testing keys.

patrickzheng200 (approved):
LGTM

patrickzheng200 (dismissed):
LGTM

specs/commandline/key.md
```diff
@@ -115,10 +115,33 @@ notation key list
...
+### Remove a specified key from Notation signing key list
+
+```shell
+notation key delete <key_name>
+```
+
+- Upon successful removal of a local testing key created by notation, the output message is printed out as follows:
+
+```text
+Removed <key_name> from Notation signing key list. The source key file still exists.
```
yizha1, line 127: If the source key is a remote key, it may not be a key file? How about "The source key still exists."

specs/commandline/key.md
```diff
@@ -115,10 +115,33 @@ notation key list
...
+```
+- Upon successful removal of a key associated with a KMS, the output message is printed out as follows:
+
+```text
+Removed <key_name> from Notation signing key list. The source key files still exist.
```
yizha1, line 132: If the source key is a remote key, it may not be a file. How about "The source key still exists."  Maybe we can use this for local key as well.

specs/commandline/key.md
yizha1, line 127: Do we need to say something if the removed key is a default key?

FeynmanZhou: @yizha1  I think it's not necessary to mention that when deleting a key whether it is a default key or not. https://github.com/notaryproject/notation/pull/606#discussion_r1154195936

patrickzheng200 (approved):
LGTM

yizha1 (approved):
LGTM but IANAM

specs/commandline/certificate.md
```diff
@@ -208,7 +208,17 @@ A prompt is showed asking user to confirm the deletion. Upon successful deletion
 notation certificate delete --type <type> --store <name> <cert_fileName>
 ```
 
-A prompt is showed asking user to confirm the deletion. Upon successful deletion, the specific certificate is deleted in trust store named `<name>` of type `<type>`. If deletion fails, an error message with specific reasons is printed out.
+A prompt is showed asking user to confirm the deletion. Upon successful deletion, the specific certificate is deleted from the trust store named `<name>` of type `<type>`. The output message is printed out as following:
```
JeyJeyGao, line 211:
```suggestion
A prompt is displayed, asking the user to confirm the deletion. Upon successful deletion, the specific certificate is deleted from the trust store named `<name>` of type `<type>`. The output message is printed out as following:
```

priteshbandi (approved):
LGTM

FeynmanZhou:
Hi @notaryproject/notaryproject-notation-maintainers , I resolved @JeyJeyGao 's comment. Could you pls review it again? Thanks

patrickzheng200 (approved):
LGTM

FeynmanZhou:
@priteshbandi There was a small update after your last approval. Could you re-approve this PR? Thanks!

priteshbandi (approved):
LGTM

PR #613, opened 2023-04-04
feat: introduce experimental feature switch

### LINKED ISSUE #614: Support experimental feature

Notation CLI

Experimental features for CLI commands refer to functionality that are not yet part of the base version of a CLI tool but can be accessed by enabling the appropriate options. The purpose of using experimental features is to enable users to try out new functionality or improvements, and to provide feedback for further development or refinement. It can also help identify issues or bugs that may not have been caught during testing. They may not work as expected and are not recommended for use in production or critical systems.

The proposal is to use environment variable to enable experimental features for Notation CLI. The proposed name is `NOTATION_EXPERIMENTAL`. 

**What is default behavior?**
By default, the experimental features are disabled. The related CLI command flags are not shown to users from Notation CLI help command. 

**How to enable experimental features?**
To enable experimental features, users set `NOTATION_EXPERIMENTAL_CLI` in their environment, for example `export NOTATION_EXPERIMENTAL_CLI=1`. 

**What will happen after experimental features are enabled?**
The CLI flags related to experimental features are shown to users and marked as `[Experimental]` in the description in the help document. 
When using experimental features, the CLI output should warn the users that these are experimental features, like `Caution: This feature is experimental and may not be fully tested or completed and may be deprecated. Report any issues to "https://github/notaryproject/notation` for further improvement."

**How do users know there are experimental features and how to use them?**
The release note or readme.md should give users guidance on how to enable/disable experimental features, other documents are also required to explain to the users on how to use experimental features.

Another alternative is that by default the experimental features are visible to users, but in order to use them, user need to set the corresponding environment variable.

### Any additional context?

_No response_

qweeah:
This PR introduces an environmental variable, `NOTATION_EXPERIMENTAL`, as the switch for experimental features
- If `NOTATION_EXPERIMENTAL` set to `1`, then experimental features are enabled. All experimental features can be used with a prompt in the stderr output:
```console
Caution: This feature is experimental and may not be fully tested or completed and may be deprecated. Report any issues to "https://github/notaryproject/notation"
```

- Otherwise experimental features are disabled, using an experimental command will show
```console
Error: "<COMMAND>" been marked as experimental and not enabled by default. To use it, please set NOTATION_EXPERIMENTAL=1 in your environment
```
or 
```console
Error: flags(s) --<FLAG1>,--<FLAG2> in "<COMMAND>" been marked as experimental and not enabled by default. To use it, please set NOTATION_EXPERIMENTAL=1 in your environment
```
Resolves #614.

### DIFF
--- a/internal/experimental/experimental.go
+++ b/internal/experimental/experimental.go
@@ -1,0 +1,77 @@
+package experimental
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+const (
+	envName = "NOTATION_EXPERIMENTAL"
+	enabled = "1"
+)
+
+// IsDisabled determines whether an experimental feature is disabled.
+func IsDisabled() bool {
+	return os.Getenv(envName) != enabled
+}
+
+// Error returns an error for a disabled experimental feature.
+func Error(description string) error {
+	return fmt.Errorf("%s been marked as experimental and not enabled by default. To use it, please set %s=%s in your environment", description, envName, enabled)
+}
+
+// CheckCommandAndWarn checks whether an experimental command can be run.
+func CheckCommandAndWarn(cmd *cobra.Command, args []string) error {
+	if err := Check(func() (string, bool) {
+		return fmt.Sprintf("%q", cmd.CommandPath()), true
+	}); err != nil {
+		return err
+	}
+	return Warn()
+}
+
+// FlagsChecker returns the function which checks whether flags are not set.
+func FlagsChecker(flags ...string) func(*cobra.Command, []string) error {
+	return func(cmd *cobra.Command, _ []string) error {
+		if err := Check(func() (string, bool) {
+			var changedFlags []string
+			flagSet := cmd.Flags()
+			for _, flag := range flags {
+				flagSet.MarkHidden(flag)
+				if flagSet.Changed(flag) {
+					changedFlags = append(changedFlags, "--"+flag)
+				}
+			}
+			if len(changedFlags) == 0 {
+				// no experimental flag used
+				return "", false
+			}
+			return fmt.Sprintf("flag(s) %s in %s", strings.Join(changedFlags, ","), cmd.CommandPath()), true
+		}); err != nil {
+			return err
+		}
+		return Warn()
+	}
+}
+
+// Check checks whether a feature can be used.
+func Check(doCheck func() (feature string, isExperimental bool)) error {
+	if IsDisabled() {
+		feature, isExperimental := doCheck()
+		if isExperimental {
+			// feature is experimental and disabled
+			return Error(feature)
+		}
+		return nil
+	}
+	return nil
+}
+
+// Warn prints a warning message for using the experimental feature.
+func Warn() error {
+	_, err := fmt.Fprintf(os.Stderr, "Caution: This feature is experimental and may not be fully tested or completed and may be deprecated. Report any issues to \"https://github/notaryproject/notation\"\n")
+	return err
+}
--- a/internal/experimental/experimental.go
+++ b/internal/experimental/experimental.go
@@ -33,7 +33,7 @@ func CheckCommandAndWarn(cmd *cobra.Command, args []string) error {
 	return Warn()
 }
 
-// CheckCommandAndWarn checks whether experimental flags can be run.
+// CheckFlagsAndWarn checks whether experimental flags can be run.
 func CheckFlagsAndWarn(cmd *cobra.Command, flags ...string) error {
 	if err := Check(func() (string, bool) {
 		var changedFlags []string
--- a/internal/experimental/experimental.go
+++ b/internal/experimental/experimental.go
@@ -20,11 +20,11 @@ func IsDisabled() bool {
 
 // Error returns an error for a disabled experimental feature.
 func Error(description string) error {
-	return fmt.Errorf("%s been marked as experimental and not enabled by default. To use it, please set %s=%s in your environment", description, envName, enabled)
+	return fmt.Errorf("%s is experimental and not enabled by default. To use, please set %s=%s environment variable", description, envName, enabled)
 }
 
 // CheckCommandAndWarn checks whether an experimental command can be run.
-func CheckCommandAndWarn(cmd *cobra.Command, args []string) error {
+func CheckCommandAndWarn(cmd *cobra.Command, _ []string) error {
 	if err := Check(func() (string, bool) {
 		return fmt.Sprintf("%q", cmd.CommandPath()), true
 	}); err != nil {
--- a/internal/experimental/experimental.go
+++ b/internal/experimental/experimental.go
@@ -33,28 +33,26 @@ func CheckCommandAndWarn(cmd *cobra.Command, args []string) error {
 	return Warn()
 }
 
-// FlagsChecker returns the function which checks whether flags are not set.
-func FlagsChecker(flags ...string) func(*cobra.Command, []string) error {
-	return func(cmd *cobra.Command, _ []string) error {
-		if err := Check(func() (string, bool) {
-			var changedFlags []string
-			flagSet := cmd.Flags()
-			for _, flag := range flags {
-				flagSet.MarkHidden(flag)
-				if flagSet.Changed(flag) {
-					changedFlags = append(changedFlags, "--"+flag)
-				}
+// CheckCommandAndWarn checks whether experimental flags can be run.
+func CheckFlagsAndWarn(cmd *cobra.Command, flags ...string) error {
+	if err := Check(func() (string, bool) {
+		var changedFlags []string
+		flagSet := cmd.Flags()
+		for _, flag := range flags {
+			flagSet.MarkHidden(flag)
+			if flagSet.Changed(flag) {
+				changedFlags = append(changedFlags, "--"+flag)
 			}
-			if len(changedFlags) == 0 {
-				// no experimental flag used
-				return "", false
-			}
-			return fmt.Sprintf("flag(s) %s in %s", strings.Join(changedFlags, ","), cmd.CommandPath()), true
-		}); err != nil {
-			return err
 		}
-		return Warn()
+		if len(changedFlags) == 0 {
+			// no experimental flag used
+			return "", false
+		}
+		return fmt.Sprintf("flag(s) %s in %q", strings.Join(changedFlags, ","), cmd.CommandPath()), true
+	}); err != nil {
+		return err
 	}
+	return Warn()
 }
 
 // Check checks whether a feature can be used.


specs/notation-cli.md
```diff
@@ -13,7 +13,7 @@ This spec contains reference information on using notation commands. Each comman
 | [login](./commandline/login.md)             | Login to registries                                                    |
 | [logout](./commandline/logout.md)           | Log out from the logged in registries                                  |
 | [plugin](./commandline/plugin.md)           | Manage plugins                                                         |
-| [policy](./commandline/policy.md)           | [Preview] Manage trust policy configuration for signature verification |
+| [policy](./commandline/policy.md)           | [Experimental] Manage trust policy configuration for signature verification |
```
shizhMSFT, line 16: nit: format table

specs/notation-cli.md
```diff
@@ -34,7 +34,7 @@ Available Commands:
   login       Login to registry
   logout      Log out from the logged in registries
   plugin      Manage plugins
-  policy      [Preview] Manage trust policy configuration for signature verification
+  policy      Manage trust policy configuration for signature verification
```
shizhMSFT, line 37: It seems we still need to keep the `[Experimental]` label.

cmd/notation/policy/cmd.go
```diff
@@ -1,12 +1,20 @@
...
+	"github.com/notaryproject/notation/internal/experimental"
+	"github.com/spf13/cobra"
+)
 
 func Cmd() *cobra.Command {
 	command := &cobra.Command{
-		Use:   "policy [command]",
-		Short: "[Preview] Manage trust policy configuration",
-		Long:  "[Preview] Manage trust policy configuration for signature verification.",
+		Use:     "policy [command]",
```
patrickzheng200, line 10: `$ notation policy -h` prints out two usages:
```
Usage:
  notation policy [command] [flags]
  notation policy [command]
```
We only need the first one.

internal/experimental/experimental.go
```diff
@@ -0,0 +1,51 @@
...
+	return os.Getenv(envName) != enabled
+}
+
+// Error returns an error for a disabled experimental feature.
+func Error(description string) error {
+	return fmt.Errorf("%s is marked as experimental and not enabled by default. To use it, please set %s=%s in your environment", description, envName, enabled)
+}
+
+// CheckCommand checks whether a experimental command can be run.
+func CheckCommand(cmd *cobra.Command, args []string) error {
```
patrickzheng200, line 26: Looks like this function only checks if a `command` is marked as experimental. How about flags? For example, under the `sign` command, there's the experimental flag `--signature-manifest`.

internal/experimental/experimental.go
```diff
@@ -0,0 +1,51 @@
...
+func IsDisabled() bool {
+	return os.Getenv(envName) != enabled
+}
+
+// Error returns an error for a disabled experimental feature.
+func Error(description string) error {
+	return fmt.Errorf("%s is marked as experimental and not enabled by default. To use it, please set %s=%s in your environment", description, envName, enabled)
+}
+
+// CheckCommand checks whether a experimental command can be run.
```
patrickzheng200, line 25: nit:
```suggestion
// CheckCommand checks whether an experimental command can be run.
```

internal/experimental/experimental.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+const (
+	envName = "NOTATION_EXPERIMENTAL"
+	enabled = "1"
+)
+
+// IsDisabled determines whether a experimental feature is disabled.
```
patrickzheng200, line 15: nit:
```suggestion
// IsDisabled determines whether an experimental feature is disabled.
```

internal/experimental/experimental.go
qweeah: This is a utility wrapper function of `func Check(cmd *cobra.Command, args []string, doCheck func(cmd *cobra.Command, args []string) (feature string, isExperimental bool)) error`, it checks whether a experimental command can be run.

If sign via an artifact manifest is experimental, you need to call `check` directly in sign command code like:
```golang
experimental.Check(cmd, args, func(cmd *cobra.Command, args []string) (string, bool)) {
    sig, err := cmd.Flags().GetString("signature-manifest")
    if err != nil || sig != "artifact" {
        // artifact not used
        return "", false
    }
    return "sign via OCI artifact", true
})
```

specs/notation-cli.md
```diff
@@ -34,7 +34,7 @@ Available Commands:
   login       Login to registry
   logout      Log out from the logged in registries
   plugin      Manage plugins
-  policy      [Preview] Manage trust policy configuration for signature verification
+  policy      [Experimental] Manage trust policy configuration for signature verification
```
yizha1, line 37: How about mark `Experimental` for `notation policy` subcommands, like `show` and `import`?

yizha1 (commented):
One comment on sub-commands.

internal/experimental/experimental.go
```diff
@@ -0,0 +1,77 @@
...
+		return err
+	}
+	return Warn()
+}
+
+// FlagsChecker returns the function which checks whether flags are not set.
+func FlagsChecker(flags ...string) func(cmd *cobra.Command, args []string) error {
+	return func(cmd *cobra.Command, args []string) error {
+		if err := Check(func() (string, bool) {
+			var changed []string
```
patrickzheng200, line 40: Can we rename it to something else? Semantically, `changed` is kind of vague.

internal/experimental/experimental.go
```diff
@@ -0,0 +1,77 @@
...
+			for _, flag := range flags {
+				fs := cmd.Flags()
```
patrickzheng200, line 42:
```suggestion
				flagSet := cmd.Flags()
```
usually, `fs` refers to package `io/fs` across notation.

patrickzheng200, line 42: Should we move this line out of the for loop?

qweeah:
@priteshbandi I have removed policy changes, can you help review the implementation

@yizha1 @FeynmanZhou @iamsamirzon @toddysm can you help review the user experience as well as the wording for the output in the PR description, thanks!

internal/experimental/experimental.go
```diff
@@ -0,0 +1,77 @@
...
+				fs.MarkHidden(flag)
+				if fs.Changed(flag) {
+					changed = append(changed, "--"+flag)
+				}
+			}
+			if len(changed) == 0 {
+				// no experimental flag used
+				return "", false
+			}
+			return fmt.Sprintf("using %q in %q", strings.Join(changed, ","), cmd.CommandPath()), true
```
patrickzheng200, line 52:
```suggestion
			return fmt.Sprintf("using flags %q in %q", strings.Join(changed, ","), cmd.CommandPath()), true
```

internal/experimental/experimental.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+// IsDisabled determines whether an experimental feature is disabled.
+func IsDisabled() bool {
+	return os.Getenv(envName) != enabled
+}
+
+// Error returns an error for a disabled experimental feature.
+func Error(description string) error {
+	return fmt.Errorf("%q is marked as experimental and not enabled by default. To use it, please set %s=%s in your environment", description, envName, enabled)
```
patrickzheng200, line 23: Had a try on the command on Linux, got the following error message:
`Error: "using flags \"--flag-a,--flag-b\" in \"notation sign\"" is marked as experimental and not enabled by default. To use it, please set NOTATION_EXPERIMENTAL=1 in your environment`
The above error message is not very easy to read through. How about:
```suggestion
	return fmt.Errorf("%s been marked as experimental and not enabled by default. To use it, please set %s=%s in your environment", description, envName, enabled)
```

internal/experimental/experimental.go
```diff
@@ -0,0 +1,77 @@
...
+				flagSet.MarkHidden(flag)
+				if flagSet.Changed(flag) {
+					changedFlags = append(changedFlags, "--"+flag)
+				}
+			}
+			if len(changedFlags) == 0 {
+				// no experimental flag used
+				return "", false
+			}
+			return fmt.Sprintf("using flags %q in %q", strings.Join(changedFlags, ","), cmd.CommandPath()), true
```
patrickzheng200, line 52:
```suggestion
			return fmt.Sprintf("flag(s) %s in %q", strings.Join(changedFlags, ","), cmd.CommandPath()), true
```

internal/experimental/experimental.go
```diff
@@ -0,0 +1,77 @@
...
+		return cmd.CommandPath(), true
+	}); err != nil {
+		return err
+	}
+	return Warn()
+}
+
+// FlagsChecker returns the function which checks whether flags are not set.
+func FlagsChecker(flags ...string) func(cmd *cobra.Command, args []string) error {
+	return func(cmd *cobra.Command, args []string) error {
```
patrickzheng200, line 38: can remove `args`, as it's not used in the function.
```suggestion
func FlagsChecker(flags ...string) func(cmd *cobra.Command) error {
	return func(cmd *cobra.Command) error {
```

internal/experimental/experimental.go
patrickzheng200: Wit the above changes, the error message becomes:
`Error: flag(s) --flag-a,--flag-b in "notation sign" been marked as experimental and not enabled by default. To use it, please set NOTATION_EXPERIMENTAL=1 in your environment`

internal/experimental/experimental.go
```diff
@@ -0,0 +1,77 @@
...
+
+// Error returns an error for a disabled experimental feature.
+func Error(description string) error {
+	return fmt.Errorf("%q been marked as experimental and not enabled by default. To use it, please set %s=%s in your environment", description, envName, enabled)
+}
+
+// CheckCommandAndWarn checks whether an experimental command can be run.
+func CheckCommandAndWarn(cmd *cobra.Command, args []string) error {
+	if err := Check(func() (string, bool) {
+		return cmd.CommandPath(), true
```
patrickzheng200, line 29:
```suggestion
		return fmt.Sprintf("%q",cmd.CommandPath()), true
```

patrickzheng200 (dismissed):
LGTM

internal/experimental/experimental.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+const (
+	envName = "NOTATION_EXPERIMENTAL"
+	enabled = "1"
+)
+
+// IsDisabled determines whether an experimental feature is disabled.
+func IsDisabled() bool {
+	return os.Getenv(envName) != enabled
```
priteshbandi, line 18: should we trim the white spaces here, for both key and value?

internal/experimental/experimental.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// Error returns an error for a disabled experimental feature.
+func Error(description string) error {
+	return fmt.Errorf("%s been marked as experimental and not enabled by default. To use it, please set %s=%s in your environment", description, envName, enabled)
```
priteshbandi, line 23:
```suggestion
	return fmt.Errorf("%s is experimental and not enabled by default. To use, please set %s=%s environment variable", description, envName, enabled)
```

internal/experimental/experimental.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// CheckCommandAndWarn checks whether an experimental command can be run.
+func CheckCommandAndWarn(cmd *cobra.Command, args []string) error {
```
priteshbandi, line 27: `args` is unused variable.

internal/experimental/experimental.go
qweeah: For what purpose do we need to trim key and value?

internal/experimental/experimental.go
qweeah: make it `_` so user can use it via cobra command pre-run handler like
```golang
func xxxCmd() *cobra. Command {
...
		PreRunE: experimental.CheckCommandAndWarn,
...
```

internal/experimental/experimental.go
qweeah: Changed accordingly. There is a case when multiple experimental flags the warning can be
> flag(s) --foo,--bar,--baz in "notation test" is experimental and not enabled by default. To use, please set %s=%s environment variable

I am okay with the `is` here, @patrickzheng200 what do you think.

patrickzheng200: > Changed accordingly. There is a case when multiple experimental flags the warning can be
> 
> > flag(s) --foo,--bar,--baz in "notation test" is experimental and not enabled by default. To use, please set %s=%s environment variable
> 
> I am okay with the `is` here, @patrickzheng200 what do you think.

Yeah, we could have multiple flags marked as experimental, in this case, using `is` here is not very accurate.

internal/experimental/experimental.go
priteshbandi: In scenario where user calls following command to set env variable.
`export NOTATION_EXPERIMENTAL=" 1"`

Its an edge case, I am fine to ignore this.

priteshbandi (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #617, opened 2023-04-04
feat: improve plugin installation UX

duffney:
Adds an `install` and `remove` subcommand to the `notation plugin` command. The install command takes a plugin binary and copies it to the plugin directory. If the plugin directory doesn't exist, it is created, and the plugin copied. If an exist plugin exists, the version is compared and copied when the current version is less than the version of the plugin binary provided as an argument. A `--force` flag is available to overwrite this logic and copy the plugin binary regardless of the existing plugin version. The remove command removes the plugin specific directory and its contents.


cmd/notation/plugin.go
```diff
@@ -62,3 +108,169 @@ func listPlugins(command *cobra.Command) error {
...
+		}
+
+		// copy plugin, if new plugin version is greater than current plugin version
+		if newSemVersion.GreaterThan(currentVersion) {
+			var confirm string
+
+			fmt.Printf("Detected new version %s. Current version is %s.\nDo you want to overwrite the plugin %s? [y/N]: ", newSemVersion.String(), currentVersion.String(), pluginName)
+			fmt.Scanln(&confirm)
+
+			if strings.ToLower(confirm) != "y" {
```
qweeah, line 207: Try do the confirmed prompting with existing function `cmdutil.AskForConfirmation`

qweeah (changes requested):

cmd/notation/plugin.go
```diff
@@ -62,3 +108,169 @@ func listPlugins(command *cobra.Command) error {
 	}
 	return tw.Flush()
 }
+
+func installPlugin(command *cobra.Command, args []string, force bool) error {
+	if len(args) != 1 {
+		return errors.New("missing plugin package")
+	}
+
+	pluginPath := args[0]
```
patrickzheng200, line 117: We should check length of `args` and assign `args[0]` in `&cobra.command{Args:xxx}`.

cmd/notation/plugin.go
```diff
@@ -62,3 +108,169 @@ func listPlugins(command *cobra.Command) error {
...
+	}
+
+	var newPlugin map[string]interface{}
+	err = json.Unmarshal([]byte(output), &newPlugin)
+	if err != nil {
+		return err
+	}
+
+	pluginName := newPlugin["name"].(string)
+	newPluginVersion := newPlugin["version"].(string)
```
patrickzheng200, line 142: As it's type assertion on map values, we need:
```suggestion
	pluginName, ok := newPlugin["name"].(string)
	if !ok {
	    xxx
	}
	newPluginVersion, ok := newPlugin["version"].(string)
	if !ok {
	    xxx
	}
```

cmd/notation/plugin.go
```diff
@@ -62,3 +108,169 @@ func listPlugins(command *cobra.Command) error {
...
+	pluginDir, err := dir.PluginFS().SysPath(pluginName)
+	if err != nil {
+		panic(err)
+	}
+
+	// Check if plugin directory exists
+	_, err = os.Stat(pluginDir)
+
+	// create the directory, if not exist
+	if os.IsNotExist(err) {
```
patrickzheng200, line 158: Are we handling the case where `err != nil` AND `os.IsNotExist(err) == false` ?

cmd/notation/plugin.go
```diff
@@ -62,3 +108,169 @@ func listPlugins(command *cobra.Command) error {
...
+		if err := os.MkdirAll(pluginDir, 0755); err != nil {
+			return err
+		}
+	}
+
+	// Check if plugin exists
+	_, err = os.Stat(pluginDir + "/" + pluginBinary)
+
+	// copy plugin, if not exist
+	if os.IsNotExist(err) {
```
patrickzheng200, line 168: Same here. Are we handling the case where `err != nil` AND `os.IsNotExist(err) == false` ?

cmd/notation/plugin.go
```diff
@@ -62,3 +108,169 @@ func listPlugins(command *cobra.Command) error {
...
+		copyPlugin(pluginPath, pluginDir+"/"+pluginBinary)
+	}
+
+	// overwrite plugin, if force flag is set
+	if err == nil && force {
```
patrickzheng200, line 173: if `err != nil` is handled above, then it's just `if force {xxx}` here.

cmd/notation/plugin.go
```diff
@@ -62,3 +108,169 @@ func listPlugins(command *cobra.Command) error {
...
+
+	return nil
+}
+
+func removePlugin(command *cobra.Command, args []string) error {
+	if len(args) != 1 {
+		return errors.New("missing plugin name")
+	}
+
+	pluginName := args[0]
```
patrickzheng200, line 230: Same here. They should be handled in cobra `Args`.

cmd/notation/plugin.go
```diff
@@ -62,3 +108,169 @@ func listPlugins(command *cobra.Command) error {
...
+
+	// get plugin directory
+	pluginDir, err := dir.PluginFS().SysPath(pluginName)
+	if err != nil {
+		return err
+	}
+
+	// Check if plugin directory exists
+	_, err = os.Stat(pluginDir)
+	if os.IsNotExist(err) {
```
patrickzheng200, line 240: Same here. We should handle any `err != nil` cases.

cmd/notation/plugin.go
```diff
@@ -62,3 +108,169 @@ func listPlugins(command *cobra.Command) error {
...
+		return err
+	}
+	defer destFile.Close()
+
+	_, err = io.Copy(destFile, sourceFile)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
patrickzheng200, line 276: Can we try to use existing function `CopyToDir()` here?https://github.com/notaryproject/notation/blob/0ec3b3dd7166e680056b14b23847a03976edcfea/internal/osutil/file.go#L43

yizha1:
@priteshbandi @shizhMSFT @FeynmanZhou I suggest we also put this new subcommand `install` as experimental command since `install` command takes a plugin binary only, how can users get to know what kind of file should be used. So, it will be good to let users try this command first and provide feedback.

cmd/notation/plugin.go
```diff
@@ -36,6 +44,44 @@ Example - List installed Notation plugins:
 	}
 }
 
+func pluginInstallCommand() *cobra.Command {
+	var force bool
```
shizhMSFT, line 48: Should we make a struct instead of a single variable?

cmd/notation/plugin.go
```diff
@@ -36,6 +44,44 @@ Example - List installed Notation plugins:
...
+
+	cmd := &cobra.Command{
+		Use:     "install [flags] <plugin package>",
+		Aliases: []string{"add"},
+		Short:   "Install a plugin",
+		Long: `Install a plugin
+
+		Example - Install a Notation plugin:
+			notation plugin install <plugin package>
+`,
```
shizhMSFT, line 58: Indentation should be removed. The output should not follow the code indentitation.
```suggestion
		Long: `Install a plugin

Example - Install a Notation plugin:
	notation plugin install <plugin package>
`,
```

cmd/notation/plugin.go
```diff
@@ -36,6 +44,44 @@ Example - List installed Notation plugins:
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return installPlugin(cmd, args, force)
+		},
+	}
+
+	cmd.Flags().BoolVarP(&force, "force", "f", false, "Overwrite existing plugin files without prompting")
```
shizhMSFT, line 64: nit: consistent description
```suggestion
	cmd.Flags().BoolVarP(&force, "force", "f", false, "overwrite existing plugin files without prompting")
```

cmd/notation/plugin.go
```diff
@@ -36,6 +44,44 @@ Example - List installed Notation plugins:
...
+
+func pluginRemoveCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:     "remove [flags] <plugin>",
+		Aliases: []string{"rm", "uninstall", "delete"},
+		Short:   "Remove a plugin",
+		Long: `Remove a plugin
+
+		Example - Remove a Notation plugin:
+			notation plugin remove <plugin>
```
shizhMSFT, line 77: same here

cmd/notation/plugin.go
```diff
@@ -62,3 +108,169 @@ func listPlugins(command *cobra.Command) error {
...
+	var pluginBinary string
+	// if plugin contains a file path split and select the last element in the array
+	if strings.Contains(pluginPath, "/") {
+		pluginBinary = strings.Split(pluginPath, "/")[len(strings.Split(pluginPath, "/"))-1]
+	} else {
+		pluginBinary = pluginPath
+	}
+
+	// get plugin metadata
+	cmd := exec.Command("./"+pluginPath, "get-plugin-metadata")
```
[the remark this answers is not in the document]
shizhMSFT: Note: There is no security concerns since golang 1.19

shizhMSFT, line 128: You can try [plugin.NewCLIPlugin](https://pkg.go.dev/github.com/notaryproject/notation-go/plugin#NewCLIPlugin) instead of manually call the `get-plugin-metadata` and parse the responses according the protocol.

cmd/notation/plugin.go
```diff
@@ -62,3 +108,169 @@ func listPlugins(command *cobra.Command) error {
...
+		return errors.New("plugin does not exist")
+	}
+
+	// remove plugin directory
+	err = os.RemoveAll(pluginDir)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
shizhMSFT, line 250: nit:
```suggestion
	return os.RemoveAll(pluginDir)
```

cmd/notation/plugin.go
```diff
@@ -62,3 +108,169 @@ func listPlugins(command *cobra.Command) error {
...
+	sourceFile, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+
+	sourceFileInfo, err := sourceFile.Stat()
+	if err != nil {
+		return err
+	}
+	fileMode := sourceFileInfo.Mode()
```
shizhMSFT, line 263: Question: Why should we preserve the file mode?

shizhMSFT:
@priteshbandi I'd like to see if `notation plugin install` from a tar gz file. There is no security risk if we require the file structure in the tar ball to be flatten and only allow regular files. No symlinks and hard links of course.

cmd/notation/plugin.go
patrickzheng200, line 142: We can reuse the [GetMetadata](https://pkg.go.dev/github.com/notaryproject/notation-go/plugin#CLIPlugin.GetMetadata) to get the name and version of plugin.

cmd/notation/plugin.go
Duffney: Without the file mode the file cannot execute. I suppose an alternative approach would be to ensure the copied plugin binary has execute permissions.

cmd/notation/plugin.go
Duffney: I'm using the exec func to run [GetMetadata](https://github.com/notaryproject/notation/blob/26187094a03643e0b283d40f34e39cf599df75dd/cmd/notation/plugin.go#L128) then using Unmarshal to convert the JSON results to a Go object. If there is a better way to do that, please lmk. :)

cmd/notation/plugin.go
```diff
@@ -62,3 +108,169 @@ func listPlugins(command *cobra.Command) error {
...
+	} else {
+		pluginBinary = pluginPath
+	}
+
+	// get plugin metadata
+	cmd := exec.Command("./"+pluginPath, "get-plugin-metadata")
+
+	output, err := cmd.Output()
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
Duffney: Great catch, thank you! :)

cmd/notation/plugin.go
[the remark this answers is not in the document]
Duffney: Ah, so instead of executing the cmd and using Unmarshal to get the data use the [proto.GetMetadataResponse](https://github.com/notaryproject/notation/blob/main/cmd/notation/plugin.go#L53) instead? I'll start working on that, thanks for the feedback. :)

cmd/notation/plugin.go
```diff
@@ -62,3 +108,169 @@ func listPlugins(command *cobra.Command) error {
...
+
+	pluginPath := args[0]
+
+	var pluginBinary string
+	// if plugin contains a file path split and select the last element in the array
+	if strings.Contains(pluginPath, "/") {
+		pluginBinary = strings.Split(pluginPath, "/")[len(strings.Split(pluginPath, "/"))-1]
+	} else {
+		pluginBinary = pluginPath
+	}
```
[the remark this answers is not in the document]
Duffney: Adding this requires the user execute the install command like this './notation plugin install ./notation-azure-kv'. If the `./` isn't added to the front of the plugin binary the exec command throws and error that the file cannot be found.

cmd/notation/plugin.go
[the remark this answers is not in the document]
Duffney: It does work but also requires the user put `./` in from of the binary or path. I've gone ahead and make the changes. :)

https://github.com/notaryproject/notation/pull/617#discussion_r1169000919

cmd/notation/plugin.go
```diff
@@ -62,3 +108,169 @@ func listPlugins(command *cobra.Command) error {
...
+		}
+	}
+
+	// Check if plugin exists
+	_, err = os.Stat(pluginDir + "/" + pluginBinary)
+
+	// copy plugin, if not exist
+	if os.IsNotExist(err) {
+		copyPlugin(pluginPath, pluginDir+"/"+pluginBinary)
+	}
```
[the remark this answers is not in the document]
Duffney: I changed `pluginPath` to `pluginSrcPath` and added a new variable `pluginDestPath` for the location to reduced the duplicate code. :)

cmd/notation/plugin.go
```diff
@@ -62,3 +108,169 @@ func listPlugins(command *cobra.Command) error {
...
+	pluginDir, err := dir.PluginFS().SysPath(pluginName)
+	if err != nil {
+		return err
+	}
+
+	// Check if plugin directory exists
+	_, err = os.Stat(pluginDir)
+	if os.IsNotExist(err) {
+		return errors.New("plugin does not exist")
+	}
```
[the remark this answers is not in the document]
Duffney: @priteshbandi good point. What do you think about a helper called `checkPluginDir` that handles this logic and is used in the install and remove funcs?

cmd/notation/plugin.go
Duffney: Do you have an example? I'm not sure how to implement it. Thanks @patrickzheng200 :)

yizha1:
> > @priteshbandi @shizhMSFT @FeynmanZhou I suggest we also put this new subcommand `install` as experimental command since `install` command takes a plugin binary only, how can users get to know what kind of file should be used. So, it will be good to let users try this command first and provide feedback.
> 
> Is there anything I need to do to mark it as experimental?

@Duffney  Sorry. Please ignore my comments for now.

cmd/notation/plugin.go
```diff
@@ -62,3 +109,160 @@ func listPlugins(command *cobra.Command) error {
...
+
+		// check if new plugin version is greater than current plugin version
+		currentPluginVersion := currentPlugin["version"].(string)
+		currentVersion, err := semver.NewVersion(currentPluginVersion)
+		if err != nil {
+			return err
+		}
+
+		// copy plugin, if new plugin version is greater than current plugin version
+		if newSemVersion.GreaterThan(currentVersion) {
```
JeyJeyGao, line 198: can we use `golang.org/x/mod/semver` instead of "github.com/Masterminds/semver"? 
Because `golang.org/x/mod/semver` was used by notation-go, we will not add an additional dependency.

cmd/notation/plugin.go
```diff
@@ -62,3 +109,160 @@ func listPlugins(command *cobra.Command) error {
...
+
+	pluginDestPath := pluginDir + "/" + pluginBinary
+
+	// Check if plugin exists
+	_, err = os.Stat(pluginDestPath)
+
+	// copy plugin, if not exist
+	if os.IsNotExist(err) {
+		copyPlugin(pluginSrcPath, pluginDestPath)
+	}
```
JeyJeyGao, line 167: Those lines may can be removed by using `CLIPlugin` and call [GetMetadata](https://github.com/notaryproject/notation/blob/26187094a03643e0b283d40f34e39cf599df75dd/cmd/notation/plugin.go#L128)

cmd/notation/plugin.go
JeyJeyGao: Yes, we need to reuse code.

Duffney:
My apologies to everyone for the delay in replying to these comments. I'm actively re-reviewing all the comments and working to resolve them. Thank you for your patience. :)

cmd/notation/plugin.go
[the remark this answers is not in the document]
Duffney: @priteshbandi, I got this working, but it also causes you to assume the plugin name. For the NewCLIManager to get the metadata it needs to know the directory name of the plugin, which happens to also be the plugin name. However, the argument for the install command is given the binary name that by naming convention includes `notation` with an optional file ext of `.exe` for the Windows binary. The easy work around is to split the prefix off and the file extension to get the name of the plugin but if the plugin for any reason doesn't follow naming standards the logic will fail. Do you see any way around that?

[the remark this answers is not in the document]
Duffney: Hey @priteshbandi, I tinkered with this implementation a bit and came up with the following:

You can use the GetMetadata method from the CLIPlugin type to get the metadata from the binary, but it needs to be in a directory structure that mimics the notation layout. 

For example, a binary called `notation-azure-kv.exe` would have to be put into a directory called `$binaryPath\azure-kv` to extract the metadata from the binary. Otherwise, you'll an error:

> ` CreateFile C:\Users\userName\downloads\azure-kv\notation-azure-kv.exe: The system cannot find the path specified.

(If there's a way to get around this please LMK)

With that said that logic can be used to get the metadata from an existing plugin within Notation's plugin directory. What do you think about using exec.Command to get the metadata from the plugin being installed and use the GetMetaData method to get information from an existing plugin, if it exists?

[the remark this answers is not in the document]
Duffney: I've got a working change to replace exec.Command with the GetMetadata method. Once I've replaced the semver logic as well I'll push the changes to the PR branch. :)

cmd/notation/plugin.go
```diff
@@ -62,3 +108,169 @@ func listPlugins(command *cobra.Command) error {
...
+				return nil
+			}
+
+			fmt.Printf("Copying plugin %s to directory %s...\n", pluginName, pluginDir)
+			copyPlugin(pluginPath, pluginDir+"/"+pluginBinary)
+		}
+
+		// do not copy plugin, if new plugin version is less than or equal to current plugin version
+		if newSemVersion.LessThan(currentVersion) || newSemVersion.Equal(currentVersion) {
+			fmt.Println("Current version is greater than or equal to new version. Skipping plugin installation.\nUse --force flag to overwrite the plugin.")
```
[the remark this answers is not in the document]
Duffney: [Changes ](https://github.com/notaryproject/notation/pull/617/commits/7f9d96fe25c330c79cb9c61266ec9ae492111098) made to use the same semver as notation-go.

cmd/notation/plugin.go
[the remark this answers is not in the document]
Duffney: Replaced the exe.Command logic with [GetMetadata ](https://github.com/notaryproject/notation/pull/617/commits/e4d7ef5f0b4d53f2e1a218350742fab9f09c21ff) from the plugin.

cmd/notation/plugin.go
Duffney: Got the changes made to implement this in a recent [commit](https://github.com/notaryproject/notation/pull/617/commits/e4d7ef5f0b4d53f2e1a218350742fab9f09c21ff). :)

cmd/notation/plugin.go
Duffney: [Changes ](https://github.com/notaryproject/notation/pull/617/commits/e4d7ef5f0b4d53f2e1a218350742fab9f09c21ff) made to use the `CLIPlugin`.

cmd/notation/plugin.go
Duffney: I certainly can, I've made the necessary changes to use `golang.org/x/mod/semver'. :)

cmd/notation/plugin.go
Duffney: Hey @patrickzheng200, thanks for letting me know about that internal helper. I [replaced](https://github.com/notaryproject/notation/pull/617/commits/3eb146c202f2e6a75d4184f5c79d43cd42bc80e5)the logic in the install command to use the CopyToDir func.

cmd/notation/plugin.go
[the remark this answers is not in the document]
Duffney: I added a helper called exists to reduce the duplicate code. It might be a helpful function to add into the internal ioultil package at some point. Same with the `splitPluginName` that I added.

cmd/notation/plugin.go
Duffney: Great feedback @patrickzheng200, I refactored the code to use a helper to check the existence of the plugin with the following pseudo logic:

```go
dirExists, err := exists(pluginDestPath)
if err != nil {
    return err
}

if dirExists {
fmt.Println("Exists")
// if force, copyToDir, then return 
// check versions
// osutil.CopyToDir
}

if !dirExists {
fmt.Println("Does not exist")
// copyToDir
}
```

cmd/notation/plugin.go
Duffney: Helper function called `exists`, as well as, error handling for `err != nil`.

cmd/notation/plugin.go
Duffney: `if force` logic has been [replaced](https://github.com/notaryproject/notation/pull/617/commits/3eb146c202f2e6a75d4184f5c79d43cd42bc80e5#diff-2ce9482b6436893530d00c1cf4d9cd0231854ce00b30bc668ee63b16fca4b373R152).

cmd/notation/plugin.go
Duffney: Great call out @patrickzheng200, I've replaced the logic to use a helper function and then handle the errors [here](https://github.com/notaryproject/notation/pull/617/commits/3eb146c202f2e6a75d4184f5c79d43cd42bc80e5#diff-2ce9482b6436893530d00c1cf4d9cd0231854ce00b30bc668ee63b16fca4b373L238).

cmd/notation/plugin.go
Duffney: I've replaced this logic with a call to [CopyToDir ](https://github.com/notaryproject/notation/blob/main/internal/osutil/file.go#L43) from within the internal package.

cmd/notation/plugin.go
Duffney: Are there any other config settings we'd like to add to the struct? If so, it might be worthwhile to look into adding [viper](https://github.com/spf13/viper) for configuration.

Duffney:
> @priteshbandi I'd like to see if `notation plugin install` from a tar gz file. There is no security risk if we require the file structure in the tar ball to be flatten and only allow regular files. No symlinks and hard links of course.

I agree. Users will expect the subcommand to handle archives because that's how they're downloaded. If we don't add it in this PR, it might be worth adding error handling for people who pass in a .zip or .tar.gz to throw and error saying it needs to be extracted.

yizha1:
Closed as plugin installation was covered by a different PR https://github.com/notaryproject/notation/pull/827

PR #619, opened 2023-04-06
fix: added warning for dangling referrers index deletion

### LINKED ISSUE #615: Improve the error msg for OCI v1.0 registries or registries that don't support deletion API

Notation CLI

When using Notation CLI to sign an artifact in OCI v1.0 compliant registries such as GHCR or registries that [don't support deletion API](https://github.com/docker/hub-feedback/issues/1759) (e.g. Docker Hub), Notation will return an unexpected error message and may mislead or confuse users. It happens when signing an artifact for the second time.

Improve the error message to be more descriptive and user-friendly, and tell users about the right result exactly. Giving an example: Successfully signed `<registry>/<repository>@<digest>`. Please note that Notation can't delete the referrers index since this is an OCI v1.0 compliant registry or the deletion API is disabled by the registry.

### How can we reproduce it?

We have a user who signed a sample image in GHCR. It returned the following error message when this user signed the same image for the second time:

```
$ notation sign --signature-format cose --key $REMOTE_KEY_NAME ghcr.io/demo-user/python@sha256:xxxx
Error: failed to push signature to registry with error: failed to push manifest: failed to delete dangling referrers index sha256:xxx for referrers tag sha256-1ba1561: DELETE "https://ghcr.io/v2/demo-user/python/manifests/sha256:e1b7d6": response status code 405: unsupported: The operation is unsupported.
```
It may confuse users why this signing behavior failed due to an old referrers index manifest deletion failure. From an end user's point of view, the user is left with the impression that the signing failed. Now the user is left wondering why this happened and if she/he doesn't know the internals of OCI registries she/he may think the issue is with Notation.  

In fact, the signing is successful and the signature has been attached to the image but it doesn't tell users the right result unless users view the referenced signatures using `notation list` as below.

```
ghcr.io/demo-user/python@sha256:xxxx
    ├── sha256:85f65b60a6e260d52ba5f8d556d444320b92c2ac140ec8d4944dbd8b03deb18c
    └── sha256:96130e5a5e314ef90e56e24559e48065d2889c39661b9b3a944e6862e1e51c0a
```

### Describe your environment

Ubuntu 20.08

Notation CLI v1.0.0-RC.3

patrickzheng200:
This PR tries to resolve #615.

### DIFF
--- a/cmd/notation/sign.go
+++ b/cmd/notation/sign.go
@@ -19,10 +19,11 @@ import (
 )
 
 const (
-	signatureManifestArtifact     = "artifact"
-	signatureManifestImage        = "image"
-	referrersTagSchemaDeleteError = "failed to delete dangling referrers index"
+	signatureManifestArtifact = "artifact"
+	signatureManifestImage    = "image"
 )
+
+const referrersTagSchemaDeleteError = "failed to delete dangling referrers index"
 
 var supportedSignatureManifest = []string{signatureManifestArtifact, signatureManifestImage}
 
--- a/cmd/notation/sign.go
+++ b/cmd/notation/sign.go
@@ -119,9 +119,10 @@ func runSign(command *cobra.Command, cmdOpts *signOpts) error {
 			if !ociImageManifest {
 				return fmt.Errorf("%v. Possible reason: target registry does not support OCI artifact manifest. Try removing the flag `--signature-manifest artifact` to store signatures using OCI image manifest", err)
 			}
-			if strings.Contains(err.Error(), referrersTagSchemaDeleteError) {
-				// failed to delete dangling referrers index
-				fmt.Fprintln(os.Stderr, "Warning: failed to delete dangling referrers index, since this is an OCI v1.0 compliant registry and the deletion API is disabled by the registry.")
+			if _, after, found := strings.Cut(err.Error(), referrersTagSchemaDeleteError); found {
+				warnMsg := fmt.Sprintf("Warning: %s %s", referrersTagSchemaDeleteError, after)
+				fmt.Fprintln(os.Stderr, warnMsg)
+				// write out
 				fmt.Println("Successfully signed", ref)
 				return nil
 			}
--- a/cmd/notation/sign.go
+++ b/cmd/notation/sign.go
@@ -119,9 +119,8 @@ func runSign(command *cobra.Command, cmdOpts *signOpts) error {
 			if !ociImageManifest {
 				return fmt.Errorf("%v. Possible reason: target registry does not support OCI artifact manifest. Try removing the flag `--signature-manifest artifact` to store signatures using OCI image manifest", err)
 			}
-			if _, after, found := strings.Cut(err.Error(), referrersTagSchemaDeleteError); found {
-				warnMsg := fmt.Sprintf("Warning: %s %s", referrersTagSchemaDeleteError, after)
-				fmt.Fprintln(os.Stderr, warnMsg)
+			if strings.Contains(err.Error(), referrersTagSchemaDeleteError) {
+				fmt.Fprintln(os.Stderr, "Warning: Removal of outdated referrers index is not supported by the remote registry. Garbage collection may be required.")
 				// write out
 				fmt.Println("Successfully signed", ref)
 				return nil
--- a/cmd/notation/sign.go
+++ b/cmd/notation/sign.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/notaryproject/notation-go"
@@ -18,8 +19,9 @@ import (
 )
 
 const (
-	signatureManifestArtifact = "artifact"
-	signatureManifestImage    = "image"
+	signatureManifestArtifact     = "artifact"
+	signatureManifestImage        = "image"
+	referrersTagSchemaDeleteError = "failed to delete dangling referrers index"
 )
 
 var supportedSignatureManifest = []string{signatureManifestArtifact, signatureManifestImage}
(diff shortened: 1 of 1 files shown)


sajayantony:
Should the exit code be zero? /cc @toddysm

toddysm:
@sajayantony I believe so and avoid braking any scripts. 

Also, is this the only message we will show:
```
Warning: failed to delete dangling referrers index sha256:xxx for referrers tag sha256-1ba1561: DELETE "https://ghcr.io/v2/demo-user/python/manifests/sha256:e1b7d6": response status code 405: unsupported: The operation is unsupported.
```

I believe there should still be a confirmation that the operation succeeded and the signature got attached to the artifact. We should show the standard message we show when signature is attached + the warning.

cmd/notation/sign.go
```diff
@@ -22,6 +23,8 @@ const (
 	signatureManifestImage    = "image"
 )
 
+const referrersTagSchemaDeleteError = "failed to delete dangling referrers index"
```
qweeah, line 26: @shizhMSFT @Wwwsylvia Since this error is generated based on oras-go-specific procedure, should oras-go export this error so caller can implement upon?

cmd/notation/sign.go
```diff
@@ -112,8 +115,16 @@ func runSign(command *cobra.Command, cmdOpts *signOpts) error {
...
 		var errorPushSignatureFailed notation.ErrorPushSignatureFailed
-		if errors.As(err, &errorPushSignatureFailed) && !ociImageManifest {
-			return fmt.Errorf("%v. Possible reason: target registry does not support OCI artifact manifest. Try removing the flag `--signature-manifest artifact` to store signatures using OCI image manifest", err)
+		if errors.As(err, &errorPushSignatureFailed) {
+			if !ociImageManifest {
+				return fmt.Errorf("%v. Possible reason: target registry does not support OCI artifact manifest. Try removing the flag `--signature-manifest artifact` to store signatures using OCI image manifest", err)
+			}
+			if strings.Contains(err.Error(), referrersTagSchemaDeleteError) {
+				// failed to delete dangling referrers index
+				fmt.Fprintln(os.Stderr, "Warning: failed to delete dangling referrers index, since this is an OCI v1.0 compliant registry and the deletion API is disabled by the registry.")
```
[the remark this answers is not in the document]
qweeah: > since this is an OCI v1.0 compliant registry and the deletion API is disabled by the registry

This might not be accurate, except the case that 405(deletion API is disabled), the error might also come from  OCI v1.1 registry with 403(client token don't have access to delete).

patrickzheng200:
> Should the exit code be zero? /cc @toddysm

Yes, the exit code should be 0. Once the warning is printed out, `Successfully signed...` is printed out and finally `return nil`.

patrickzheng200:
> @sajayantony I believe so and avoid braking any scripts.
> 
> Also, is this the only message we will show:
> 
> ```
> Warning: failed to delete dangling referrers index sha256:xxx for referrers tag sha256-1ba1561: DELETE "https://ghcr.io/v2/demo-user/python/manifests/sha256:e1b7d6": response status code 405: unsupported: The operation is unsupported.
> ```
> 
> I believe there should still be a confirmation that the operation succeeded and the signature got attached to the artifact. We should show the standard message we show when signature is attached + the warning.

This warning message is an addition to the original result. Can refer this in the code.

cmd/notation/sign.go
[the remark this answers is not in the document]
patrickzheng200: Code updated.

cmd/notation/sign.go
qweeah: It's beyond this PR's scope, I created an [issue ](https://github.com/oras-project/oras-go/issues/479) for tracking, let's move discussion there.

cmd/notation/sign.go
```diff
@@ -112,8 +115,17 @@ func runSign(command *cobra.Command, cmdOpts *signOpts) error {
...
 	if err != nil {
 		var errorPushSignatureFailed notation.ErrorPushSignatureFailed
-		if errors.As(err, &errorPushSignatureFailed) && !ociImageManifest {
-			return fmt.Errorf("%v. Possible reason: target registry does not support OCI artifact manifest. Try removing the flag `--signature-manifest artifact` to store signatures using OCI image manifest", err)
+		if errors.As(err, &errorPushSignatureFailed) {
+			if !ociImageManifest {
+				return fmt.Errorf("%v. Possible reason: target registry does not support OCI artifact manifest. Try removing the flag `--signature-manifest artifact` to store signatures using OCI image manifest", err)
+			}
+			if _, after, found := strings.Cut(err.Error(), referrersTagSchemaDeleteError); found {
+				warnMsg := fmt.Sprintf("Warning: %s %s", referrersTagSchemaDeleteError, after)
```
qweeah, line 123: Why not using `err.Error()` directly here?
```suggestion
				warnMsg := fmt.Sprintf("Warning: signature successfully pushed with minor cleanup failure: %s", err.Error())
```

P.S. I also met issue on how to describe this error to customer without too much low-level detail for oras' fix.

patrickzheng200: Because the full content of `err.Error()` starts with:  `Error: failed to push signature to registry with error: failed to push manifest:...`. However, in this case, it's not an error, both push signature and push manifest were succeeded. 
Note: `failed to push manifest` is an error message from oras-go: https://github.com/oras-project/oras-go/blob/e8225cb1e125bd4c13d6b586ae6d862050c3fae2/pack.go#L174.

cmd/notation/sign.go
Wwwsylvia: Yeah we need to think about how to improve `oras-go` for such use case.

cmd/notation/sign.go
Wwwsylvia: How about something like "Warning: signature is successfully pushed but the old referrer index failed to be cleaned up"?

patrickzheng200:
Closing this PR due to suggestion here: https://github.com/notaryproject/notation/issues/615#issuecomment-1500053692.
IMO, we should think through this process again before implementation.

patrickzheng200:
Reopened based on discussion.

byronchien (approved):

shizhMSFT (approved):
LGTM

cmd/notation/sign.go
shizhMSFT: Once `oras-go` is updated and have a new release, we can update here. But for now, let's do a string match.

PR #649, opened 2023-04-25
fix: Throw error when extra arguments are passed to command

priteshbandi:
fix: Throw error when extra arguments are passed to command

### DIFF
--- a/internal/osutil/file_test.go
+++ b/internal/osutil/file_test.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime"
 	"testing"
 )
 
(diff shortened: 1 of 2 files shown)


internal/arg/arg.go
```diff
@@ -0,0 +1,16 @@
+package arg
+
+import (
+	"errors"
+	"fmt"
+)
+
+func ValidateCount(args []string, expLen int, missingErrMsg string) error {
```
patrickzheng200, line 8: nit:
Although it's internal, I think we should add a func doc for it?

cmd/notation/inspect.go
```diff
@@ -71,8 +72,8 @@ Example - Inspect signatures on an OCI artifact identified by a digest and outpu
   notation inspect --output json <registry>/<repository>@<digest>
 `,
 		Args: func(cmd *cobra.Command, args []string) error {
-			if len(args) == 0 {
-				return errors.New("missing reference")
+			if err := arg.ValidateCount(args, 1, "missing reference"); err != nil {
```
JeyJeyGao, line 75: Can we use Cobra buit-in [Positional and Custom Arguments](https://github.com/spf13/cobra/blob/main/user_guide.md#positional-and-custom-arguments)?

Issue #600, opened 2023-03-25
Simplify Docker Credential Helper configuration for Notation authentication


FeynmanZhou:
### What is the areas you would like to add the new feature to?

Notation CLI

### Is your feature request related to a problem?

To achieve maximum compatibility with existing systems, [docker credential helpers](https://github.com/docker/docker-credential-helpers) and its [protocol](https://docs.docker.com/engine/reference/commandline/login/#credential-helper-protocol) are adopted as the credential helpers for Notation registry authentication.

It requires cumbersome steps to configure [docker credential helpers](https://github.com/docker/docker-credential-helpers) when users log in to a registry. Users will encounter an authentication problem in Linux as follows if the Docker Credential Helper configuration is not configured or misconfigured:

```
Error: could not get the credentials store: failed to load config file, error: credentials store config was not set up
```

We need to simplify the authentication experience and remove dependency on Docker Credential Helper configuration.

### What solution do you propose?

Integrate [oras-credentials-go](https://github.com/oras-project/oras-credentials-go) library to simplify this authentication experience and remove the dependency. This library will be able to read credentials from the Docker configuration file. 

My expectation is that if there is Docker Hub registry credential stored in the Docker configuration file, users will be able to log in to Docker Hub without additional configuration and input username/password. 

### What alternatives have you considered?

No

yizha1:
update milestone to `v1.0.0-rc.6` per discussion in community call on 5/23/2023

PR #654, opened 2023-04-27
refactor: use `oras-credentials-go` for credential management

Wwwsylvia:
Resolves: #600 
Resolves: #637


Wwwsylvia:
## Behaviors of this PR

### notation login
- If there is a credentials store configured in the notation config file
  - Save the credential into the configured credentials store
- If no credentials store is configured, or the notation config file does not exist
  - Use the platform-default credentials store, and set it back to the notation config file
  - If the platform-default credentials store is not installed, throw an error

### notation logout
- If there is a credentials store configured in the notation config file
  - Remove the credential of the registry
- If no credentials store is configured, or the notation config file does not exist
  - No operation

### Other notation operations that require authentication
Look for the credential in following order:
1. notation config file
  - Configured credentials store
  - Platform-default credentials store
  - Plaintext credentials
2. Docker config file
  - Configured credentials store
  - Platform-default credentials store
  - Plaintext credentials

Wwwsylvia:
Login error example:
image

cmd/notation/registry.go
```diff
@@ -159,9 +161,7 @@ func setHttpDebugLog(ctx context.Context, authClient *auth.Client) {
 	authClient.Client.Transport = trace.NewTransport(authClient.Client.Transport)
 }
 
-func getAuthClient(ctx context.Context, opts *SecureFlagOpts, ref registry.Reference) (*auth.Client, bool, error) {
-	var plainHTTP bool
-
+func getAuthClient(ctx context.Context, opts *SecureFlagOpts, ref registry.Reference) (authClient *auth.Client, plainHTTP bool, err error) {
```
shizhMSFT, line 164: Avoid using named return parameters unless it is necessary.

Wwwsylvia: A comment will also work though.

pkg/auth/native_store.go
```diff
@@ -1,97 +0,0 @@
...
-	return nil, fmt.Errorf("could not get the configured credentials store for registry: %s", registryHostname)
-}
-
-// newNativeAuthStore creates a new native store that uses a remote helper
-// program to manage credentials. Note: it's different from the nativeStore in
-// docker-cli which may fall back to plain text store
-func newNativeAuthStore(ctx context.Context, helperSuffix string) CredentialStore {
-	logger := log.GetLogger(ctx)
-	name := remoteCredentialsPrefix + helperSuffix
-	logger.Infoln("Executing remote credential helper program:", name)
```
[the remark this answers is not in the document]
Wwwsylvia: Will do.

cmd/notation/logout.go
```diff
@@ -40,16 +41,12 @@ func logoutCommand(opts *logoutOpts) *cobra.Command {
...
-	nativeStore, err := auth.GetCredentialsStore(ctx, serverAddress)
+	credsStore, err := auth.NewCredentialsStore()
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get credentials store: %v", err)
 	}
-	err = nativeStore.Erase(serverAddress)
-	if err != nil {
-		return err
+	if err := credentials.Logout(ctx, credsStore, opts.server); err != nil {
```
JeyJeyGao, line 48: What will happen if no native credential store?

Wwwsylvia: `Logout` is to delete the credential for the given server address. If there is no native credentials store, `oras-credentials-go` will try to delete the credentials from the config file. If the credential is not present in the config file, or the config file does not exist, the deletion will result in no operation.
See https://github.com/notaryproject/notation/pull/654#issuecomment-1525562629 for details.

cmd/notation/registry.go
Wwwsylvia: Removed

pkg/auth/native_store.go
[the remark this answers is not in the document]
Wwwsylvia: Issues open:
- https://github.com/notaryproject/notation/issues/662
- https://github.com/oras-project/oras-credentials-go/issues/63

qweeah:
There are suppressed E2E tests for `notation login`, any thought to enable cases without using cred store?
https://github.com/notaryproject/notation/blob/5516199cd2d83349468682379c1e1f4ccd24414a/test/e2e/suite/command/login.go#L15

go.sum
```diff
@@ -25,6 +25,10 @@ github.com/opencontainers/go-digest v1.0.0 h1:apOUWs51W5PlhuyGyz9FCeeBIOUDA/6nW8
 github.com/opencontainers/go-digest v1.0.0/go.mod h1:0JzlMkj0TRzQZfJkVvzbP0HBR3IKzErnv2BNG4W4MAM=
 github.com/opencontainers/image-spec v1.1.0-rc2 h1:2zx/Stx4Wc5pIPDvIxHXvXtQFW/7XWJGmnM7r3wg034=
 github.com/opencontainers/image-spec v1.1.0-rc2/go.mod h1:3OVijpioIKYWTqjiG0zfF6wvoJ4fAXGbjdZuI2NgsRQ=
+github.com/oras-project/oras-credentials-go v0.0.0-20230424070720-ba6b33c40845 h1:lE/TKx3cblnIRMazUdXgHmoM8TXs4fZqv3EXvEF3A2I=
+github.com/oras-project/oras-credentials-go v0.0.0-20230424070720-ba6b33c40845/go.mod h1:yww9XqMCWjNh4Z7S5Ek7KeV9j/yhbyVxgSinGdNNwEg=
+github.com/oras-project/oras-credentials-go v0.1.0 h1:XaGxQD4qK8jxOGSNqrPlh/lKlVSL/PBH89WraYjyZkE=
+github.com/oras-project/oras-credentials-go v0.1.0/go.mod h1:wxHUtRSULM8DVTGgc8DFtIKXDZr2JLD/48WYPqvKoJs=
```
shizhMSFT, line 31: Have you tried `go mod tidy` to remove the `v0.0.0` entry?

go.sum
```diff
@@ -67,3 +71,5 @@ gopkg.in/yaml.v3 v3.0.1 h1:fxVm/GzAzEWqLHuvctI91KS9hhNmmWOoWu0XTYJS7CA=
 gopkg.in/yaml.v3 v3.0.1/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
 oras.land/oras-go/v2 v2.0.2 h1:3aSQdJ7EUC0ft2e9PjJB9Jzastz5ojPA4LzZ3Q4YbUc=
 oras.land/oras-go/v2 v2.0.2/go.mod h1:PWnWc/Kyyg7wUTUsDHshrsJkzuxXzreeMd6NrfdnFSo=
```
shizhMSFT, line 73: and also this one

cmd/notation/login.go
```diff
@@ -76,55 +78,45 @@ func runLogin(ctx context.Context, opts *loginOpts) error {
...
+		return fmt.Errorf("failed to get credentials store: %v", err)
 	}
-
-	nativeStore, err := auth.GetCredentialsStore(ctx, serverAddress)
+	registry, err := getRegistryClient(ctx, &opts.SecureFlagOpts, serverAddress)
 	if err != nil {
-		return fmt.Errorf("could not get the credentials store: %v", err)
+		return fmt.Errorf("failed to get registry client: %v", err)
 	}
+	if err := credentials.Login(ctx, credsStore, registry, cred); err != nil {
```
shizhMSFT, line 97: It is strange to login with a `registry` with `cred` pre-baked.

cmd/notation/login.go
```diff
@@ -76,55 +78,45 @@ func runLogin(ctx context.Context, opts *loginOpts) error {
...
-		opts.Password,
-	)
-	if err = nativeStore.Store(serverAddress, creds); err != nil {
-		return fmt.Errorf("failed to store credentials: %v", err)
+		// ErrPlaintextPutDisabled returned by Login() indicates that the credential is validated
+		// but is not saved because there is no native credentials store available
+		if savedCred, err := credsStore.Get(ctx, registryName); err == nil && savedCred == cred {
+			// there is an existing identical credential, ignore saving error
+			fmt.Fprintf(os.Stderr, "Warning: The credentials store is not set up. It is recommended to configure the credentials store to securely store your credentials. See %s.\n", urlDocHowToAuthenticate)
+			fmt.Println()
```
shizhMSFT, line 108: The new line seems not necessary?

cmd/notation/login.go
```diff
@@ -76,55 +78,45 @@ func runLogin(ctx context.Context, opts *loginOpts) error {
...
+			fmt.Println("Authenticating with existing credentials...")
```
shizhMSFT, line 109: It is not true, right? We did not implement that with `notation login <registry>` without explicitly input credentials.

go.sum
Wwwsylvia: fixed

go.sum
Wwwsylvia: fixed

cmd/notation/login.go
Wwwsylvia: It's just to remind users that there are some existing credentials that can be used.

cmd/notation/login.go
Wwwsylvia: It's to separate the warning and the message

shizhMSFT: But the warning goes to `stderr` and others go to `stdout`. They are two different channels.

Wwwsylvia: Yeah but from human's point of view they are displayed together.🤔

ningziwen (approved):

shizhMSFT:
`oras-credentials-go` is basically a similar implementation of https://github.com/docker/cli/tree/master/cli/config/configfile

`oras-credentials-go` does not have any crypto component in it and it just does CRUD and thus security review is not required.

Wwwsylvia:
Hi @priteshbandi , the design discussion for the `oras-credentials-go` library can be found [here](https://github.com/oras-project/oras-credentials-go/discussions/18).
As @shizhMSFT has mentioned, the library is just to cover CRUD of credentials and does not handle any encryption/decryption.

cmd/notation/login.go
yizha1: @shizhMSFT As Sylvia said, it is just to inform users that Notation is authenticating with existing credentials, I am thinking maybe we can be more specific, `Authenticating with existing credentials from docker config...`

cmd/notation/login.go
Wwwsylvia: Removed the new line.

cmd/notation/login.go
Wwwsylvia: Changed it to

> Authenticat**ed** with existing credentials

Wwwsylvia:
I'm working on the conflicts.

cmd/notation/login.go
```diff
@@ -70,61 +72,65 @@ func runLogin(ctx context.Context, opts *loginOpts) error {
...
-	if err = nativeStore.Store(serverAddress, creds); err != nil {
-		return fmt.Errorf("failed to store credentials: %v", err)
+		// ErrPlaintextPutDisabled returned by Login() indicates that the credential is validated
+		// but is not saved because there is no native credentials store available
+		if savedCred, err := credsStore.Get(ctx, registryName); err == nil && savedCred == cred {
+			// there is an existing identical credential, ignore saving error
+			fmt.Fprintf(os.Stderr, "Warning: The credentials store is not set up. It is recommended to configure the credentials store to securely store your credentials. See %s.\n", urlDocHowToAuthenticate)
+			fmt.Println("Authenticated with existing credentials")
+		} else {
+			return fmt.Errorf("failed to log in to %s: the credential could not be saved because a credentials store is required to securely store the password. See %s",
```
shizhMSFT, line 125: The error message for the `else` block is not accurate. It is possible that `Get` failed with `err != nil`.

cmd/notation/login.go
```diff
@@ -70,61 +72,65 @@ func runLogin(ctx context.Context, opts *loginOpts) error {
...
+				registryName, urlDocHowToAuthenticate)
+		}
```
shizhMSFT, line 127:
```suggestion
		if savedCred, err := credsStore.Get(ctx, registryName); err != nil {
			return err // a customized error message should be returned
		}
		if savedCred != cred {
			return fmt.Errorf("failed to log in to %s: the credential could not be saved because a credentials store is required to securely store the password. See %s",
				registryName, urlDocHowToAuthenticate)
		}
		// there is an existing identical credential, ignore saving error
		fmt.Fprintf(os.Stderr, "Warning: The credentials store is not set up. It is recommended to configure the credentials store to securely store your credentials. See %s.\n", urlDocHowToAuthenticate)
		fmt.Println("Authenticated with existing credentials")
```

Wwwsylvia: IMHO, if the credentials store is not set up, we should only print "login succeeded" when the provided credential exists.
If `err != nil` and we cannot check the existence, we should still throw the "the credential could not be saved because a credentials store is required to securely store the password." error.

cmd/notation/login.go
Wwwsylvia: If the credentials store is not set up, we print "login succeeded" only when the provided credential exists.

If `err != nil` and we cannot check the existence, we should still throw the "the credential could not be saved because a credentials store is required to securely store the password." error. 
IMHO, customizing the error message for `Get` failure might no be necessary as it does not make sense to users.

cmd/notation/login.go
```diff
@@ -70,61 +72,65 @@ func runLogin(ctx context.Context, opts *loginOpts) error {
...
 
-	// init creds
-	creds := newCredentialFromInput(
-		opts.Username,
-		opts.Password,
-	)
-	if err = nativeStore.Store(serverAddress, creds); err != nil {
-		return fmt.Errorf("failed to store credentials: %v", err)
+		// ErrPlaintextPutDisabled returned by Login() indicates that the credential is validated
```
Two-Hearts, line 118: nit:
This comment exceeds the 80-character limit.

Two-Hearts:
> ## Behaviors of this PR
> ### notation login
> * If the provided credential is invalid or the connection to registry fails, throw an error
>   Screenshot 2023-05-11 113659
> * If there is a credentials store configured in the notation config file
>   
>   * Save the credential into the configured credentials store
> * If no credentials store is configured, or the notation config file does not exist
>   
>   * Use the platform-default credentials store, and set it back to the notation config file
>   * If the platform-default credentials store is not installed
>     
>     * If the provided credential is identical with the existing credential, print:
>       image
>     * If the provided credential is different with the saved one, throw an error
>       image
> 
> ### notation logout
> * If there is a credentials store configured in the notation config file
>   
>   * Remove the credential of the registry
> * If no credentials store is configured, or the notation config file does not exist
>   
>   * No operation
> 
> ### Other notation operations that require authentication
> Look for the credentials in following order:
> 
> 1. notation config file
> 
> * Configured credential helpers
> * Configured credentials store
> * Platform-default credentials store
> * Plaintext credentials
> 
> 2. Docker config file
> 
> * Configured credential helpers
> * Configured credentials store
> * Platform-default credentials store
> * Plaintext credentials

@Wwwsylvia I'm still trying to understand this process. So, under which circumstances does Notation login/logout perform `write` operation (if there is any) on the Docker config file?

Wwwsylvia:
> I'm still trying to understand this process. So, under what circumstances does Notation login/logout perform write operations (if any) on the Docker config file?

notation will never write to the Docker config file.
If no credentials store is configured, or the notation config file does not exist, notation will use the platform-default credentials store, and set the "credsStore" field back to the notation config file.

internal/auth/credentials.go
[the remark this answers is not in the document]
Wwwsylvia: I think `cmd/notation/internal` is for functions that are specific to cmd opersions, while `NewCrdentialStore` is generic and is not related to cmd. Hence I prefer to keep it here.

cmd/notation/login.go
shizhMSFT: At least, we should document why the case `err != nil` is conveyed and merged to the case `err == nil && savedCred != cred`.

cmd/notation/login.go
```diff
@@ -70,61 +72,65 @@ func runLogin(ctx context.Context, opts *loginOpts) error {
...
+			}
+		} else {
+			// username is not empty, prompt for password
+			opts.Password, err = readPasswordFromPrompt(reader, false)
+			if err != nil {
+				return err
+			}
+			if opts.Password == "" {
+				return errors.New("password required")
+			}
```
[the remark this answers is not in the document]
Wwwsylvia: Thanks! The suggested code is more concise.

cmd/notation/login.go
Wwwsylvia: Added a warning log for the case of `err != nil`.

cmd/notation/registry.go
```diff
@@ -118,9 +120,10 @@ func setHttpDebugLog(ctx context.Context, authClient *auth.Client) {
 	authClient.Client.Transport = trace.NewTransport(authClient.Client.Transport)
 }
 
-func getAuthClient(ctx context.Context, opts *SecureFlagOpts, ref registry.Reference) (*auth.Client, bool, error) {
+// getAuthClient returns an *auth.Client and a bool indicating if plain HTTP
+// should be used.
```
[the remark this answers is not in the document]
Wwwsylvia: Added.

Two-Hearts (approved):
LGTM

JeyJeyGao (approved):
LGTM

cmd/notation/login.go
```diff
@@ -70,61 +73,70 @@ func runLogin(ctx context.Context, opts *loginOpts) error {
...
+		opts.Password, err = readPasswordFromPrompt(reader, isToken)
 		if err != nil {
 			return err
 		}
+		if opts.Password == "" {
+			if isToken {
+				return errors.New("token required")
+			} else {
+				return errors.New("password required")
+			}
```
shizhMSFT, line 97: nit:
```suggestion
			}
			return errors.New("password required")
```

shizhMSFT (approved):
LGTM with nit.

cmd/notation/login.go
Wwwsylvia: updated

JeyJeyGao (dismissed):
LGTM

Two-Hearts (dismissed):
LGTM

shizhMSFT (dismissed):
LGTM

cmd/notation/login.go
```diff
@@ -70,61 +73,69 @@ func runLogin(ctx context.Context, opts *loginOpts) error {
...
+		// credentials store available
+		if savedCred, err := credsStore.Get(ctx, registryName); err != nil || savedCred != cred {
+			if err != nil {
+				// if we fail to get the saved credential, log a warning
+				// but do not throw the GET error, as the error could be
+				// confusing to users
+				logger := log.GetLogger(ctx)
+				logger.Warnf("Failed to get the existing credentials for %s: %v", registryName, err)
+			}
+			return fmt.Errorf("failed to log in to %s: the credential could not be saved because a credentials store is required to securely store the password. See %s",
```
Two-Hearts, line 126: Question: when `err == nil && savedCred != cred`, the current error message is `...the credential could not be saved because a credentials store is required to securely store the password...` But I think the correct error message should be something like `...user provided credential does not match with the existing one...`?

cmd/notation/login.go
```diff
@@ -70,61 +73,69 @@ func runLogin(ctx context.Context, opts *loginOpts) error {
...
+				registryName, urlDocHowToAuthenticate)
+		}
+
+		// the credential already exists somewhere, ignore the saving error
+		fmt.Fprintf(os.Stderr, "Warning: The credentials store is not set up. It is recommended to configure the credentials store to securely store your credentials. See %s.\n", urlDocHowToAuthenticate)
+		fmt.Println("Authenticated with existing credentials")
```
Two-Hearts, line 132: Question: When the printouts are line 131 + line 132, the user might be curious about where exactly the existing credentials are saved? Do we want to provide more details here?

cmd/notation/login.go
Wwwsylvia: Comparing the existing credential with the provided one is a trick - If the credentials store is not set up, but there is an identical existing credential, the user can still do signing.
If the provided credential is different, we cannot save it because the credentials store is not available, and the login fails. 
A message like "user provided credential does not match with the existing one..." does not make sense to the user.

cmd/notation/login.go
Wwwsylvia: From the notation standpoint, we don't know where exactly the existing credential is... Will this information be useful in any scenarios?

priteshbandi (approved):
LGTM

cmd/notation/login.go
Two-Hearts: > From the notation standpoint, we don't know where exactly the existing credential is... Will this information be useful in any scenarios?

Yeah, because the printouts claim that the `login` succeeded without a credential store. As a user, it's nature to ask, then where are my credentials saved?

JeyJeyGao (approved):
LGTM

cmd/notation/login.go
Two-Hearts: Discussed offline with @Wwwsylvia, we may update this error message in the future.

cmd/notation/login.go
Wwwsylvia: I think the user should know better than notation as they have logged in using other tools by themselves. 😂

PR #663, opened 2023-05-12
update: align Notation with OCI specs

### LINKED ISSUE #659: Mark feature "support OCI Referrer API" as experimental feature

Notation CLI

Currently [Referrer API](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers) is by default verified by Notation since `v1.0.0-rc.4`. For example, when using `notation sign`, if registries support referrer API, notation will not use an image index pushed to a tag described by the [referrers tag schema](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#referrers-tag-schema). Since OCI spec 1.1 (including both image spec and distribution spec) is not officially released yet, support of Referrer API could lead to breaking changes or portability issue after Notation v1 is released. On the other hand, users may be confused that sometimes image index is pushed, sometimes not, which is an issue of consistency.

Mark feature "support OCI Referrer API" as experimental feature, and document it properly. Users can use experimental features only after env variable `NOTATION_EXPERIMENTAL=1` is set. 
The default behavior of Notation is always using an image index pushed to a tag described by the [referrers tag schema](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc2/spec.md#referrers-tag-schema). If users set env variable `NOTATION_EXPERIMENTAL=1`, Notation will check the support of OCI referrer API first, if registries support Referrer API, then notation will not push image index; if registries don't support Referrer API, notation will push image index as the default behavior. 

### How can we reproduce it?

N/A

### Describe your environment

WSL

Notation `v1.0.0-rc.4`

### LINKED ISSUE #660: Deprecate --signature-manifest flag from notation sign

Notation CLI

`notation sign` has a flag `--signature-manifest` to sign an artifact and store the signature using OCI artifact manifest. This flag is defined in [notation sign](https://github.com/notaryproject/notation/blob/main/specs/commandline/sign.md#experimental-sign-an-artifact-and-store-the-signature-using-oci-artifact-manifest).

[image-spec v1.1.0-rc.3](https://github.com/opencontainers/image-spec/releases/tag/v1.1.0-rc.3) removed OCI artifact manifest. This is a breaking change that we need to rethink when delivering a stable Notation v1.0.0.

Deprecate `--signature-manifest` flag from `notation sign` but still enable Notation users to read and verify the existing signatures stored with the OCI artifact manifest type in registries that comply with OCI Distribution Spec v1.1.0-RC.1. 
There might be some users who are using previous Notation releases. So Notation should still care for those old artifact manifests (signatures). 

N/A

### Any additional context?

_No response_

Two-Hearts:
In this PR:
1. Removed Notation's support of signing with OCI artifact manifest. (One can still consume signatures in this type.)
2.  Moved the Referrers API behind experimental as discussed in community meeting. This change takes effect on `sign`, `list`, `inspect`, and `verify` commands. When a user wants to try the Referrers API, they need to set both `NOTATION_EXPERIMENTAL=1` AND `--allow-referrers-api`; if the Referrers API is not supported, fallback to the Referrers tag schema automatically. Otherwise, by default, Notation would use the Referrers tag schema in all commands deterministically.
References: 
https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers
https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#referrers-tag-schema

This PR has been tested and:
resolves #659
resolves #660

### DIFF
--- a/cmd/notation/inspect.go
+++ b/cmd/notation/inspect.go
@@ -70,7 +70,7 @@ Example - Inspect signatures on an OCI artifact identified by a digest and outpu
   notation inspect --output json <registry>/<repository>@<digest>
 `
 	experimentalExamples := `
-Example - [Experimental] Inspect signatures on an OCI artifact identified by a digest using the Referrers API, if not supported, fallback to the Referrers tag schema
+Example - [Experimental] Inspect signatures on an OCI artifact identified by a digest using the Referrers API, if not supported (returns 404), fallback to the Referrers tag schema
   notation inspect --allow-referrers-api <registry>/<repository>@<digest>
 `
 	command := &cobra.Command{
@@ -95,7 +95,7 @@ Example - [Experimental] Inspect signatures on an OCI artifact identified by a d
 	opts.LoggingFlagOpts.ApplyFlags(command.Flags())
 	opts.SecureFlagOpts.ApplyFlags(command.Flags())
 	cmd.SetPflagOutput(command.Flags(), &opts.outputFormat, cmd.PflagOutputUsage)
-	command.Flags().BoolVar(&opts.allowReferrersAPI, "allow-referrers-api", false, "[Experimental] use the Referrers API to inspect signatures, if not supported, fallback to the Referrers tag schema")
+	cmd.SetPflagReferrersAPI(command.Flags(), &opts.allowReferrersAPI, cmd.PflagReferrersAPIInspectUsage)
 	experimental.HideFlags(command, experimentalExamples, []string{"allow-referrers-api"})
 	return command
 }
--- a/cmd/notation/inspect.go
+++ b/cmd/notation/inspect.go
@@ -61,17 +61,17 @@ func inspectCommand(opts *inspectOpts) *cobra.Command {
 	longMessage := `Inspect all signatures associated with the signed artifact.
 
 Example - Inspect signatures on an OCI artifact identified by a digest:
-	notation inspect <registry>/<repository>@<digest>
+  notation inspect <registry>/<repository>@<digest>
 
 Example - Inspect signatures on an OCI artifact identified by a tag  (Notation will resolve tag to digest):
-	notation inspect <registry>/<repository>:<tag>
+  notation inspect <registry>/<repository>:<tag>
 
 Example - Inspect signatures on an OCI artifact identified by a digest and output as json:
-	notation inspect --output json <registry>/<repository>@<digest>
+  notation inspect --output json <registry>/<repository>@<digest>
 `
 	experimentalExamples := `
 Example - [Experimental] Inspect signatures on an OCI artifact identified by a digest using the Referrers API, if not supported, fallback to the Referrers tag schema
-	notation inspect --allow-referrers-api <registry>/<repository>@<digest>
+  notation inspect --allow-referrers-api <registry>/<repository>@<digest>
 `
 	command := &cobra.Command{
 		Use:   "inspect [reference]",
--- a/cmd/notation/inspect.go
+++ b/cmd/notation/inspect.go
@@ -95,7 +95,7 @@ Example - [Experimental] Inspect signatures on an OCI artifact identified by a d
 	opts.LoggingFlagOpts.ApplyFlags(command.Flags())
 	opts.SecureFlagOpts.ApplyFlags(command.Flags())
 	cmd.SetPflagOutput(command.Flags(), &opts.outputFormat, cmd.PflagOutputUsage)
-	cmd.SetPflagReferrersAPI(command.Flags(), &opts.allowReferrersAPI, cmd.PflagReferrersAPIInspectUsage)
+	cmd.SetPflagReferrersAPI(command.Flags(), &opts.allowReferrersAPI, fmt.Sprintf(cmd.PflagReferrersUsageFormat, "inspect"))
 	experimental.HideFlags(command, experimentalExamples, []string{"allow-referrers-api"})
 	return command
 }
--- a/cmd/notation/inspect.go
+++ b/cmd/notation/inspect.go
@@ -14,6 +14,7 @@ import (
 	"github.com/notaryproject/notation-core-go/signature"
 	"github.com/notaryproject/notation-go/plugin/proto"
 	"github.com/notaryproject/notation-go/registry"
+	"github.com/notaryproject/notation/cmd/notation/internal/experimental"
 	"github.com/notaryproject/notation/internal/cmd"
 	"github.com/notaryproject/notation/internal/envelope"
 	"github.com/notaryproject/notation/internal/ioutil"
@@ -25,8 +26,9 @@ import (
 type inspectOpts struct {
 	cmd.LoggingFlagOpts
 	SecureFlagOpts
-	reference    string
-	outputFormat string
+	reference         string
+	outputFormat      string
+	allowReferrersAPI bool
 }
 
 type inspectOutput struct {
@@ -56,26 +58,34 @@ func inspectCommand(opts *inspectOpts) *cobra.Command {
 	if opts == nil {
 		opts = &inspectOpts{}
 	}
+	longMessage := `Inspect all signatures associated with the signed artifact.
+
+Example - Inspect signatures on an OCI artifact identified by a digest:
+	notation inspect <registry>/<repository>@<digest>
+
+Example - Inspect signatures on an OCI artifact identified by a tag  (Notation will resolve tag to digest):
+	notation inspect <registry>/<repository>:<tag>
+
+Example - Inspect signatures on an OCI artifact identified by a digest and output as json:
+	notation inspect --output json <registry>/<repository>@<digest>
+`
+	experimentalExamples := `
+Example - [Experimental] Inspect signatures on an OCI artifact identified by a digest using the Referrers API, if not supported, fallback to the Referrers tag schema
+	notation inspect --allow-referrers-api <registry>/<repository>@<digest>
+`
 	command := &cobra.Command{
 		Use:   "inspect [reference]",
 		Short: "Inspect all signatures associated with the signed artifact",
-		Long: `Inspect all signatures associated with the signed artifact.
-
-Example - Inspect signatures on an OCI artifact identified by a digest:
-  notation inspect <registry>/<repository>@<digest>
-
-Example - Inspect signatures on an OCI artifact identified by a tag  (Notation will resolve tag to digest):
-  notation inspect <registry>/<repository>:<tag>
-
-Example - Inspect signatures on an OCI artifact identified by a digest and output as json:
-  notation inspect --output json <registry>/<repository>@<digest>
-`,
+		Long:  longMessage,
 		Args: func(cmd *cobra.Command, args []string) error {
 			if len(args) == 0 {
 				return errors.New("missing reference")
 			}
 			opts.reference = args[0]
 			return nil
+		},
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			return experimental.CheckFlagsAndWarn(cmd, "allow-referrers-api")
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runInspect(cmd, opts)
@@ -85,6 +95,8 @@ Example - Inspect signatures on an OCI artifact identified by a digest and outpu
 	opts.LoggingFlagOpts.ApplyFlags(command.Flags())
 	opts.SecureFlagOpts.ApplyFlags(command.Flags())
 	cmd.SetPflagOutput(command.Flags(), &opts.outputFormat, cmd.PflagOutputUsage)
+	command.Flags().BoolVar(&opts.allowReferrersAPI, "allow-referrers-api", false, "[Experimental] use the Referrers API to inspect signatures, if not supported, fallback to the Referrers tag schema")
+	experimental.HideFlags(command, experimentalExamples, []string{"allow-referrers-api"})
 	return command
 }
 
@@ -98,7 +110,7 @@ func runInspect(command *cobra.Command, opts *inspectOpts) error {
 
 	// initialize
 	reference := opts.reference
-	sigRepo, err := getRemoteRepository(ctx, &opts.SecureFlagOpts, reference)
+	sigRepo, err := getRemoteRepository(ctx, &opts.SecureFlagOpts, reference, opts.allowReferrersAPI)
 	if err != nil {
 		return err
 	}
--- a/cmd/notation/list.go
+++ b/cmd/notation/list.go
@@ -52,7 +52,7 @@ func listCommand(opts *listOpts) *cobra.Command {
 	}
 	opts.LoggingFlagOpts.ApplyFlags(command.Flags())
 	opts.SecureFlagOpts.ApplyFlags(command.Flags())
-	cmd.SetPflagReferrersAPI(command.Flags(), &opts.allowReferrersAPI, cmd.PflagReferrersAPIListUsage)
+	cmd.SetPflagReferrersAPI(command.Flags(), &opts.allowReferrersAPI, fmt.Sprintf(cmd.PflagReferrersUsageFormat, "list"))
 	command.Flags().BoolVar(&opts.ociLayout, "oci-layout", false, "[Experimental] list signatures stored in OCI image layout")
 	experimental.HideFlags(command, "", []string{"allow-referrers-api", "oci-layout"})
 	return command
--- a/cmd/notation/list.go
+++ b/cmd/notation/list.go
@@ -16,9 +16,10 @@ import (
 type listOpts struct {
 	cmd.LoggingFlagOpts
 	SecureFlagOpts
-	reference string
-	ociLayout bool
-	inputType inputType
+	reference         string
+	allowReferrersAPI bool
+	ociLayout         bool
+	inputType         inputType
 }
 
 func listCommand(opts *listOpts) *cobra.Command {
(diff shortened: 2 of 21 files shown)


cmd/notation/sign.go
```diff
@@ -69,14 +61,14 @@ Example - Sign an OCI artifact stored in a registry and specify the signature ex
   notation sign --expiry 24h <registry>/<repository>@<digest>
 `
 	experimentalExamples := `
+Example - [Experimental] Sign an OCI artifact and store signature using the Referrers API, if not supported, fallback to the Referrers tag schema
```
JeyJeyGao, line 64:
```suggestion
Example - [Experimental] Sign an OCI artifact and store signature using the Referrers API. If it's not supported, fallback to the Referrers tag schema
```

cmd/notation/inspect.go
```diff
@@ -56,27 +58,35 @@ func inspectCommand(opts *inspectOpts) *cobra.Command {
...
+	notation inspect <registry>/<repository>:<tag>
 
 Example - Inspect signatures on an OCI artifact identified by a digest and output as json:
-  notation inspect --output json <registry>/<repository>@<digest>
-`,
+	notation inspect --output json <registry>/<repository>@<digest>
+`
+	experimentalExamples := `
+Example - [Experimental] Inspect signatures on an OCI artifact identified by a digest using the Referrers API, if not supported, fallback to the Referrers tag schema
+	notation inspect --allow-referrers-api <registry>/<repository>@<digest>
```
shizhMSFT, line 74: Example indentation should not be changed.

cmd/notation/list.go
```diff
@@ -43,16 +44,17 @@ func listCommand(opts *listOpts) *cobra.Command {
...
-			return experimental.CheckFlagsAndWarn(cmd, "oci-layout")
+			return experimental.CheckFlagsAndWarn(cmd, "allow-referrers-api", "oci-layout")
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runList(cmd.Context(), opts)
 		},
 	}
 	opts.LoggingFlagOpts.ApplyFlags(cmd.Flags())
 	opts.SecureFlagOpts.ApplyFlags(cmd.Flags())
+	cmd.Flags().BoolVar(&opts.allowReferrersAPI, "allow-referrers-api", false, "[Experimental] use the Referrers API to list signatures, if not supported, fallback to the Referrers tag schema")
```
shizhMSFT, line 55: Should we refactor it in `flags.go`?

cmd/notation/registry.go
```diff
@@ -100,24 +68,15 @@ func getRemoteRepositoryForSign(ctx context.Context, opts *SecureFlagOpts, refer
...
-	// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers
-	// 2. OCI image manifest uses the Referrers API and automatically fallback
-	// 	  to Referrers Tag Schema if Referrers API is not supported.
-	// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#referrers-tag-schema
-	if !ociImageManifest {
-		logger.Info("Use OCI artifact manifest to store signature")
-		// ping Referrers API
-		if err := pingReferrersAPI(ctx, remoteRepo); err != nil {
+	if experimental.IsDisabled() || !allowReferrersAPI {
+		logger.Info("By default, using the Referrers tag schema")
```
shizhMSFT, line 72: The prefix "By default, " seems to be redundant.

cmd/notation/registry.go
```diff
@@ -100,24 +68,15 @@ func getRemoteRepositoryForSign(ctx context.Context, opts *SecureFlagOpts, refer
...
-	// 1. OCI artifact manifest uses the Referrers API.
-	// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers
-	// 2. OCI image manifest uses the Referrers API and automatically fallback
-	// 	  to Referrers Tag Schema if Referrers API is not supported.
-	// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#referrers-tag-schema
-	if !ociImageManifest {
-		logger.Info("Use OCI artifact manifest to store signature")
-		// ping Referrers API
-		if err := pingReferrersAPI(ctx, remoteRepo); err != nil {
+	if experimental.IsDisabled() || !allowReferrersAPI {
```
shizhMSFT, line 71: Can we change it to `!experimental.IsDisabled() && allowReferrersAPI` and swap the `if` and `else` code block for easier if condition understanding?

cmd/notation/sign.go
```diff
@@ -94,13 +86,9 @@ Example - [Experimental] Sign an OCI artifact and use OCI artifact manifest to s
 			if opts.ociLayout {
 				opts.inputType = inputTypeOCILayout
 			}
-			return experimental.CheckFlagsAndWarn(cmd, "signature-manifest", "oci-layout")
+			return experimental.CheckFlagsAndWarn(cmd, "allow-referrers-api", "oci-layout")
```
sajayantony, line 89: Don't think we need API suffix in the flag. `--allow-referrers` would be my preference.

Two-Hearts: > Don't think we need API suffix in the flag. `--allow-referrers` would be my preference.

@sajayantony Just an open discussion, since the `referrers tag schema` also has the word `referrers` in its name, probably we should keep the `api` suffix in the flag? @shizhMSFT @yizha1 @FeynmanZhou Do you have any preferences on this one?

cmd/notation/inspect.go
Two-Hearts: Oh yeah, nice catch. My VS Code is using tab as default.

cmd/notation/list.go
Two-Hearts: Since it's an experimental flag, I think we could keep it here? (similar to flag `--oci-layout`). There are `experimental.CheckFlagsAndWarn` and `experimental.HideFlags` in the file as well, maybe moving the flag into flags.go is not as clear as the current way?

Two-Hearts: > Should we refactor it in `flags.go`?

Since it's an experimental flag, I think we could keep it here? (similar to flag `--oci-layout`). There are `experimental.CheckFlagsAndWarn` and `experimental.HideFlags` in the file as well, maybe moving the flag into flags.go is not as clear as the current way?

cmd/notation/verify.go
```diff
@@ -48,6 +49,9 @@ Example - Verify a signature on an OCI artifact identified by a tag  (Notation w
   notation verify <registry>/<repository>:<tag>
 `
 	experimentalExamples := `
+Example - [Experimental] Verify an OCI artifact using the Referrers API, if not supported, fallback to the Referrers tag schema
```
priteshbandi, line 52:
```suggestion
Example - [Experimental] Verify an OCI artifact using the Referrers API, if not supported (return's 404), to the Referrers tag schema
```

specs/commandline/inspect.md
```diff
@@ -33,11 +33,14 @@ Usage:
     notation inspect [flags] <reference>
   
 Flags:
-   -h, --help              help for describing the signature
-   -o, --output json       output on command line sets the output to json
-   -p, --password string   password for registry operations (default to $NOTATION_PASSWORD if not specified)
-       --plain-http        registry access via plain HTTP
-   -u, --username string   username for registry operations (default to $NOTATION_USERNAME if not specified)
+      --allow-referrers-api   [Experimental] use the Referrers API to inspect signatures, if not supported, fallback to the Referrers tag schema
```
priteshbandi, line 36:
```suggestion
      --allow-referrers-api   [Experimental] use the Referrers API to inspect signatures, if not supported (return's 404), fallback to the Referrers tag schema
```

cmd/notation/registry.go
```diff
@@ -100,24 +68,15 @@ func getRemoteRepositoryForSign(ctx context.Context, opts *SecureFlagOpts, refer
...
-	// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers
-	// 2. OCI image manifest uses the Referrers API and automatically fallback
-	// 	  to Referrers Tag Schema if Referrers API is not supported.
-	// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#referrers-tag-schema
-	if !ociImageManifest {
-		logger.Info("Use OCI artifact manifest to store signature")
-		// ping Referrers API
-		if err := pingReferrersAPI(ctx, remoteRepo); err != nil {
+	if !experimental.IsDisabled() && allowReferrersAPI {
+		logger.Info("Using the Referrers API, if not supported, automatically fallback to the Referrers tag schema")
```
priteshbandi, line 72:
```suggestion
		logger.Info("Trying to use referrers API")
```

cmd/notation/inspect.go
```diff
@@ -85,6 +95,8 @@ Example - Inspect signatures on an OCI artifact identified by a digest and outpu
 	opts.LoggingFlagOpts.ApplyFlags(command.Flags())
 	opts.SecureFlagOpts.ApplyFlags(command.Flags())
 	cmd.SetPflagOutput(command.Flags(), &opts.outputFormat, cmd.PflagOutputUsage)
+	command.Flags().BoolVar(&opts.allowReferrersAPI, "allow-referrers-api", false, "[Experimental] use the Referrers API to inspect signatures, if not supported, fallback to the Referrers tag schema")
```
priteshbandi, line 98: `command.Flags().BoolVar(&opts.allowReferrersAPI, "allow-referrers-api", false, "[Experimental] use the Referrers API to${operation} signatures, if not supported, fallback to the Referrers tag schema")` is repeated 3 time should we create a helper function for this?

test/e2e/suite/trustpolicy/registry_scope.go
```diff
@@ -16,7 +16,7 @@ var _ = Describe("notation trust policy registryScope test", func() {
 			vhost.SetOption(AddTrustPolicyOption("empty_registry_scope_trustpolicy.json"))
 
 			// test localhost:5000/test-repo
-			OldNotation().Exec("sign", artifact.ReferenceWithDigest()).MatchKeyWords(SignSuccessfully)
+			notation.Exec("sign", artifact.ReferenceWithDigest()).MatchKeyWords(SignSuccessfully)
```
priteshbandi, line 19: Why did we originally ran test using old release?

test/e2e/suite/trustpolicy/trust_store.go
```diff
@@ -22,12 +22,12 @@ var _ = Describe("notation trust policy trust store test", func() {
 	})
 
 	It("invalid trust store", func() {
-		Host(BaseOptions(), func(notation *utils.ExecOpts, _ *Artifact, vhost *utils.VirtualHost) {
+		Host(BaseOptionsWithExperimental(), func(notation *utils.ExecOpts, _ *Artifact, vhost *utils.VirtualHost) {
```
priteshbandi, line 25: Should we create another set of test without experimental flags?

cmd/notation/list.go
Two-Hearts: > Should we refactor it in `flags.go`?

Refactored into flags.go

test/e2e/suite/trustpolicy/registry_scope.go
Two-Hearts: We are testing `empty registry scope` in this test case. So yes, using the `OldNotation` should also work. Will change it back.

test/e2e/suite/trustpolicy/trust_store.go
Two-Hearts: We should, but IMO not in this PR. Our e2e `testdata` is still using OCI artifact manifest and referrers API as default. Changes of e2e test cases in this PR are only making the existing tests runnable again. We need another PR, maybe after rc.5, to update our e2e framework (there's a certain amount of work regarding this).

Two-Hearts: > Should we create another set of test without experimental flags?

We should, but IMO not in this PR. Our e2e `testdata` is still using OCI artifact manifest and referrers API as default. Changes of e2e test cases in this PR are only making the existing tests runnable again. We need another PR, maybe after rc.5, to update our e2e framework (there's a certain amount of work regarding this). /cc: @JeyJeyGao

Two-Hearts: I created a new issue to track this: https://github.com/notaryproject/notation/issues/668

cmd/notation/sign.go
sajayantony: As per call. This is to disambiguate between referrers tag schema vs API and hence having the api suffix makes this clearer. 
With that in mind, I agree that `--allow-referrers-api` LGTM.

internal/cmd/flags.go
```diff
@@ -93,6 +93,17 @@ var (
 		fs.StringArrayVarP(p, PflagUserMetadata.Name, PflagUserMetadata.Shorthand, nil, usage)
 	}
 
+	PflagReferrersAPI = &pflag.Flag{
+		Name: "allow-referrers-api",
+	}
+	PflagReferrersAPISignUsage    = "[Experimental] use the Referrers API to store signatures in the registry, if not supported (returns 404), fallback to the Referrers tag schema"
+	PflagReferrersAPIVerifyUsage  = "[Experimental] use the Referrers API to verify signatures, if not supported (returns 404), fallback to the Referrers tag schema"
+	PflagReferrersAPIListUsage    = "[Experimental] use the Referrers API to list signatures, if not supported (returns 404), fallback to the Referrers tag schema"
+	PflagReferrersAPIInspectUsage = "[Experimental] use the Referrers API to inspect signatures, if not supported (returns 404), fallback to the Referrers tag schema"
```
priteshbandi, line 102: nit: 
```
PflagReferrersUsageFormat = "[Experimental] use the Referrers API to %s signatures, if not supported (returns 404), fallback to the Referrers tag schema"
```
and then we do something like this in inspect.go 
```
cmd.SetPflagReferrersAPI(command.Flags(), &opts.allowReferrersAPI, fmt.Sprintf(cmd.PflagReferrersUsageFormat, "inspect"))
```

test/e2e/suite/trustpolicy/trusted_identity.go
```diff
@@ -76,12 +76,13 @@ var _ = Describe("notation trust policy trusted identity test", func() {
 				AddTrustPolicyOption("multiple_trusted_identity_trustpolicy.json"),
 				AddTrustStoreOption("e2e", filepath.Join(NotationE2ELocalKeysDir, "new_e2e.crt")),
 				AddTrustStoreOption("e2e", filepath.Join(NotationE2ELocalKeysDir, "e2e.crt")),
+				EnableExperimental(),
 			)
 
-			notation.Exec("verify", artifact1.ReferenceWithDigest(), "-v").
+			notation.Exec("verify", "--allow-referrers-api", artifact1.ReferenceWithDigest(), "-v").
```
priteshbandi, line 82: As discussed over slack please open an issue to fix this once rc5 is released.
```
Pritesh Bandi [11:34 PM] Hi Patrick - I was reviewing https://github.com/notaryproject/notation/pull/663/files#diff-719e0e8132d59c0c8f273651d38b7b382e16959ecc8de341080367a1e779bb0e
and have question on why do we need to update trust policy test to use experimental flag?

Patrick Zheng [11:42 PM]
Hi Pritesh, that's because some of those tests are expecting successful verification. Since the OldNotation is using the Referrers API by default for sign, whereas RC.5 logic is using the Referrers tag schema as default for verify. In this case, the Referrers tag schema does not exist, and thus, the test would fail with error no signature associated. To make these tests runnable again, we rather need to Sign with RC.5 notation OR Verify with the Referrers API (i.e., turning on the experimental and set --allow-referrers-api).
```

priteshbandi: https://github.com/notaryproject/notation/issues/668

Two-Hearts: > As discussed over slack please open an issue to fix this once rc5 is released.
> 
> ```
> Pritesh Bandi [11:34 PM] Hi Patrick - I was reviewing https://github.com/notaryproject/notation/pull/663/files#diff-719e0e8132d59c0c8f273651d38b7b382e16959ecc8de341080367a1e779bb0e
> and have question on why do we need to update trust policy test to use experimental flag?
> 
> Patrick Zheng [11:42 PM]
> Hi Pritesh, that's because some of those tests are expecting successful verification. Since the OldNotation is using the Referrers API by default for sign, whereas RC.5 logic is using the Referrers tag schema as default for verify. In this case, the Referrers tag schema does not exist, and thus, the test would fail with error no signature associated. To make these tests runnable again, we rather need to Sign with RC.5 notation OR Verify with the Referrers API (i.e., turning on the experimental and set --allow-referrers-api).
> ```

Thanks @priteshbandi, I've alreday created an issue regarding this: https://github.com/notaryproject/notation/issues/668. We will work on it after RC.5 and before v1.0.

cmd/notation/registry.go
```diff
@@ -68,13 +68,13 @@ func getRemoteRepository(ctx context.Context, opts *SecureFlagOpts, reference st
 		return nil, err
 	}
 
-	if experimental.IsDisabled() || !allowReferrersAPI {
-		logger.Info("By default, using the Referrers tag schema")
+	if !experimental.IsDisabled() && allowReferrersAPI {
+		logger.Info("Trying to use referrers API")
+	} else {
+		logger.Info("Using the Referrers tag schema")
```
shizhMSFT, line 74: nit:
```suggestion
		logger.Info("Using the referrers tag schema")
```

shizhMSFT (approved):
LGTM

JeyJeyGao (approved):
LGTM

vaninrao10 (approved):
LGTM

priteshbandi (approved):
LGTM

PR #664, opened 2023-05-13
fix: add number of arguments validation

priteshbandi:
fix: add number of arguments validation

---

This PR is continuation of https://github.com/notaryproject/notation/pull/649 

https://github.com/notaryproject/notation/pull/649#discussion_r1184665327
@JeyJeyGao I did try using [positional and Custom Arguments](https://github.com/spf13/cobra/blob/main/user_guide.md#positional-and-custom-arguments) but the error messages were not user friendly. Also, we needed to rewrite bunch of unit tests to support this.
```shell
./notation verify someImage -m "adasd=1" sadad=2
Error: accepts 1 arg(s), received 2
```


cmd/notation/list.go
```diff
@@ -33,8 +32,8 @@ func listCommand(opts *listOpts) *cobra.Command {
 		Short:   "List signatures of the signed artifact",
 		Long:    "List all the signatures associated with signed artifact",
 		Args: func(cmd *cobra.Command, args []string) error {
-			if len(args) == 0 {
-				return errors.New("no reference specified")
+			if err := arg.ValidateCount(args, 1, "missing reference"); err != nil {
+				return err
```
Two-Hearts, line 36: The cobra package has the following method: [ExactArgs](https://pkg.go.dev/github.com/spf13/cobra#ExactArgs). Therefore, we can have something like this:
```
Args: cobra.ExactArgs(1),
PreRunE: func(cmd *cobra.Command, args []string) error {
			if opts.ociLayout {
				opts.inputType = inputTypeOCILayout
			}
            opts.reference = args[0]
			return experimental.CheckFlagsAndWarn(cmd, "oci-layout")
		},
```
Similar applies to the other commands.

internal/arg/arg.go
Two-Hearts, line 1: If we are using `cobra.ExactArgs(x)` as suggested above, we don't need this file now.

cmd/notation/list.go
priteshbandi: I did try using [ositional and Custom Arguments](https://github.com/spf13/cobra/blob/main/user_guide.md#positional-and-custom-arguments) but the error messages were not user friendly. Also, we need to rewrite bunch of integ tests to support this.

```
./notation verify
Error: accepts 1 arg(s), received 0

./notation verify someImage -m "adasd=1" sadad=2
Error: accepts 1 arg(s), received 2
```

Two-Hearts: > I did try using [positional and Custom Arguments](https://github.com/spf13/cobra/blob/main/user_guide.md#positional-and-custom-arguments) but the error messages were not user friendly. Also, we would need to rewrite bunch of unit tests to support this because they verifies that passed arguments are stored in `opts`.
> 
> ```
> ./notation verify
> Error: accepts 1 arg(s), received 0
> 
> ./notation verify someImage -m "adasd=1" sadad=2
> Error: accepts 1 arg(s), received 2
> ```

I see, you are trying to customize the error messages here. Makes sense to me.

internal/arg/arg.go
```diff
@@ -0,0 +1,17 @@
+package arg
+
+import (
+	"errors"
+	"fmt"
+)
+
+// ValidateCount validates the arguments are of expected length.
+func ValidateCount(args []string, expLen int, missingErrMsg string) error {
```
Two-Hearts, line 9: nit:
I think this function should be renamed to `ValidateArgsCount`

internal/arg/arg.go
Two-Hearts, line 1: Since package `arg` is used in notation command line only, can we move it to `cmd/notation/internal`?

internal/arg/arg.go
priteshbandi: the usage would be `arg.ValidateCount`. IMO adding args to function name is redundant.

Two-Hearts: Oh yeah, that's true.

Two-Hearts (dismissed):
LGTM

Two-Hearts (approved):
LGTM

cmd/notation/internal/arg/arg.go
```diff
@@ -0,0 +1,17 @@
...
+	"fmt"
+)
+
+// ValidateCount validates the arguments are of expected length.
+func ValidateCount(args []string, expLen int, missingErrMsg string) error {
+	argsLength := len(args)
+	if argsLength < expLen {
+		return errors.New(missingErrMsg)
+	} else if argsLength > expLen {
+		return fmt.Errorf("expected %d argument(s) but found %d, arguments: %v", expLen, argsLength, args)
```
shizhMSFT, line 14: The error message is inconsistent with `cobra`: see https://github.com/spf13/cobra/blob/v1.7.0/args.go#L94

cmd/notation/inspect.go
```diff
@@ -71,8 +72,8 @@ Example - Inspect signatures on an OCI artifact identified by a digest and outpu
   notation inspect --output json <registry>/<repository>@<digest>
 `,
 		Args: func(cmd *cobra.Command, args []string) error {
-			if len(args) == 0 {
-				return errors.New("missing reference")
+			if err := arg.ValidateCount(args, 1, "missing reference"); err != nil {
```
shizhMSFT, line 75: Consider:
```suggestion
			if err := cobra.ExactArgs(1)(cmd, args); err != nil {
```

cmd/notation/internal/arg/arg.go
shizhMSFT, line 1: We've already have a package named `cmd/notation/internal/cmdutil`. You may put `args.go` there.

cmd/notation/inspect.go
```diff
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"github.com/notaryproject/notation/cmd/notation/internal/arg"
```
shizhMSFT, line 9: The import should not be in this import block

cmd/notation/inspect.go
priteshbandi: Please see https://github.com/notaryproject/notation/pull/664#discussion_r1194132149

PR #673, opened 2023-05-18
feat: added notation cert cleanup-test command

### LINKED ISSUE #647: Support clean up the source key and certificate generated by Notation

Notation CLI

`notation key delete` can only remove the key from the signing key list and `notation cert delete` can only remove the self-signed certificate from the trust store. This is by design since Notation doesn't support signing with [local keys](https://github.com/notaryproject/notation/issues/539) and managing local keys. 

Per discussion in https://github.com/notaryproject/notation/pull/606#discussion_r1154664392 and another issue #604 , users want to delete the source key and certificate generated by `notation cert generate-test` in a convenient way.  

Providing a flag `--cleanup` to `notation cert generate-test` to allow users to delete the specified source key and certificate generated by `notation cert generate-test`. This flag is only used to delete the test key and self-signed certificate. The keys and certificates that are not generated by Notation will not be able to be deleted with this flag. 

For example, delete a source key and cert generated by `notation cert generate-test "wabbit-networks.io"`:

```
$ notation cert generate-test --cleanup wabbit-networks.io
Deleted <key_name> and <cert_name> 
```

N/A

### Any additional context?

_No response_

Two-Hearts:
In this PR:
1. Adds `notation cert cleanup-test` command, this is to cleanup key-cert pairs created by `notation cert generate-test`. On success, both key and cert files are deleted; the self-signed cert is removed from trust store; the private key is removed from notation's signing key list.
2. This new command is behind `experimental` at the moment, because we may support signing with local keys in the future. Open to suggestions about whether we should move it out of 'experimental'.

Resolves #647.


cmd/notation/cert/cleanupTest.go
```diff
@@ -0,0 +1,97 @@
...
+			return errors.New("cleanup-test can only apply to non-external keys created by the generate-test command")
+		}
+		if keySuite.X509KeyPair == nil {
+			return errors.New("cleanup-test requires key pair files created by the generate-test command")
+		}
+		// delete the key file from keyPath
+		keyPath := keySuite.X509KeyPair.KeyPath
+		err = osutil.DeleteFile(keyPath)
+		if err != nil {
+			finalError = append(finalError, fmt.Errorf("cannot delete the key file from %q during cleanup-test, %v", keyPath, err))
```
JeyJeyGao, line 58: Why not print the error directly?

internal/osutil/file_test.go
```diff
@@ -2,19 +2,18 @@ package osutil
...
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"runtime"
 	"testing"
 )
 
 func validFileContent(t *testing.T, filename string, content []byte) {
-	b, err := ioutil.ReadFile(filename)
+	b, err := os.ReadFile(filename)
```
JeyJeyGao, line 12: Is this the unit test for ` ioutil.ReadFile`?

cmd/notation/cert/cleanupTest.go
JeyJeyGao: We may don't need to print an error if the file doesn't exist.

Two-Hearts: This PR and its related issue are pending discussions. Will address comments once everyone agrees on the design.

cmd/notation/cert/generateTest.go
```diff
@@ -40,20 +40,20 @@ func certGenerateTestCommand(opts *certGenerateTestOpts) *cobra.Command {
 	}
 	command := &cobra.Command{
 		Use:   "generate-test [flags] <common_name>",
-		Short: "Generate a test RSA key and a corresponding self-signed certificate.",
+		Short: "[Experimental] Generate a test RSA key and a corresponding self-signed certificate.",
```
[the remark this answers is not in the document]
sajayantony: +1 mainline scenarios for developing or testing should not be experimental.

cmd/notation/cert/cleanupTest.go
Two-Hearts: As we discussed offline, there are, in total, 4 deletions/removals in cleanup-test. We want all deletions/removals to be run even some of them may fail during this process. Eventually, we will print out all the succeeded operations followed by all the failed cases. If we return the error directly, then the user may need to call `notation cert cleanup-test` multiple times to remove all the related files which is not very user friendly.

cmd/notation/cert/generateTest.go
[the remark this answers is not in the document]
Two-Hearts: @priteshbandi @sajayantony Got it. I updated the code to reflect your suggestions.

cmd/notation/cert/cleanupTest.go
```diff
@@ -0,0 +1,116 @@
...
+type certCleanupTestOpts struct {
+	keyName string
+}
+
+func certCleanupTestCommand(opts *certCleanupTestOpts) *cobra.Command {
+	if opts == nil {
+		opts = &certCleanupTestOpts{}
+	}
+	command := &cobra.Command{
+		Use:   "cleanup-test <common_name>",
```
shizhMSFT, line 24: It seems the `notation cert cleanup-test` command is not behind an experimental flag yet.

test/e2e/suite/scenario/quickstart.go
```diff
@@ -22,7 +22,10 @@ var _ = Describe("notation quickstart E2E test", Ordered, func() {
 		if err != nil {
 			panic(err)
 		}
-		vhost.SetOption(AuthOption("", ""))
+		vhost.SetOption(
+			AuthOption("", ""),
+			EnableExperimental(),
```
shizhMSFT, line 27: Given `notation cert generate-test` is not an experimental feature as discussed in the above comments by others, do should we still need to `EnableExperimental()`?

test/e2e/suite/command/sign.go
```diff
@@ -66,7 +66,7 @@ var _ = Describe("notation sign", func() {
 	})
 
 	It("with specific key", func() {
-		Host(BaseOptions(), func(notation *utils.ExecOpts, artifact *Artifact, vhost *utils.VirtualHost) {
+		Host(BaseOptionsWithExperimental(), func(notation *utils.ExecOpts, artifact *Artifact, vhost *utils.VirtualHost) {
```
shizhMSFT, line 69: same here

cmd/notation/cert/cmd.go
```diff
@@ -1,6 +1,8 @@
 package cert
 
-import "github.com/spf13/cobra"
+import (
+	"github.com/spf13/cobra"
+)
```
shizhMSFT, line 5: nit: should revert the change

cmd/notation/cert/cleanupTest.go
```diff
@@ -0,0 +1,116 @@
...
+			return errors.New("cleanup-test requires key pair files created by the generate-test command")
+		}
+		// delete the key file from keyPath
+		keyPath := keySuite.X509KeyPair.KeyPath
+		err = osutil.DeleteFile(keyPath)
+		if err != nil {
+			if os.IsNotExist(err) {
+				fmt.Printf("Success: key file %q does not exist. It is already been cleaned up.\n", keyPath)
+			} else {
+				finalError = append(finalError, fmt.Errorf("cannot delete the key file %q during cleanup-test, %v", keyPath, err))
```
shizhMSFT, line 66: nit: instead of having a `finalError` slice. Should we print the error directly?
```suggestion
				fmt.Fprintf(os.Stderr, "Failed: cannot delete the key file %q during cleanup-test: %v\n", keyPath, err))
```

Two-Hearts: > nit: instead of having a `finalError` slice. Should we print the error directly?

@shizhMSFT I think the readability of the current printouts is higher. All succeeded cases are grouped at the top followed by all the failed cases grouped together as well. In this way, the printouts are very clear and easy to read. What do you think? 
/cc: @yizha1 @FeynmanZhou

cmd/notation/cert/cleanupTest.go
Two-Hearts: > It seems the `notation cert cleanup-test` command is not behind an experimental flag yet.

Yes, this is something that we need to discuss. Since we decided to not put `notation cert generate-test` behind `experimental`, I think we shouldn't put `notation cert cleanup-test` behind the flag either? This is to make the user experience more consistent before and after.

cmd/notation/cert/cleanupTest.go
shizhMSFT: I think the output should reflect the order of the execution.

cmd/notation/cert/cleanupTest.go
shizhMSFT: @Two-Hearts @yizha1 @FeynmanZhou Let's discuss it in the community meeting.

Two-Hearts:
Converted to draft until we are ready to discuss this again.

PR #674, opened 2023-05-19
update: renamed --plain-http to --insecure-registry

### LINKED ISSUE #623: Flag `--plain-http` didn't explicitly remind users the insecure connection to registries

Notation CLI

The flag `--plain-http` allows notation commands to establish an insecure connection to registries, which is implicit to users. This could be a security risk if users connect to an insecure registry.

This flag exists for the following commands:
- `notation sign`
- `notation verify`
- `notation list`
- `notation inspect`
- `notation login`

Update the description of the flag to explicitly inform users that this flag allows insecure connection to registries. for example:

`--plain-http`               Use HTTP protocol while connecting to registries. Use it only for testing purposes.

### How can we reproduce it?

Run `help` command

### Describe your environment

WSL

`v1.0.0-rc.3`

Two-Hearts:
This PR is based on discussions in issue #623 with following changes:
1. Renamed flag `--plain-http` to `--insecure-registry` for notation `sign`, `verify`, `list`, `inspect` and `login`.
2. Updated the flag's description to `use HTTP protocol while connecting to registries. Use it only for testing purposes`. Related specs are updated as well.

Resolves #623.

### DIFF
--- a/cmd/notation/common.go
+++ b/cmd/notation/common.go
@@ -31,27 +31,27 @@ var (
 		fs.StringVarP(p, flagPassword.Name, flagPassword.Shorthand, "", flagPassword.Usage)
 	}
 
-	flagPlainHTTP = &pflag.Flag{
-		Name:     "plain-http",
-		Usage:    "registry access via plain HTTP",
+	flagInsecureRegistry = &pflag.Flag{
+		Name:     "insecure-registry",
+		Usage:    "use HTTP protocol while connecting to registries. Use it only for testing purposes",
 		DefValue: "false",
 	}
-	setFlagPlainHTTP = func(fs *pflag.FlagSet, p *bool) {
-		fs.BoolVar(p, flagPlainHTTP.Name, false, flagPlainHTTP.Usage)
+	setFlagInsecureRegistry = func(fs *pflag.FlagSet, p *bool) {
+		fs.BoolVar(p, flagInsecureRegistry.Name, false, flagInsecureRegistry.Usage)
 	}
 )
 
 type SecureFlagOpts struct {
-	Username  string
-	Password  string
-	PlainHTTP bool
+	Username         string
+	Password         string
+	InsecureRegistry bool
 }
 
 // ApplyFlags set flags and their default values for the FlagSet
 func (opts *SecureFlagOpts) ApplyFlags(fs *pflag.FlagSet) {
 	setflagUsername(fs, &opts.Username)
 	setFlagPassword(fs, &opts.Password)
-	setFlagPlainHTTP(fs, &opts.PlainHTTP)
+	setFlagInsecureRegistry(fs, &opts.InsecureRegistry)
 	opts.Username = os.Getenv(defaultUsernameEnv)
 	opts.Password = os.Getenv(defaultPasswordEnv)
 }
--- a/specs/commandline/verify.md
+++ b/specs/commandline/verify.md
@@ -38,9 +38,9 @@ Flags:
        --allow-referrers-api         [Experimental] use the Referrers API to verify signatures, if not supported (returns 404), fallback to the Referrers tag schema
   -d,  --debug                       debug mode
   -h,  --help                        help for verify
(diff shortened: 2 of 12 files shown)


shizhMSFT (commented):
@yizha1 @Two-Hearts Should we just drop `--plain-http` in RC versions?

yizha1:
> @yizha1 @Two-Hearts Should we just drop `--plain-http` in RC versions?

RC-5 is the last RC release. The next is v1 release.

shizhMSFT:
> > @yizha1 @Two-Hearts Should we just drop `--plain-http` in RC versions?
> 
> RC-5 is the last RC release. The next is v1 release.

Then we should drop the `--plain-http` flag and can be added if needed. After `1.0.0`, we are not able to remove any non-experimental flag but we are able to add new flags.

cmd/notation/common.go
```diff
@@ -31,27 +31,36 @@ var (
 		fs.StringVarP(p, flagPassword.Name, flagPassword.Shorthand, "", flagPassword.Usage)
 	}
 
-	flagPlainHTTP = &pflag.Flag{
-		Name:     "plain-http",
-		Usage:    "registry access via plain HTTP",
+	flagInsecureRegistry = &pflag.Flag{
+		Name:     "insecure-registry",
+		Usage:    "use HTTP protocol while connecting to registries. Use it only for testing purposes. (alias: --plain-http)",
```
sajayantony, line 36:
```suggestion
		Usage:    "use HTTP protocol while connecting to registries. Should be used only for testing",
```

cmd/notation/common.go
```diff
@@ -31,27 +31,36 @@ var (
...
 	setflagUsername(fs, &opts.Username)
 	setFlagPassword(fs, &opts.Password)
-	setFlagPlainHTTP(fs, &opts.PlainHTTP)
+	setFlagInsecureRegistry(fs, &opts.InsecureRegistry)
+	fs.SetNormalizeFunc(aliasNormalizeFunc)
 	opts.Username = os.Getenv(defaultUsernameEnv)
 	opts.Password = os.Getenv(defaultPasswordEnv)
 }
+
+func aliasNormalizeFunc(f *pflag.FlagSet, name string) pflag.NormalizedName {
```
sajayantony, line 60: Recommend removing aliasing to avoid breaking post RC to GA. Also this should be used only for test either way.

specs/commandline/verify.md
```diff
@@ -38,9 +38,9 @@ Flags:
        --allow-referrers-api         [Experimental] use the Referrers API to verify signatures, if not supported (returns 404), fallback to the Referrers tag schema
   -d,  --debug                       debug mode
   -h,  --help                        help for verify
+       --insecure-registry           use HTTP protocol while connecting to registries. Use it only for testing purposes. (alias: --plain-http)
```
sajayantony, line 41: Need to update the wording if we remove the flag altogether.

cmd/notation/common.go
Two-Hearts: > Recommend removing aliasing to avoid breaking post RC to GA. Also this should be used only for test either way.

Got it. Updated the code to remove aliasing.

Two-Hearts:
> > > @yizha1 @Two-Hearts Should we just drop `--plain-http` in RC versions?
> > 
> > 
> > RC-5 is the last RC release. The next is v1 release.
> 
> Then we should drop the `--plain-http` flag and it can be added if needed. After `1.0.0`, we are not able to remove any non-experimental flag but we are able to add new flags.

Thanks @shizhMSFT, good point! I updated the code to reflect it.

cmd/notation/common.go
FeynmanZhou: Per our recent discussion around the Notation security best practice, we could remove the alias `--plain-http` entirely in RC.6.

yizha1:
Agree to drop `--plain-http` flag in upcoming RC release

priteshbandi (approved):
LGTM

PR #682, opened 2023-05-25
chore: updated warning printout logic for Sign with --allow-referrers-api flag

### LINKED ISSUE #683: Sign with --allow-referrers-api and --oci-layout flag shouldn't print the warning from allowReferrersAPI

Notation CLI

When user enables `--allow-referrers-api` and `--oci-layout`, the message `Warning: using the Referrers API to store signature. On success, must set the --allow-referrers-api flag to list, inspect, and verify the signature.` is still printed out.
The warning should be valid only when target artifact is stored in a registry.
If the user enables `--oci-layout`, i.e., target artifact is stored in an `OCI layout`, enabling `--allow-referrers-api` won't take any effect (this is because [oci.Store](https://github.com/oras-project/oras-go/blob/0e20275bbbd71f4619d9c1b9da150c4f586dff3b/content/oci/oci.go#L49) does not support Referrers API yet).

The warning is not printed out.

### How can we reproduce it?

`notation sign --oci-layout ----allow-referrers-api <ref>`

### Describe your environment

all environments

Notation CLI v1.0.0-rc.5

Two-Hearts:
This quick PR updates the printout logic for `sign`, when user enables `--allow-referrers-api`.
The warning should be valid only when target artifact is stored in a `registry`. 
If the user enables `--oci-layout`, i.e., target artifact is stored in an OCI layout, enabling `--allow-referrers-api` won't take any effect (this is because [oci.Store](https://github.com/oras-project/oras-go/blob/0e20275bbbd71f4619d9c1b9da150c4f586dff3b/content/oci/oci.go#L49) does not support Referrers API yet). In this case, we shouldn't print out the warning.

The fix: During `Sign`, mark flags `--oci-layout` and `--allow-referrers-api` mutually exclusive.

Resolves #683.

### DIFF
--- a/cmd/notation/sign.go
+++ b/cmd/notation/sign.go
@@ -100,6 +100,7 @@ Example - [Experimental] Sign an OCI artifact identified by a tag and referenced
 	cmd.SetPflagUserMetadata(command.Flags(), &opts.userMetadata, cmd.PflagUserMetadataSignUsage)
 	cmd.SetPflagReferrersAPI(command.Flags(), &opts.allowReferrersAPI, fmt.Sprintf(cmd.PflagReferrersUsageFormat, "sign"))
 	command.Flags().BoolVar(&opts.ociLayout, "oci-layout", false, "[Experimental] sign the artifact stored as OCI image layout")
+	command.MarkFlagsMutuallyExclusive("oci-layout", "allow-referrers-api")
 	experimental.HideFlags(command, experimentalExamples, []string{"allow-referrers-api", "oci-layout"})
 	return command
 }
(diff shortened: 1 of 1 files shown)


priteshbandi (dismissed):
LGTM

priteshbandi:
Should we also add a check that `--oci-layout` and `--allow-referrers-api` cannot be used simultaneously?

Two-Hearts:
> Should we also add a check that `--oci-layout` and `--allow-referrers-api` cannot be used simultaneously?

@priteshbandi Then that would be an error for the user instead of dropping the `--allow-referrers-api` flag silently. (I'm okay with both of the options though.) /cc: @shizhMSFT @yizha1 @FeynmanZhou

priteshbandi (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #684, opened 2023-05-25
update: based on spec, updated messages of notation key command

Two-Hearts:
This PR is based on spec changes in here: https://github.com/notaryproject/notation/pull/606.


cmd/notation/key.go
```diff
@@ -243,10 +240,11 @@ func deleteKeys(ctx context.Context, opts *keyDeleteOpts) error {
 	}
 
 	// write out
-	for _, name := range deletedNames {
-		if prevDefault == name {
-			fmt.Printf("%s: unmarked as default\n", name)
```
priteshbandi, line 248: We also need to show that notation removed key that was marked defualt.

Two-Hearts: Oh yeah, based on the [PR](https://github.com/notaryproject/notation/pull/606) of the spec change, we decided to remove this message.
This change was suggested by @yizha1 and @FeynmanZhou.

yizha1: If delete one key and it is default, then print out

`Removed default key <key_name> from Notation signing key list. The source key still exists.`

if delete multiple keys and one of the keys is default, then print out
```
Removed the following keys from Notation signing key list. The source keys still exist.
<key_name_1>
<key_name_2> (default)
<key_name_2>
```

Two-Hearts: Had another discussion, will add the logic back. @priteshbandi @shizhMSFT

FeynmanZhou: Agree with @priteshbandi we should remind users when it is a default key. @Two-Hearts Could you pls also update related [CLI Spec](https://github.com/notaryproject/notation/blob/main/specs/commandline/key.md#remove-two-keys-from-notation-signing-key-list) in this PR?

specs/commandline/key.md
```diff
@@ -130,6 +130,11 @@ Removed <key_name> from Notation signing key list. The source key still exists.
 
 ```text
 Removed <key_name> from Notation signing key list. The source key still exists.
+
```
shizhMSFT, line 133: Missing ending ` ``` `

shizhMSFT (approved):
LGTM

PR #690, opened 2023-05-26
fix: fixed e2e test after dependency bump up

Two-Hearts:
This PR fixed the e2e test case, which was failed due to error message updates in notation-go.


priteshbandi (approved):
LGTM

JeyJeyGao (approved):
LGTM

PR #703, opened 2023-06-05
fix: fix the issue with getting credentials for `docker.io`

### LINKED ISSUE #697: `notation login` fails to detect existing credentials for `docker.io`

Notation CLI

On a machine without credentials store, after `docker login docker.io` succeeds, `notation login` with the same credential should also succeed with a warning.
image

However, it fails to detect the existing credential and throws an error:
image

The bug is at here:
https://github.com/notaryproject/notation/blob/ebfb9ef707996e1dc11898db8b90faa8e8816ae6/cmd/notation/login.go#L118-L128

where `credsStore.Get()` attempts to get the credential for `docker.io`, while the credential for `docker.io` is stored in the entry `https://index.docker.io/v1/` (expected behavior), hence `credsStore.Get("docker.io")` returns an empty credential `savedCred` that does not match the provided credential `cred`.

A mapping from `docker.io` to `https://index.docker.io/v1/` is required.

`notation login` should be able to detect the existing credentials for `docker.io`

Expected output:
image

### How can we reproduce it?

On a machine without credentials store,
1. Run
```bash
docker login docker.io -u $username -p $password
```

2. Then run
```bash
notation login docker.io -u $username -p $password
```

### Describe your environment

Windows 11/powershell

Version:     1.0.0-rc.7
Go version:  go1.20.4
Git commit:  ebfb9ef707996e1dc11898db8b90faa8e8816ae6

Wwwsylvia:
Fixes #697

### DIFF
--- a/cmd/notation/login.go
+++ b/cmd/notation/login.go
@@ -115,7 +115,13 @@ func runLogin(ctx context.Context, opts *loginOpts) error {
 		// ErrPlaintextPutDisabled returned by Login() indicates that the
 		// credential is validated but is not saved because there is no native
 		// credentials store available
-		if savedCred, err := credsStore.Get(ctx, registryName); err != nil || savedCred != cred {
+		credKeyName := registryName
+		if registryName == "docker.io" {
+			// it's as expected that credentials for "docker.io" are stored
+			// under the key "https://index.docker.io/v1/"
+			credKeyName = "https://index.docker.io/v1/"
+		}
+		if savedCred, err := credsStore.Get(ctx, credKeyName); err != nil || savedCred != cred {
 			if err != nil {
 				// if we fail to get the saved credential, log a warning
 				// but do not throw the GET error, as the error could be


cmd/notation/login.go
```diff
@@ -115,7 +115,13 @@ func runLogin(ctx context.Context, opts *loginOpts) error {
 		// ErrPlaintextPutDisabled returned by Login() indicates that the
 		// credential is validated but is not saved because there is no native
 		// credentials store available
-		if savedCred, err := credsStore.Get(ctx, registryName); err != nil || savedCred != cred {
+		credKeyName := registryName
+		if registryName == "docker.io" {
+			// it's as expected that credentials for "docker.io" are stored
+			// under the key "https://index.docker.io/v1/"
+			credKeyName = "https://index.docker.io/v1/"
```
Two-Hearts, line 122: How about saving `https://index.docker.io/v1/` in a `const` at the top of the file?

priteshbandi: Are there any other usecase where registry hostname might be different from how key name? If so probably we should create map for this mapping.

shizhMSFT: @priteshbandi This is the only special case.

@Wwwsylvia We may promote this key mapping in the future versions of `oras-credentials-go`.

shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

cmd/notation/login.go
Wwwsylvia: @Two-Hearts I think it's clearer and more concise to use the literal here.

Wwwsylvia: @shizhMSFT Good point!

Issue #696, opened 2023-05-30
`desktop.exe` credential store is not supported in WSL


JeyJeyGao:
### What is the areas you experience the issue in?

Notation CLI

### What is not working as expected?

Docker Desktop was installed on Windows, but I cannot login by `notation login` command
![image](https://github.com/notaryproject/notation/assets/43160897/56bf3b02-6a9d-423d-982c-5baf62f924c6)

### What did you expect to happen?

I should login with `notation login` command successfully because I have installed the Docker Desktop with desktop.exe credential store.

Here is my docker config:
```json
{
        "auths": {
        },
        "credsStore": "desktop.exe"
}
```

Here is the files under notation config directory:

$ ~/.config/notation
```sh
total 28K
drwx------ 2 jj jj 4.0K Mar 13 16:34 localkeys
-rw-r--r-- 1 jj jj  253 Jan 11 15:10 pluginkeys.json
drwxr-xr-x 5 jj jj 4.0K May 29 16:56 plugins
-rw------- 1 jj jj  440 May 26 13:43 signingkeys.json
-rw-r--r-- 1 jj jj  360 Dec 28 17:28 trustpolicy.json
drwx------ 3 jj jj 4.0K Mar 13 16:34 truststore
```

### How can we reproduce it?

1. Install Docker Desktop on Windows
2. run `notation login` in WSL

### Describe your environment

Windows 11 with WSL2 Ubuntu 22.04

### What is the version of your Notation CLI or Notation Library?

Notation v1.0.0-rc.7

shizhMSFT:
The current behavior is that for `notation login`, the `oras-credentials-go` library only takes look at the notation config for credential helper. If there is no credential helper, it will detect a default one even if the credential helper is set in the docker config.

The bug should be fixed at the `oras-credentials-go` side.

Wwwsylvia:
Related issue on `oras-credentials-go`: https://github.com/oras-project/oras-credentials-go/issues/71

yizha1:
Thanks @jimmyraywv 

`Notation login` will use credentials-store to securely store the credentials if authenticated with registries successfully.  The current order of detecting credentials store is `Notation config -> Docker config -> OS default credentials store`. This issue is a bug of `notation login` command, that it failed to detect the credentials store if it was not found in notation config. However, it works for `notation sign/verify/list` commands since these commands just read credentials if they exist.

From the requirement point of view, 
[...]

PR #712, opened 2023-06-08
fix: `login` and `logout` will leverage docker config and os default store

Wwwsylvia:
Fixes: #696 

---
## Behaviors of this PR
- If the notation config has auth configuration
  - Use the notation config for credentials GET, PUT and DELETE
- If the notation config does not have auth configuration, or the the notation config file does not exist
  - If the docker config has auth configuration
    - Use the notation config for credentials GET, PUT and DELETE
  - If there is no authentication configured in the docker config, or the the docker config file does not exist
    - If there is a platform-default native keychain installed
      - Use the platform-default native keychain for credentials GET, PUT and DELETE
    - If there is no platform-default native keychain installed
      - GET will always return an empty credential
      - PUT will always fail (ErrPlaintextPutDisabled)
      - DELETE will result in no operation

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/notaryproject/notation-go v1.0.0-rc.6
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.1.0-rc3
-	github.com/oras-project/oras-credentials-go v0.1.1
+	github.com/oras-project/oras-credentials-go v0.1.2-0.20230609082027-b60b11bcba14
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/cobra v1.7.0
 	github.com/spf13/pflag v1.0.5
@@ -17,7 +17,6 @@ require (
 
 require (
 	github.com/Azure/go-ntlmssp v0.0.0-20221128193559-754e69321358 // indirect
-	github.com/docker/docker-credential-helpers v0.7.0 // indirect
 	github.com/fxamacker/cbor/v2 v2.4.0 // indirect
 	github.com/go-asn1-ber/asn1-ber v1.5.4 // indirect
 	github.com/go-ldap/ldap/v3 v3.4.4 // indirect
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/notaryproject/notation-go v1.0.0-rc.6
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.1.0-rc3
-	github.com/oras-project/oras-credentials-go v0.1.2-0.20230609082027-b60b11bcba14
+	github.com/oras-project/oras-credentials-go v0.1.2-0.20230609092929-bf5244c85f2a
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/cobra v1.7.0
 	github.com/spf13/pflag v1.0.5
--- a/internal/auth/credentials.go
+++ b/internal/auth/credentials.go
@@ -15,15 +15,30 @@ func NewCredentialsStore() (credentials.Store, error) {
 		return nil, fmt.Errorf("failed to load config file: %w", err)
 	}
 
+	// use notation config
 	opts := credentials.StoreOptions{AllowPlaintextPut: false}
-	primaryStore, err := credentials.NewStore(configPath, opts)
+	notationStore, err := credentials.NewStore(configPath, opts)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create credential store from config file: %w", err)
 	}
+	if notationStore.IsAuthConfigured() {
+		return notationStore, nil
+	}
 
-	fallbackStore, err := credentials.NewStoreFromDocker(opts)
+	// use docker config
+	dockerStore, err := credentials.NewStoreFromDocker(opts)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create credential store from docker config file: %w", err)
 	}
-	return credentials.NewStoreWithFallbacks(primaryStore, fallbackStore), nil
+	if dockerStore.IsAuthConfigured() {
+		return dockerStore, nil
+	}
+
+	// detect platform-default native store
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -30,3 +30,5 @@ require (
 	golang.org/x/sync v0.2.0 // indirect
 	golang.org/x/sys v0.8.0 // indirect
 )
+
+replace github.com/oras-project/oras-credentials-go => github.com/Wwwsylvia/oras-credentials-go v0.0.0-20230608042921-c9e5bc5c8687
```
Two-Hearts, line 34: Just a tag, this will be removed once dependency is updated.

internal/auth/credentials.go
```diff
@@ -16,14 +16,24 @@ func NewCredentialsStore() (credentials.Store, error) {
...
+	if !notationStore.IsAuthConfigured() && !dockerStore.IsAuthConfigured() {
+		// if both notation config and docker config do not have auth
+		// configuration, fallback to the platform-default credentials helper
+		// (if available)
+		osDefaultStore, ok := credentials.NewDefaultNativeStore()
+		if ok {
+			return credentials.NewStoreWithFallbacks(notationStore, dockerStore, osDefaultStore), nil
+		}
+	}
+	return credentials.NewStoreWithFallbacks(notationStore, dockerStore), nil
```
JeyJeyGao, line 38: If credential store is missing for notationStore, dockerStore, do we need to fallback to osDefaultStore?

Wwwsylvia: Yes, see line 29 - 37.

Wwwsylvia:
Binary for testing: https://github.com/Wwwsylvia/notation/releases/tag/v1.0.8-sylvia

internal/auth/credentials.go
JeyJeyGao:
```go
func (cfg *Config) IsAuthConfigured() bool {
	return cfg.credentialsStore != "" ||
		len(cfg.credentialHelpers) > 0 ||
		len(cfg.authsCache) > 0
}
```
If the `notationStore` and `dockerStore` only configure the `credentialHelpers` field, and `credentialsStore` is empty, `IsAuthConfigured` will return ture. In this case,  line 29 - 37 will be skipped and run line 38. Will line 38 also include osDefaultStore since the credentialsStore is currently empty?

internal/auth/credentials.go
```diff
@@ -16,14 +16,24 @@ func NewCredentialsStore() (credentials.Store, error) {
...
+
+	if !notationStore.IsAuthConfigured() && !dockerStore.IsAuthConfigured() {
+		// if both notation config and docker config do not have auth
+		// configuration, fallback to the platform-default docker credentials
+		// helper (if available)
+		osDefaultStore, ok := credentials.NewDefaultNativeStore()
+		if ok {
+			return credentials.NewStoreWithFallbacks(notationStore, dockerStore, osDefaultStore), nil
+		}
+	}
```
priteshbandi, line 37: Why cant we return defualt trust store here when we know both notation's and docker's are not configured?

internal/auth/credentials.go
Wwwsylvia: No, and I don't think we should fallback to `osDefaultStore` in this case.
IMHO, as long as there is any auth configuration in the config file, even if it's just `credHelpers`, we should honor that. Maybe the user just wanted to authenticate with a few specific registries and block others? 🤔

internal/auth/credentials.go
Wwwsylvia: Good point! I think it makes sense to return the default store here.

internal/auth/credentials.go
JeyJeyGao: Okey, it is reasonable.

JeyJeyGao: In this case, if user want to login, the error message will still be `credential store wasn't installed`, but the user may have installed the credential store?

JeyJeyGao: Is it possible that user installed Docker on linux without credential store, so the docker config `credsStore` field is empty. The user try to login with Notation, and Notation asked the user to install credential store. After the user installed the credential store, the docker config `credsStore` field is still empty, and user try to do Notation login again but still get the same error. I'm not sure that is it a valid case.

Wwwsylvia: > Is it possible that user installed Docker on linux without credential store, so the docker config credsStore field is empty. The user try to login with Notation, and Notation asked the user to install credential store. After the user installed the credential store, the docker config credsStore field is still empty, and user try to do Notation login again but still get the same error. I'm not sure that is it a valid case.

I think in this case the user should just configure the creds store in the notation config file or the docker config file. 😀

shizhMSFT:
It seems that we were on the wrong track as we were figuring out the auth from the perspective of `oras` instead of `notation`.

For the consideration of `notation`, should we think about the following decision table?

ID | Is Notation config auth configured? | Is Docker config auth configured? | Decision
-- | -- | -- | --
1 | No | No | Detect default native store for GET, PUT, and DELETE
2 | Yes | No | Use Notation config only for GET, PUT, and DELETE. If plaintext cred is configured but cred store is not configured, fail on PUT.
3 | No | Yes | Use docker config only for GET, PUT, and DELETE. If plaintext cred is configured but cred store is not configured, fail on PUT.
4 | Yes | Yes | Use Notation config only for GET, PUT, and DELETE. If plaintext cred is configured but cred store is not configured, fail on PUT.

Here are detailed descriptions:
1. No auth configured at all. In this case, we should detect and use the default native store for better OOBE.
2. Notation auth config is configured but docker is not. This is a typical scenario where `docker` is not installed. In this case, we should use the notation auth config only.
3. Notation auth config is not configured but docker is. This is a typical scenario where `docker` is installed and the user want to share the docker auth with notation.
4. Both notation auth config and docker auth config are configured. This is a typical scenario that the user want to use `notation` as an isolated and independent component from docker.

shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

Issue #718, opened 2023-06-21
Add Golang lint to GitHub Actions for static Go code formatting scanning


FeynmanZhou:
### Is your feature request related to a problem?

N/A

### What solution do you propose?

[Golang lint](https://github.com/golangci/golangci-lint) is a fast Go linters runner. It runs linters in parallel, uses caching, supports yaml config. It’s used to make sure code adheres to common Go practices. It supports enable many different types of linters. 

I think we can introduce it to Notation to enable static code formatting scanning in each pull request. 

### What alternatives have you considered?

N/A

tungbq:
Hi, I'm interested in the CI task like this. Do you need help? Can I pick up this task?

FeynmanZhou:
Hi @tungbq , thanks for your interest! Yes, we can assign this issue to you.

tungbq:
**Updates:**
- Research the https://github.com/golangci/golangci-lint-action official GitHub action for [golangci-lint](https://github.com/golangci/golangci-lint)
 - Add `golangci-lint-action` to the CI workflows as it's the official GitHub action for [golangci-lint](https://github.com/golangci/golangci-lint) from its authors. The action runs [golangci-lint](https://github.com/golangci/golangci-lint) and reports issues from linters.
 - Name the new workflow as `golangci-lint-action` in `.github/workflows/go-lint.yml`
 - Current behaviour:
   - Run `golangci-lint` on every PR
   - Report only new issue generated from the PR
   - Beside of PR events, there has an option in the trigger event to allow us manually run the workflow if needed (`workflow_dispatch:`)

**Created new PR #723 for this task.**

ogulcanaydogan:
Hi! I noticed this issue has been open for a while and the previous PR (#723) was auto-closed by the stale bot. Is @tungbq still working on this?

If not, I'd be happy to pick this up. My approach would be:
1. Add a `golangci-lint-action` GitHub Actions workflow based on PR #723's approach
2. Update to the latest `golangci-lint` version and action version
3. Ensure it runs on PRs and catches common Go issues

Let me know if this is available to work on!

yizha1:
Hi @ogulcanaydogan, thanks for your interest in this issue. Please feel free to proceed.

PR #723, opened 2023-06-24
feat: introduce golang ci lint action

tungbq:
**Resolves:** #718 

**Background (mentioned in the issue #718 ):**

> [Golang lint](https://github.com/golangci/golangci-lint) is a fast Go linters runner. It runs linters in parallel, uses caching, supports yaml config. It’s used to make sure code adheres to common Go practices. It supports enable many different types of linters.

**Changes:**
 - Add `golangci-lint-action` to the CI workflows as it's the official GitHub action for [golangci-lint](https://github.com/golangci/golangci-lint) from its authors. The action runs [golangci-lint](https://github.com/golangci/golangci-lint) and reports issues from linters.
 - Name the new workflow as `golangci-lint` in `.github/workflows/go-lint.yml`
 - Current behaviour:
   - Run `golangci-lint` on every PR
   - Report only new issue generated from the PR
   - Beside of PR events, there has an option in the trigger event to allow us manually run the workflow if needed (`workflow_dispatch:`)

**Testing:**
- This was tested on my fork repository
- For example, https://github.com/tungbq/notation/pull/4


.github/workflows/go-lint.yml
```diff
@@ -0,0 +1,25 @@
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.20'
+          cache: false
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: latest
+          only-new-issues: true
```
shizhMSFT, line 25: Since this is the first time we introduce the CI lint, we should sweep the code for any existing issues and resolve them. Therefore, we need to set `only-new-issues` to `false`, which is the default value.

shizhMSFT (changes requested):
This PR decreases the code coverage for `17.04%`. Could you take a look?

.github/workflows/go-lint.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.20'
+          cache: false
```
shizhMSFT, line 20: If we use a fixed version of golang, which is recommended, we can resuse the cache for perf and cost.

.github/workflows/go-lint.yml
JeyJeyGao: This issue that caused by Zot V2.0.0-rc.5 update will be resolved in #727

JeyJeyGao:
This issue that caused by Zot V2.0.0-rc.5 update will be resolved in https://github.com/notaryproject/notation/pull/727

> This PR decreases the code coverage for `17.04%`. Could you take a look?

.github/workflows/go-lint.yml
tungbq: Thanks for your suggestion. I updated in the latest commit

.github/workflows/go-lint.yml
tungbq: Updated. Using the default value to catch all the existing issues

tungbq:
> This issue that caused by Zot V2.0.0-rc.5 update will be resolved in #727
> 
> > This PR decreases the code coverage for `17.04%`. Could you take a look?

Thanks for pointing this @JeyJeyGao

Issue #741, opened 2023-07-12
Notation to sign and verify arbitrary data


priteshbandi:
### Is your feature request related to a problem?

We would like to use Notation to sign and verify any arbitrary data and keep using the fine grain control it provides using trust policy and plugin model.

### What solution do you propose?

Notation support sign and verify any arbitrary data 
```
notation sign-blob ./somefile.txt
OR
notation sign ./somefile.txt --blob
```

cc: @yizha1 @toddysm @shizhMSFT @patrickzheng200 @gokarnm @iamsamirzon

Two-Hearts:
Thanks @priteshbandi, it's on my list, will share a PoC once it's ready, and we could start from there.

priteshbandi:
Thanks @patrickzheng200 for proactively working on this.
Do you have ECD (Estimated Completion Date) for PoC? Also, can you share any early draft around CLI interface?

gokarnm:
Thanks @patrickzheng200 ! There would be some spec work associated for this feature, covering changes (if any) for payload format, signature storage on disk, and trust policy.

yizha1:
@gokarnm @priteshbandi I remember we had some discussions around local artifacts/arbitrary files/signature storage at least one year ago. At that time, we tracked the features in roadmap repo, I did some searches and find related issues like
- [[User Story] - notation sign/verify Offline/Local artifacts](https://github.com/notaryproject/roadmap/issues/15)
- [Local storage of Signatures](https://github.com/notaryproject/roadmap/issues/46)

There were also some discussions and investigations during implementing feature of "sign/verify local OCI artifact", which is currently an experimental feature.  But we didn't continue the discussion recently since it is not in v1 scope any more.

This could be a good topic for community meeting, we can discuss it further on the plan. Does it make sense?

FeynmanZhou:
I agree that this is a significant feature request for Notation and has been discussed several times. This request was also captured in [Requirement (scenario-1-local-build-sign-validate) ](https://github.com/notaryproject/notaryproject/blob/main/requirements/scenarios.md#) and [Goals](https://github.com/notaryproject/notaryproject/blob/main/requirements/requirements.md#goals).

Before having an ECD for this issue, let's figure out the end-to-end scenarios, how it works with trust policy, have an approachable design proposal and CLI Spec, and split the work items. 

Let's discuss it in the next community meeting.

FeynmanZhou:
[...]

PR #765, opened 2023-08-21
spec: add spec for sign/verify an arbitrary file using Notation

Two-Hearts:
This PR adds specs for sign and verify on arbitrary file.

### DIFF
--- a/specs/commandline/sign.md
+++ b/specs/commandline/sign.md
@@ -26,7 +26,7 @@ The file content, i.e. the file blob, is signed.
 Upon successful signing, the generated signature is stored to user specified signature path in file system. The output message is printed out as following:
 
 ```text
-Successfully signed <target_file>
+Successfully signed <target_file>, stored signature at <signature_file>
 ```
 
 ## Outline
@@ -178,7 +178,7 @@ notation sign --file --signature <signature_path> <target_file_path>
 An example of a successful signing:
 ```console
 $ notation sign --file --signature ./mySignature.sig ./myFile.txt 
-Successfully signed ./myFile.txt
+Successfully signed ./myFile.txt, stored signature at ./mySignature.sig
 ```
 
 ### [Experimental] Sign container images stored in OCI layout directory
--- a/specs/commandline/sign.md
+++ b/specs/commandline/sign.md
@@ -5,6 +5,7 @@
 Use `notation sign` to sign artifacts.
 
 ### Sign an OCI artifact stored in registry
+
 Always sign artifact using digest(`@sha256:...`) rather than a tag(`:v1`) because tags are mutable and a tag reference can point to a different artifact than the one signed. If a tag is used, notation resolves the tag to the `digest` before signing.
 
 Upon successful signing, the generated signature is pushed to the registry and associated with the signed OCI artifact. The output message is printed out as following:
@@ -21,6 +22,7 @@ Successfully signed <registry>/<repository>@<digest>
 ```
 
 ### Sign an arbitrary file stored in file system
+
 The file content, i.e. the file blob, is signed.
 
 Upon successful signing, the generated signature is stored to user specified signature path in file system. The output message is printed out as following:
@@ -41,7 +43,7 @@ Flags:
        --allow-referrers-api        [Experimental] use the Referrers API to store signatures in the registry, if not supported (returns 404), fallback to the Referrers tag schema
   -d,  --debug                      debug mode
   -e,  --expiry duration            optional expiry that provides a "best by use" time for the artifact. The duration is specified in minutes(m) and/or hours(h). For example: 12h, 30m, 3h20m
-       --file                       enable signing a file's content, if set, the reference argument is the file path in the file system (required if --signature is set)
+       --file                       enable signing file content, if set, the reference argument is the file path in the file system (required if --signature is set)
   -h,  --help                       help for sign
        --id string                  key id (required if --plugin is set). This is mutually exclusive with the --key flag
        --insecure-registry          use HTTP protocol while connecting to registries. Should be used only for testing
@@ -166,7 +168,8 @@ Successfully signed localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da
 ```
 
 ### Sign an arbitrary file located in file system
-Notation supports signing a file located in user's file system. The file's content (blob) gets signed and the generated signature is stored in the same file system.
+
+Notation supports signing files located in file systems.
 ```shell
 # Prerequisites:
 # A default signing key is configured using CLI "notation key"
@@ -182,11 +185,11 @@ notation sign --file --signature <signature_path> <target_file_path>
 ```
 Examples of successful signing:
 ```console
-$ notation sign --file ./myFile.txt 
-Successfully signed ./myFile.txt and created signature at ./myFile.sig
+$ notation sign --file myFile.txt 
+Successfully signed myFile.txt and created signature at myFile.sig
 
 $ notation sign --file --signature ./mySignature.sig ./myFile.txt 
-Successfully signed ./myFile.txt and created signature at ./mySignature.sig
+Successfully signed myFile.txt and created signature at mySignature.sig
 ```
 
 ### [Experimental] Sign container images stored in OCI layout directory
--- a/specs/commandline/sign.md
+++ b/specs/commandline/sign.md
@@ -26,7 +26,7 @@ The file content, i.e. the file blob, is signed.
 Upon successful signing, the generated signature is stored to user specified signature path in file system. The output message is printed out as following:
 
 ```text
-Successfully signed <target_file>, stored signature at <signature_file>
+Successfully signed <target_file> and created signature at <signature_file>
 ```
 
 ## Outline
@@ -183,10 +183,10 @@ notation sign --file --signature <signature_path> <target_file_path>
 Examples of successful signing:
 ```console
 $ notation sign --file ./myFile.txt 
-Successfully signed ./myFile.txt, stored signature at ./myFile.sig
+Successfully signed ./myFile.txt and created signature at ./myFile.sig
 
 $ notation sign --file --signature ./mySignature.sig ./myFile.txt 
-Successfully signed ./myFile.txt, stored signature at ./mySignature.sig
+Successfully signed ./myFile.txt and created signature at ./mySignature.sig
 ```
 
 ### [Experimental] Sign container images stored in OCI layout directory
--- a/specs/commandline/sign.md
+++ b/specs/commandline/sign.md
@@ -4,7 +4,8 @@
 
 Use `notation sign` to sign artifacts.
 
-Signs an OCI artifact stored in the registry. Always sign artifact using digest(`@sha256:...`) rather than a tag(`:v1`) because tags are mutable and a tag reference can point to a different artifact than the one signed. If a tag is used, notation resolves the tag to the `digest` before signing.
+### Sign an OCI artifact stored in registry
+Always sign artifact using digest(`@sha256:...`) rather than a tag(`:v1`) because tags are mutable and a tag reference can point to a different artifact than the one signed. If a tag is used, notation resolves the tag to the `digest` before signing.
 
 Upon successful signing, the generated signature is pushed to the registry and associated with the signed OCI artifact. The output message is printed out as following:
 
@@ -19,6 +20,15 @@ Warning: Always sign the artifact using digest(`@sha256:...`) rather than a tag(
 Successfully signed <registry>/<repository>@<digest>
 ```
 
+### Sign an arbitrary file stored in file system
+The file content, i.e. the file blob, is signed.
+
+Upon successful signing, the generated signature is stored to user specified signature path in file system. The output message is printed out as following:
+
+```text
+Successfully signed <target_file>
+```
+
 ## Outline
 
 ```text
@@ -31,6 +41,7 @@ Flags:
        --allow-referrers-api        [Experimental] use the Referrers API to store signatures in the registry, if not supported (returns 404), fallback to the Referrers tag schema
   -d,  --debug                      debug mode
   -e,  --expiry duration            optional expiry that provides a "best by use" time for the artifact. The duration is specified in minutes(m) and/or hours(h). For example: 12h, 30m, 3h20m
+       --file                       enable signing a file located in file system, if set, the reference argument is the file path (required if --signature is set)
   -h,  --help                       help for sign
        --id string                  key id (required if --plugin is set). This is mutually exclusive with the --key flag
        --insecure-registry          use HTTP protocol while connecting to registries. Should be used only for testing
@@ -39,6 +50,7 @@ Flags:
   -p,  --password string            password for registry operations (default to $NOTATION_PASSWORD if not specified)
        --plugin string              signing plugin name. This is mutually exclusive with the --key flag
        --plugin-config stringArray  {key}={value} pairs that are passed as it is to a plugin, refer plugin's documentation to set appropriate values.
+       --signature string           output path of generated signature when signing a file (required if --file is set)
        --signature-format string    signature envelope format, options: "jws", "cose" (default "jws")
   -u,  --username string            username for registry operations (default to $NOTATION_USERNAME if not specified)
   -m,  --user-metadata stringArray  {key}={value} pairs that are added to the signature payload
@@ -151,6 +163,22 @@ An example for a successful signing:
 $ notation sign localhost:5000/net-monitor:v1
 Warning: Always sign the artifact using digest(`@sha256:...`) rather than a tag(`:v1`) because tags are mutable and a tag reference can point to a different artifact than the one signed.
 Successfully signed localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
+```
+
+### Sign an arbitrary file located in file system
+Notation supports signing a file located in user's file system. The file's content (blob) gets signed and the generated signature is stored in the same file system.
+```shell
+# Prerequisites:
+# A default signing key is configured using CLI "notation key"
+
+# Use flag "--file" to enable signing a file
+# Use flag "--signature" to specify path where the generated signature is stored
+notation sign --file --signature <signature_path> <target_file_path>
+```
+An example of a successful signing:
+```console
+$ notation sign --file --signature ./mySignature.sig ./myFile.txt 
+Successfully signed ./myFile.txt
 ```
 
 ### [Experimental] Sign container images stored in OCI layout directory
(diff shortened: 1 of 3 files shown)


specs/commandline/verify.md
```diff
@@ -173,6 +190,48 @@ Warning:  Always verify the artifact using digest(@sha256:...) rather than a tag
...
+2. If user ignores the `--scope` flag, then the [global trust policy](https://github.com/notaryproject/specifications/blob/main/specs/trust-store-trust-policy.md#registry-scopes-constraints) is used as default.
+```shell
+# Prerequisites: Both target file and signatures are stored in user's file system
+
+# Use flag "--file" to enable verifying a file
+# Use flag "--signature" to speicfy path where the signatures are stored
+# The global trust policy is used by default
+notation verify --file --signature ./mySignature1.sig --signature ./mySignature2.sig ./myFile.txt
+
+export NOTATION_EXPERIMENTAL=1
```
priteshbandi, line 223: Why do we want this to be experimental?

specs/commandline/sign.md
```diff
@@ -31,6 +41,7 @@ Flags:
        --allow-referrers-api        [Experimental] use the Referrers API to store signatures in the registry, if not supported (returns 404), fallback to the Referrers tag schema
   -d,  --debug                      debug mode
   -e,  --expiry duration            optional expiry that provides a "best by use" time for the artifact. The duration is specified in minutes(m) and/or hours(h). For example: 12h, 30m, 3h20m
+       --file                       enable signing a file located in file system, if set, the reference argument is the file path (required if --signature is set)
```
priteshbandi, line 44: We should also support reading data from stdin or as an parameter ?

specs/commandline/verify.md
```diff
@@ -37,14 +52,16 @@ Usage:
...
        --max-signatures int          maximum number of signatures to evaluate or examine (default 100)
        --oci-layout                  [Experimental] verify the artifact stored as OCI image layout
   -p,  --password string             password for registry operations (default to $NOTATION_PASSWORD if not specified)
        --plugin-config stringArray   {key}={value} pairs that are passed as it is to a plugin, if the verification is associated with a verification plugin, refer plugin documentation to set appropriate values
-       --scope string                [Experimental] set trust policy scope for artifact verification, required and can only be used when flag "--oci-layout" is set
+       --scope string                [Experimental] set trust policy scope for artifact verification, required when flag "--oci-layout" is set, can only be used when "--oci-layout" or "--file" is set
   -u,  --username string             username for registry operations (default to $NOTATION_USERNAME if not specified)
+       --signature stringArray       path of signatures when verifying a file, required and used if and only if the target file is stored in file system
```
priteshbandi, line 64: What will be output if user passed multiple signatures? Also, how can user consume that programmatically ?

specs/commandline/sign.md
```diff
@@ -153,6 +165,30 @@ Warning: Always sign the artifact using digest(`@sha256:...`) rather than a tag(
...
 ```
 
+### Sign an arbitrary file located in file system
+Notation supports signing a file located in user's file system. The file's content (blob) gets signed and the generated signature is stored in the same file system.
+```shell
+# Prerequisites:
+
+# Use flag "--file" to enable signing a file
+# The generated signature is saved at the same dir as the target file with name
```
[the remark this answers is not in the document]
Two-Hearts: @priteshbandi It's `abc.sig`, corresponding example is also provided in the same PR.

specs/commandline/sign.md
```diff
@@ -153,6 +165,30 @@ Warning: Always sign the artifact using digest(`@sha256:...`) rather than a tag(
...
+# <target_file>.sig
+notation sign --file <target_file_path>
+
+# Use flag "--file" to enable signing a file
+# Use flag "--signature" to specify path where the generated signature is stored
+notation sign --file --signature <signature_path> <target_file_path>
```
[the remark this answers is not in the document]
Two-Hearts: > What will happen if I form command like `notation sign --file <target_file_path> --signature <signature_path>`, will it succeed or fail?

This command will succeed.

> And should we use `--blob` instead of `--file`?

It was actually called `--blob`, but on a second thought I changed it to `--file`. The reason is that a user who wants to sign/verify a file might looking for a `--file` flag instead of `--blob`. In fact, they don't have to understand what a `blob` is in their scenario. All they need to know is they are signing/verifying a file.

specs/commandline/verify.md
Two-Hearts: Because `--scope` is an existing flag in the `notation v1.0.0 verify command`, and it is marked as `experimental` at the moment.

specs/commandline/verify.md
```diff
@@ -37,14 +52,16 @@ Usage:
...
   -h,  --help                        help for verify
        --insecure-registry           use HTTP protocol while connecting to registries. Should be used only for testing
        --max-signatures int          maximum number of signatures to evaluate or examine (default 100)
        --oci-layout                  [Experimental] verify the artifact stored as OCI image layout
   -p,  --password string             password for registry operations (default to $NOTATION_PASSWORD if not specified)
        --plugin-config stringArray   {key}={value} pairs that are passed as it is to a plugin, if the verification is associated with a verification plugin, refer plugin documentation to set appropriate values
-       --scope string                [Experimental] set trust policy scope for artifact verification, required and can only be used when flag "--oci-layout" is set
+       --scope string                [Experimental] set trust policy scope for artifact verification, required when flag "--oci-layout" is set, can only be used when "--oci-layout" or "--file" is set
```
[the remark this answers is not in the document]
Two-Hearts: @priteshbandi This is an existing flag. It was initially introduced for verifying artifact as `oci-layout`. In verifying a `file` scenario, it's only needed when the file and the signature are in user's file system. In this case, the user has two choices, they are mentioned in this PR (verify.md) as well, please take a look there.

specs/commandline/sign.md
Two-Hearts: @priteshbandi I think we could discuss this further, but it's kind of another topic than `sign/verify a file` in this PR (another flag, maybe `--stdin`, needs to be introduced in our future iterations).

specs/commandline/verify.md
Two-Hearts: It's very similar to our current behavior. Currently, if there are multiple signatures attached to the target artifact in the registry, Notation would verify those signatures one by one.
For verifying a file, signatures are still verified one by one. Again, example on multiple signatures is provided in this PR. Please take a look there.

specs/commandline/verify.md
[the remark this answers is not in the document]
priteshbandi: Yes but `registryScope` doesn't mean anything if user is verifying a blob/file.

Also as per https://github.com/notaryproject/specifications/blob/main/specs/trust-store-trust-policy.md#registry-scopes-constraints `List of one or more fully qualified repository URIs. The repository URI MUST NOT contain the asterisk character *.`

specs/commandline/sign.md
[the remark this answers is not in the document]
Two-Hearts: @priteshbandi Correct me if I'm wrong, you would like a flag to sign arbitrary things, by things I mean it can be a file, it can be data from stdin, etc.
But I think if we want to support data from stdin, then we do need a new flag for it instead of sharing the same flag with `file`. Here was the proposal from @shizhMSFT when we brought in `--oci-layout`: https://github.com/notaryproject/notation/pull/601#discussion_r1150810148. Basically, we are following the practice of `one type per flag` here. This design aligns with what we already have.
Given above, the `--blob` becomes a bit general. In the future, if we want to support sign/verify a descriptor, then we would introduce a `--descriptor` flag. If the digest of this descriptor belongs to a `blob`, then there's ambiguity, user might confuse on which flag they should use, `--blob` or `--descriptor`.
Coming back to the name `--file`, I think we can explicitly say that it sign/verify the `content` of a file in the flag's description of the CLI.

specs/commandline/verify.md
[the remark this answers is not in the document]
Two-Hearts: @priteshbandi Agree. Yes, we should have a spec change in the https://github.com/notaryproject/specifications repo as well especially for the `trust policy` part.

specs/commandline/verify.md
Two-Hearts: @priteshbandi Yes, you are right, I'm not marking this new feature as experimental. Only the flag `--scope` is experimental here.

specs/commandline/sign.md
Two-Hearts: Tracking this discussion here: https://github.com/notaryproject/notation/pull/765#discussion_r1301119361

specs/commandline/verify.md
Two-Hearts: That's a good point. Should we still follow the current behavior though? In other words, we still claim that as long as 1 signature passed the verification, the whole verification succeeded. @yizha1 @FeynmanZhou Do you have any comment on this one?

specs/commandline/verify.md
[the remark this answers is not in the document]
Two-Hearts: @priteshbandi I will add the related spec changes into the specification's repo. Here's the suggestion from @yizha1 on the spec changes: https://github.com/notaryproject/notation/issues/741#issuecomment-1686351235

specs/commandline/sign.md
```diff
@@ -19,6 +20,15 @@ Warning: Always sign the artifact using digest(`@sha256:...`) rather than a tag(
 Successfully signed <registry>/<repository>@<digest>
 ```
 
+### Sign an arbitrary file stored in file system
+The file content, i.e. the file blob, is signed.
```
shizhMSFT, line 24: nit: there usually is an empty line.

specs/commandline/sign.md
```diff
@@ -31,6 +41,7 @@ Flags:
        --allow-referrers-api        [Experimental] use the Referrers API to store signatures in the registry, if not supported (returns 404), fallback to the Referrers tag schema
   -d,  --debug                      debug mode
   -e,  --expiry duration            optional expiry that provides a "best by use" time for the artifact. The duration is specified in minutes(m) and/or hours(h). For example: 12h, 30m, 3h20m
+       --file                       enable signing a file's content, if set, the reference argument is the file path in the file system (required if --signature is set)
```
shizhMSFT, line 44:
```suggestion
       --file                       enable signing file content, if set, the reference argument is the file path in the file system (required if --signature is set)
```

specs/commandline/sign.md
shizhMSFT, line 24: We probably need to reword this sentence or remove it.

specs/commandline/sign.md
```diff
@@ -153,6 +165,30 @@ Warning: Always sign the artifact using digest(`@sha256:...`) rather than a tag(
 Successfully signed localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
 ```
 
+### Sign an arbitrary file located in file system
+Notation supports signing a file located in user's file system. The file's content (blob) gets signed and the generated signature is stored in the same file system.
```
shizhMSFT, line 169:
```suggestion
Notation supports signing files located in file systems.
```

shizhMSFT: File systems can be local (e.g. `ntfs`, `ext4`, `apfs`) or remote (`cifs`) or anything. Since the signed content and the generated signatures are files, they can be on any file systems and can be on different file systems.

specs/commandline/sign.md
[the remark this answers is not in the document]
shizhMSFT: @Two-Hearts Shouldn't it be `abc.txt.sig`?

The issue is very simple that not all files have an extension. If the target file is `LICENSE`, what's the signature name?

specs/commandline/sign.md
```diff
@@ -153,6 +165,30 @@ Warning: Always sign the artifact using digest(`@sha256:...`) rather than a tag(
...
+notation sign --file <target_file_path>
+
+# Use flag "--file" to enable signing a file
+# Use flag "--signature" to specify path where the generated signature is stored
+notation sign --file --signature <signature_path> <target_file_path>
+```
+Examples of successful signing:
+```console
+$ notation sign --file ./myFile.txt 
+Successfully signed ./myFile.txt and created signature at ./myFile.sig
```
shizhMSFT, line 186: `./` should be removed from the output.

specs/commandline/sign.md
```diff
@@ -153,6 +165,30 @@ Warning: Always sign the artifact using digest(`@sha256:...`) rather than a tag(
...
+
+$ notation sign --file --signature ./mySignature.sig ./myFile.txt 
+Successfully signed ./myFile.txt and created signature at ./mySignature.sig
```
shizhMSFT, line 189: same here for `./`

specs/commandline/verify.md
priteshbandi: So does that mean initially we will only support wildcard trust policy?

specs/commandline/verify.md
priteshbandi: I think this will require more discussion and to start with we can support only one signature.

specs/commandline/sign.md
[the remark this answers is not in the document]
Two-Hearts: I see. Will make the change.

specs/commandline/verify.md
Two-Hearts: @priteshbandi We will also support user specified trust policy, I'm working on specs in the specification repo to make this concept clear.

specs/commandline/verify.md
Two-Hearts: I see. Then as a start, we will make `--signature` a string type instead of stringArray.

specs/commandline/verify.md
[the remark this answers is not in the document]
Two-Hearts: @priteshbandi Answering them below:
> This will also require updates to [descriptor](https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#payload) which is being signing.

To be more precise, this is the update to the signature `payload`. 
And yes, I'm working on the spec in the [specifications](https://github.com/notaryproject/specifications) repo to introduce non-OCI-signature-specifications which includes `signature payload`, `signature storage`, `trust policy`, and any change on `sign and verification workflow`. This work is in progress, and I will create the PR once it's ready. 

> ```
> {
>   "targetArtifact": {
>     "mediaType": "sbom/example", 
>     "digest": "sha256:9834876dcfb05cb167a5c24953eba58c4ac89b1adf57f28f2f9d09af107ee8f0",
>     "size": 32654 <-- 
>   }
> }
> ```
> 
> * What would be the media type ?

Good point. We should allow users to pick their own media type. We could do this by:
`notation sign --file myFile:sbom/example`
i.e., using ':' to delimit the file path and its media type. What do you think?
/cc: @shizhMSFT 

> * Should we give control to user on which hashing algo to use in `digest` calculation?

Currently, I'm using [opencontainers/go-digest](https://pkg.go.dev/github.com/opencontainers/go-digest@v1.0.0#FromReader) to compute the digest. The algorithm is SHA256, which is the primary storage digest (https://pkg.go.dev/github.com/opencontainers/go-digest#Canonical).

> * `size`, is this the size of actual content that we are trying to sign?

Yes, since the actual content is signed, the size corresponds to the size of the actual content.

> ##### Signing process look like this?
> 1. Notation Reads the content that needs be signed (can be read from file or stdin or flag value)
> 2. Notation creates descriptor from the content, which needs following operation
> 3. Calculate hash/digest and size of content.
> 4. Calculate hash/digest of descriptor. This part will be either done by plugin or by notation depending upon plugin type.
> 5. Sign the hash/digest of descriptor. This would be done by plugin
> 6. return detached signature to user.
> 
> _Here we are calculating digest twice instead of once (ideal). Should we worry about this?_ Thinking more we will probably need this to support user defined metadata.

I don't think we should worry about this? Because even for the current Notation, we create the OCI descriptor out of the target OCI artifact, then sign/verify the descriptor.

> ##### For verification
> 1. Notation reads the detached signature
> 2. Verify the signature is valid (authenticity, integrity, expiry, etc checks)
> 3. Read the descriptor from signature.
> 4. Notation Reads the content to be verified
> 5. Calculate hash/digest of content and its size
> 6. Verify that ash/digest of content and its size matches with the descriptor stored in signature.

Yes, it reuses what we have in the current Notation workflow. The changes for sign and verify are actually small in our code base.

Two-Hearts:
Closing this PR as the spec work needs more discussions: https://github.com/notaryproject/notation/discussions/767. We need to have spec changes in the [specifications](https://github.com/notaryproject/specifications) repo, then CLI spec changes in Notation.

specs/commandline/verify.md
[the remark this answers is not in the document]
priteshbandi: > > * What would be the media type ?
> 
> Good point. We should allow users to pick their own media type. We could do this by: `notation sign --file myFile:sbom/example` i.e., using ':' to delimit the file path and its media type. We should have a default if user doesn't pass their own media type, we should define this default in the specifications repo though. What do you think? /cc: @shizhMSFT
> Good point. We should allow users to pick their own media type. 
 
IMO, for future extensibility and robustness, we should control the media type.



> > * Should we give control to user on which hashing algo to use in `digest` calculation?
> 
> Currently, I'm using [opencontainers/go-digest](https://pkg.go.dev/github.com/opencontainers/go-digest@v1.0.0#FromReader) to compute the digest. The algorithm is SHA256, which is the primary storage digest (https://pkg.go.dev/github.com/opencontainers/go-digest#Canonical).

 IMO, we can decide for sane default but allow user to control hashing algo. Depending upon usecase and compliance need user might want to use different hashing algos.

> > ##### Signing process look like this?
> > 
> > 1. Notation Reads the content that needs be signed (can be read from file or stdin or flag value)
> > 2. Notation creates descriptor from the content, which needs following operation
> > 3. Calculate hash/digest and size of content.
> > 4. Calculate hash/digest of descriptor. This part will be either done by plugin or by notation depending upon plugin type.
> > 5. Sign the hash/digest of descriptor. This would be done by plugin
> > 6. return detached signature to user.
> > 
> > _Here we are calculating digest twice instead of once (ideal). Should we worry about this?_ Thinking more we will probably need this to support user defined metadata.
> 
> I don't think we should worry about this? Because even for the current Notation, we create the OCI descriptor out of the target OCI artifact, then sign/verify the descriptor.
True, we could have directly signed pre-calculated digest(without rehashing). Lets discuss this in meeting

priteshbandi:
We area also missing a usecase where user might want to pass recalculated digest to notation. Reason can be privacy or compliance need.

specs/commandline/verify.md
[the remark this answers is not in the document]
shizhMSFT: @priteshbandi Since this PR is closed, could you move the discussions to https://github.com/notaryproject/notation/discussions/767?

PR #794, opened 2023-10-09
fix: fix trust policy import

### LINKED ISSUE #773: Policy import should check if a file exists first

It would be better to fast fail if a trust policy file doesn't exist

For example:
```
$ notation policy import trustpolicy.json                                                                     
Existing trust policy configuration found, do you want to overwrite it? [y/N] y
Error: failed to read trust policy file: open trustpolicy.json: no such file or directory
```

If this is not actionable anyway, no point asking the user the question in the first place

If file doesn't exist:
```
$ notation policy import trustpolicy.json                                                                     
Error: failed to read trust policy file: open trustpolicy.json: no such file or directory
```

If file exists:
```
$ notation policy import trustpolicy.json                                                                     
Existing trust policy configuration found, do you want to overwrite it? [y/N] y
...
```

n/a

### Any additional context?

_No response_

Two-Hearts:
Currently, `notation policy import mypolicy.json` would pop up confirmation before checking whether `mypolicy.json` file exists or valid. 
This PR checks the `mypolicy.json` first and fast fail if the file does not exist or is invalid. If the file exists and valid, then pop up the confirmation to the end user. 

Resolves #773.

### DIFF
--- a/cmd/notation/policy/import.go
+++ b/cmd/notation/policy/import.go
@@ -53,21 +53,6 @@ Example - Import trust policy configuration from a file:
 }
 
 func runImport(command *cobra.Command, opts importOpts) error {
-	// optional confirmation
-	if !opts.force {
-		if _, err := trustpolicy.LoadDocument(); err == nil {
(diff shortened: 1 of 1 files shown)


JeyJeyGao (approved):
LGTM

PR #809, opened 2023-10-27
doc: update plugin spec

FeynmanZhou:
Update notation plugin spec based on #549 and [PRD](https://hackmd.io/1nQa69DeROyqrO_IB5yGZQ?view#Notation-plugin-management-PRD)

Added new commands and examples including:

- notation plugin install     Install a plugin
- notation plugin ninstall   Uninstall a plugin

Updated the output of `notation plugin list`.

### DIFF
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -65,7 +65,9 @@ Usage:
   notation plugin upgrade [flags] <plugin_source>
 
 Flags:
-  -h, --help                    help for upgrade
+  -h, --help                     help for upgrade
+      --plugin-name string       plugin name 
+      --plugin-version           plugin version           
       --sha256sum string         must match SHA256 of the plugin source                 
 ```
 
@@ -161,7 +163,19 @@ To install a plugin, use "notation plugin install".
 
 ### Upgrade a plugin from URL
 
-Upgrade a Notation plugin from a remote shared address and verify the plugin checksum. Notation only supports upgrade plugins from an HTTPS URL.
+When upgrading a Notation plugin from GitHub release page, Notation upgrades the plugin to the latest version by default.  
+
+```
+$ notation plugin upgrade --plugin-name <plugin-name>
+```
+
+When upgrading a Notation plugin from GitHub release page, users can also specify a plugin version to upgrade.
+
+```
+$ notation plugin upgrade --plugin-version <plugin-version>
+```
+
+Upgrade a Notation plugin from a remote shared address (e.g, object storage) and verify the plugin checksum. Notation only supports upgrade a plugin from an HTTPS URL.
 
 ```shell
 $ notation plugin upgrade --sha256sum <digest> <URL>
@@ -172,7 +186,7 @@ $ notation plugin upgrade --sha256sum <digest> <URL>
 Upgrade a Notation plugin from a registry. Users can verify the plugin's signature with `notation verify` before the plugin installation.
 
 ```shell
-$ notation plugin upgrade <registry>/<repository>@<digest>
+$ notation plugin upgrade --plugin-name <plugin-name>
 ```
 
 ### Uninstall a plugin
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -76,8 +76,16 @@ Aliases:
 
 ### Install a plugin from file system
 
+Install a Notation plugin from file system and verify the plugin checksum.
+
 ```shell
-$ notation plugin install --file <file_path> --checksum <digest>
+$ notation plugin install --source file <file_path> --checksum <digest>
+```
+
+If the user doesn't specify the `--source` flag, Notation will install the plugin from file system by default.
+
+```shell
+$ notation plugin install <file_path> --checksum <digest>
 ```
 
 Upon successful execution, the plugin is copied to Notation's plugin directory. The name and version of the installed plugin is displayed as follows. 
@@ -94,11 +102,15 @@ Error: failed to install the plugin, <plugin name> already installed
 
 ### Install a plugin from URL
 
+Install a Notation plugin from a remote shared address and verify the plugin checksum.
+
 ```shell
-$ notation plugin install --source <URL> --checksum <digest>
+$ notation plugin install --source url <URL> --checksum <digest>
 ```
 
 ### Install a plugin as an OCI artifact from a registry (for future iteration)
+
+Install a Notation plugin from a registry. Users can verify the plugin's signature with `notation verify` before the plugin installation.
 
 ```shell
 $ notation plugin install --source registry <artifact_reference>
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -47,11 +47,13 @@ Usage:
   notation plugin install [flags] <--file|--url> <plugin_source>
 
 Flags:
-      --file                    install plugin from a file in file system
-  -h, --help                    help for install
-  -f, --force                   force the installation of a plugin
-      --sha256sum string        must match SHA256 of the plugin source
-      --url                     install plugin from an HTTPS URL
+  -d, --debug              debug mode
+      --file               install plugin from a file in file system
+      --force              force the installation of a plugin
+  -h, --help               help for install
+      --sha256sum string   must match SHA256 of the plugin source
+      --url                install plugin from an HTTPS URL
+  -v, --verbose            verbose mode
 
 Aliases:
   install, add
@@ -112,8 +114,8 @@ To view a list of installed plugins, use "notation plugin list"
 If the plugin version is lower than the existing plugin, Notation will return an error message and will not start installation. Users can use a flag `--force` to skip plugin version check and force the installation.
 
 ```console
-Error: failed to install the plugin: <plugin-name>. The installing version <a.b.c> is lower than the existing plugin. 
-To view a list of installed plugins, use "notation plugin list". It is not recommended to install older versions. Use "--force" to force the installation if you want to install an old version in a certain scenario.
+Error: failed to install the plugin: <plugin-name>. The installing version <a.b.c> is lower than the existing plugin with version <x.y.z>.
+It is not recommended to install an older version. Use "--force" to force the installation if you want to do so. To view a list of installed plugins, use "notation plugin list"
 ```
 ### Install a plugin from URL
 
@@ -146,7 +148,7 @@ If the plugin is not found, an error is returned showing the syntax for the plug
 
 ```shell
 Error: unable to find plugin <plugin_name>. 
-To view a list of installed plugins, use "notation plugin list".
+To view a list of installed plugins, use "notation plugin list"
 ```
 
 ### List installed plugins
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -105,7 +105,9 @@ Successfully installed plugin <plugin name>, version <x.y.z>
 If the plugin directory does not exist, it will be created. When an existing plugin is detected, it fails to install and returns the error as follows. Users can use a flag `--force` to skip existence check and force the installation.
 
 ```console
-Error: failed to install the plugin, <plugin_name> already installed
+Error: failed to install the plugin, <plugin_name> already installed. 
+To view a list of installed plugins, use "notation plugin list".
+To force the installation, use a flag `--force`.
 ```
 
 If the entered plugin checksum digest doesn't match the published checksum, Notation will return an error message and will not start installation.
@@ -152,8 +154,9 @@ Error: failed to upgrade the plugin, <plugin name> version should be higher than
 
 If the plugin does not exist, Notation will return an error message and will not start upgrade.
 
-```
-Error: failed to upgrade the plugin, <plugin name> version 
+```console
+Error: failed to upgrade the plugin, <plugin name> does not exist.
+To install a plugin, use "notation plugin install".
 ```
 
 ### Upgrade a plugin from URL
@@ -194,7 +197,8 @@ notation plugin uninstall <plugin_name> --yes
 If the plugin is not found, an error is returned showing the syntax for the plugin list command to show the installed plugins.
 
 ```shell
-Error: unable to find plugin <plugin_name>. To view a list of installed plugins, use "notation plugin list".
+Error: unable to find plugin <plugin_name>. 
+To view a list of installed plugins, use "notation plugin list".
 ```
 
 ### List installed plugins
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -114,7 +114,7 @@ To view a list of installed plugins, use "notation plugin list"
 If the plugin version is lower than the existing plugin, Notation will return an error message and will not start installation. Users can use a flag `--force` to skip plugin version check and force the installation.
 
 ```console
-Error: failed to install the plugin: <plugin-name>. The installing version <a.b.c> is lower than the existing plugin with version <x.y.z>.
+Error: failed to install the plugin: <plugin-name>. The installing version <a.b.c> is lower than the existing plugin version <x.y.z>.
 It is not recommended to install an older version. Use "--force" to force the installation if you want to do so. To view a list of installed plugins, use "notation plugin list"
 ```
 ### Install a plugin from URL
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -19,7 +19,6 @@ Available Commands:
   list        List installed plugins
   uninstall   Uninstall a plugin
   upgrade     Upgrade a plugin
-  
 
 Flags:
   -h, --help          help for plugin
@@ -51,8 +50,8 @@ Usage:
 Flags:
   -h, --help                    help for install
   -f, --force                   force the installation of a plugin
-      --checksum string         must match SHA256 of the plugin source
-      --source string           the plugin installation source, options: "file", "url","registry" (default "file")                  
+      --checksum string         must match SHA256 of the plugin source               
+
 Aliases:
   install, add
 ```
@@ -67,8 +66,7 @@ Usage:
 
 Flags:
   -h, --help                    help for upgrade
-      --checksum string         must match SHA256 of the plugin source
-      --source string           the plugin installation source, options: "file", "url","registry" (default "file")                  
+      --checksum string         must match SHA256 of the plugin source                 
 ```
 
 ### notation plugin uninstall
@@ -83,7 +81,7 @@ Flags:
   -h, --help          help for remove
   -y, --yes           do not prompt for confirmation
 Aliases:
-  remove, rm, uninstall
+  uninstall, remove, rm
 ```
 
 ## Usage
@@ -92,16 +90,10 @@ Aliases:
 
 ### Install a plugin from file system
 
-Install a Notation plugin from file system and verify the plugin checksum.
+Install a Notation plugin from file system. 
 
 ```shell
-$ notation plugin install --source file --checksum <digest> <file_path>
-```
-
-If the user doesn't specify the `--source` flag, Notation will install the plugin from file system by default.
-
-```shell
-$ notation plugin install --checksum <digest> <file_path>
+$ notation plugin install <file_path>
 ```
 
 Upon successful execution, the plugin is copied to Notation's plugin directory. The name and version of the installed plugin is displayed as follows. 
@@ -124,10 +116,10 @@ Error: failed to install the plugin, input checksum does not match the published
 
 ### Install a plugin from URL
 
-Install a Notation plugin from a remote shared address and verify the plugin checksum.
+Install a Notation plugin from a remote shared address and verify the plugin checksum. Notation only supports installing plugins from an HTTPS URL.
 
 ```shell
-$ notation plugin install --source url <URL> --checksum <digest>
+$ notation plugin install --checksum <digest>
 ```
 
 ### Install a plugin as an OCI artifact from a registry (for future iteration)
@@ -135,7 +127,7 @@ $ notation plugin install --source url <URL> --checksum <digest>
 Install a Notation plugin from a registry. Users can verify the plugin's signature with `notation verify` before the plugin installation.
 
 ```shell
-$ notation plugin install --source registry <artifact_reference>
+$ notation plugin install <registry>/<repository>@<digest>
 ```
 
 ### Upgrade a plugin to a higher version from file system 
@@ -143,7 +135,7 @@ $ notation plugin install --source registry <artifact_reference>
 Upgrade a Notation plugin to a higher version from file system and verify the plugin checksum.
 
 ```shell
-$ notation plugin upgrade --source file <file_path> --checksum <digest>
+$ notation plugin upgrade <file_path> --checksum <digest>
 ```
 
 Upon successful execution, the plugin is copied to Notation's plugin directory. The name and version of the installed plugin is displayed as follows. 
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -44,14 +44,14 @@ Aliases:
 Install a plugin
 
 Usage:
-  notation plugin install [flags] <plugin_source>
+  notation plugin install [flags] <--file|--url> <plugin_source>
 
 Flags:
-      --file                    install a plugin from a local file
+      --file                    install plugin from a file in file system
   -h, --help                    help for install
   -f, --force                   force the installation of a plugin
       --sha256sum string        must match SHA256 of the plugin source
-      --url                     install a plugin from a remote HTTPS address
+      --url                     install plugin from an HTTPS URL
 
 Aliases:
   install, add
@@ -93,7 +93,7 @@ Successfully installed plugin <plugin name>, version <x.y.z>
 If the entered plugin checksum digest doesn't match the published checksum, Notation will return an error message and will not start installation.
 
 ```console
-Error: failed to install the plugin, input checksum does not match the published checksum, expected <digest>
+Error: failed to install the plugin: plugin checksum does not match user input. Expecting <sha256sum>
 ```
 
 If the plugin version is higher than the existing plugin, Notation will start installation and overwrite the existing plugin.
@@ -105,17 +105,15 @@ Successfully installed plugin <plugin name>, updated the version from <x.y.z> to
 If the plugin version is equal to the existing plugin, Notation will not start installation and return the following message. Users can use a flag `--force` to skip plugin version check and force the installation.
 
 ```console
-Plugin <plugin name>-<x.y.z> already exists. 
-To view a list of installed plugins, use "notation plugin list".
-If you want to install the same version, use "notation plugin uninstall" to uninstall it before re-installing the same version or use "--force" to force the installation
+Plugin <plugin name> with version <x.y.z> already exists. 
+To view a list of installed plugins, use "notation plugin list"
 ```
 
 If the plugin version is lower than the existing plugin, Notation will return an error message and will not start installation. Users can use a flag `--force` to skip plugin version check and force the installation.
 
 ```console
-Error: failed to install the plugin, plugin version is lower than the existing plugin. 
-To view a list of installed plugins, use "notation plugin list".
-If you want to install an old version, use "notation plugin uninstall" to uninstall the existing plugin before installation or use "--force" to force the installation
+Error: failed to install the plugin: <plugin-name>. The installing version <a.b.c> is lower than the existing plugin. 
+To view a list of installed plugins, use "notation plugin list". It is not recommended to install older versions. Use "--force" to force the installation if you want to install an old version in a certain scenario.
 ```
 ### Install a plugin from URL
 
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -55,6 +55,20 @@ Aliases:
   install, add
 ```
 
+### notation plugin upgrade
+
+```text
+Upgrade a plugin
+
+Usage:
+  notation plugin upgrade [flags] <plugin_source>
+
+Flags:
+  -h, --help                    help for install
+      --checksum string         must match SHA256 of the plugin source
+      --source string           the location of plugin installation file, options: "file", "url","registry" (default "file")                  
+```
+
 ### notation plugin uninstall
 
 ```text
@@ -97,7 +111,13 @@ Successfully installed plugin <plugin name>, version <x.y.z>
 If the plugin directory does not exist, it will be created. When an existing plugin is detected and the version is the same as the installing plugin, it fails to install and returns the error as follows. Users can use a flag `--force` to skip existence check and force the installation with a specified version.
 
 ```console
-Error: failed to install the plugin, <plugin name> already installed
+Error: failed to install the plugin, <plugin_name> already installed
+```
+
+If the entered plugin checksum digest doesn't match the published checksum, Notation will return an error message and will not start installation.
+
+```console
+Error: failed to install the plugin, input checksum does not match the published checksum, expected <digest>
 ```
 
 ### Install a plugin from URL
@@ -114,6 +134,26 @@ Install a Notation plugin from a registry. Users can verify the plugin's signatu
 
 ```shell
 $ notation plugin install --source registry <artifact_reference>
+```
+
+### Upgrade a plugin to a higher version from file system 
+
+Upgrade a Notation plugin to a higher version from file system and verify the plugin checksum.
+
+```shell
+$ notation plugin upgrade --source file <file_path> --checksum <digest>
+```
+
+Upon successful execution, the plugin is copied to Notation's plugin directory. The name and version of the installed plugin is displayed as follows. 
+
+```console
+Successfully upgraded plugin <plugin name> to version <x.y.z>
+```
+
+If the upgrade version is equal to or lower than an existing plugin, Notation will return an error message and will not start upgrade.
+
+```console
+Error: failed to upgrade the plugin, <plugin name> version should be higher than <x.y.z>
 ```
 
 ### Uninstall a plugin
@@ -153,6 +193,6 @@ An example of output from `notation plugin list`:
 
 ```text
 NAME                                   DESCRIPTION                                   VERSION       CAPABILITIES                                                                                            ERROR
-azure-kv                               Sign artifacts with keys in Azure Key Vault   v0.5.0-rc.1   [SIGNATURE_GENERATOR.RAW]                                                                                <nil>
-com.amazonaws.signer.notation.plugin   AWS Signer plugin for Notation                1.0.290       [SIGNATURE_GENERATOR.ENVELOPE SIGNATURE_VERIFIER.TRUSTED_IDENTITY SIGNATURE_VERIFIER.REVOCATION_CHECK]   <nil>
+azure-kv                               Sign artifacts with keys in Azure Key Vault   v1.0.0   Signature generation                                                                                <nil>
+com.amazonaws.signer.notation.plugin   AWS Signer plugin for Notation                1.0.290       Signature envelope generation, Trusted Identity validation, Certificate chain revocation check   <nil>
 ```
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -74,7 +74,7 @@ Aliases:
 
 ## Install a plugin 
 
-### Install a plugin from a file system
+### Install a plugin from file system
 
 ```shell
 $ notation plugin install --file <file_path> --checksum <digest>
@@ -107,26 +107,26 @@ $ notation plugin install --source registry <artifact_reference>
 ### Uninstall a plugin
 
 ```shell
-notation plugin uninstall <plugin name>
+notation plugin uninstall <plugin_name>
 ```
 
 Upon successful execution, the plugin is uninstalled from the plugin directory. 
 
 ```shell
 Are you sure you want to uninstall plugin "<plugin name>"? [y/N] y
-Successfully uninstalled <plugin name> 
+Successfully uninstalled <plugin_name> 
 ```
 
 Uninstall the plugin without prompt for confirmation.
 
 ```shell
-notation plugin uninstall <plugin name> --yes
+notation plugin uninstall <plugin_name> --yes
 ```
 
 If the plugin is not found, an error is returned showing the syntax for the plugin list command to show the installed plugins.
 
 ```shell
-Error: <plugin name> does not exist
+Error: <plugin_name> does not exist
 ```
 
 ### List installed plugins
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -107,7 +107,7 @@ Successfully installed plugin <plugin name>, updated the version from <x.y.z> to
 If the plugin version is equal to the existing plugin, Notation will not start installation and return the following message. Users can use a flag `--force` to skip plugin version check and force the installation.
 
 ```console
-Error: Plugin <plugin name> with version <x.y.z> already exists.
+Error: failed to install the plugin: <plugin-name> with version <x.y.z> already exists.
 ```
 
 If the plugin version is lower than the existing plugin, Notation will return an error message and will not start installation. Users can use a flag `--force` to skip plugin version check and force the installation.
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -16,7 +16,8 @@ Usage:
 
 Available Commands:
   list        List installed plugins
-  install     Installs a plugin
+  install     Install a plugin
+  upgrade     Upgrade a plugin
   uninstall   Uninstall a plugin
 
 Flags:
--- a/specs/commandline/plugin.md
+++ b/specs/commandline/plugin.md
@@ -2,7 +2,7 @@
 
 ## Description
 
-Use `notation plugin` to manage plugins. See notation [plugin documentation](https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md) for more details. The `notation plugin` command by itself performs no action. In order to operate on a plugin, one of the subcommands must be used.
+Use `notation plugin` to manage plugins. See notation [plugin documentation](https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md) for more details. The `notation plugin` command by itself performs no action. In order to manage notation plugins, one of the subcommands must be used.
 
 ## Outline
 
@@ -17,7 +17,8 @@ Usage:
 Available Commands:
   list        List installed plugins
   install     Installs a plugin
-  remove      Removes a plugin
+  uninstall   Uninstall a plugin
+  upgrade     Upgrade a plugin
 
 Flags:
   -h, --help          help for plugin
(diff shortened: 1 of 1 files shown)


specs/commandline/plugin.md
```diff
@@ -41,51 +42,87 @@ Aliases:
 ### notation plugin install
 
 ```text
-Installs a plugin
+Install a plugin
 
 Usage:
-  notation plugin install [flags] <plugin path>
+  notation plugin install [flags] <plugin location>
```
Two-Hearts, line 48: how about:
```suggestion
  notation plugin install [flags] <plugin source>
```

specs/commandline/plugin.md
```diff
@@ -41,51 +42,87 @@ Aliases:
...
 
 Flags:
   -h, --help                    help for install
   -f, --force                   force the installation of a plugin
-
+      --checksum                verify the checksum digest of a plugin
```
Two-Hearts, line 53: To be more precise:
```suggestion
      --checksum string               must match SHA256 of the plugin source
```

specs/commandline/plugin.md
```diff
@@ -41,51 +42,87 @@ Aliases:
...
-  remove, rm, uninstall, delete
+  remove, rm, uninstall
 ```
 
 ## Usage
 
-### Install a plugin
+## Install a plugin 
+
+### Install a plugin from a local file
```
Two-Hearts, line 78:
```suggestion
### Install a plugin from file system
```

specs/commandline/plugin.md
```diff
@@ -41,51 +42,87 @@ Aliases:
...
 Usage:
-  notation plugin install [flags] <plugin path>
+  notation plugin install [flags] <plugin location>
 
 Flags:
   -h, --help                    help for install
   -f, --force                   force the installation of a plugin
-
+      --checksum                verify the checksum digest of a plugin
+      --source string           the location of plugin installation file, options: "file", "url","registry" (default "file")
```
Two-Hearts, line 54: To align with other designs, can we use separate flags? Something like:
```suggestion
      --file bool           if set, install the plugin from file system
      --url bool            if set, install the plugin from URL
```
We could simply make these flags mutual exclusive. (Based on PRD, we will support two flags in the first iteration.)

specs/commandline/plugin.md
```diff
@@ -41,51 +42,87 @@ Aliases:
...
 
 ## Usage
 
-### Install a plugin
+## Install a plugin 
+
+### Install a plugin from a local file
+
+```shell
+$ notation plugin install <path> --checksum <digest>
```
Two-Hearts, line 81:
```suggestion
$ notation plugin install --file --checksum <digest> <file_path>
```

specs/commandline/plugin.md
```diff
@@ -41,51 +42,87 @@ Aliases:
...
+```
+
+Upon successful execution, the plugin is copied to plugin directory. The name and version of the installed plugin is displayed as follows.
```
Two-Hearts, line 84:
```suggestion
Upon successful execution, the plugin is copied to Notation's plugin directory. The name and version of the installed plugin is displayed as follows. 
```

specs/commandline/plugin.md
```diff
@@ -41,51 +42,87 @@ Aliases:
...
+$ notation plugin install <path> --checksum <digest>
+```
+
+Upon successful execution, the plugin is copied to plugin directory. The name and version of the installed plugin is displayed as follows. 
+
+```console
+Successfully installed plugin <plugin name>, version <x.y.z>
+```
+
+If the plugin directory does not exist, it will be created. When an existing plugin is detected and the version is the same as the installing plugin, it fails to install and returns the error as follows. Users can use a flag `--force` to skip version check and force the installation a specified version.
```
Two-Hearts, line 90: Since we have the `notation plugin upgrade` command, we are not checking versions in the `notation plugin install` command anymore. Instead, if there is an existing plugin, the install would fail.
```suggestion
If the plugin directory does not exist, it will be created. When an existing plugin is detected, it fails to install and returns the error as follows. Users can use a flag `--force` to skip existence check and force the installation with a specified version.
```

specs/commandline/plugin.md
```diff
@@ -41,51 +42,87 @@ Aliases:
...
+
+```console
+Error: failed to install the plugin, <plugin name> already installed
+```
+
+### Install a plugin from URL
 
 ```shell
-notation plugin install <plugin path>
+$ notation plugin install --source url <path> --checksum <digest>
```
Two-Hearts, line 99:
```suggestion
$ notation plugin install --url --checksum <digest> <URL>
```

specs/commandline/plugin.md
```diff
@@ -41,51 +42,87 @@ Aliases:
...
 ```
 
-Upon successful execution, the plugin is copied to plugins directory and name+version of plugin is displayed. If the plugin directory does not exist, it will be created. When an existing plugin is detected, the versions are compared and if the existing plugin is a lower version then it is replaced by the newer version.
+### Install a plugin as an OCI artifact from a registry
+
+```shell
+$ notation plugin install --source registry <artifact reference>
+```
```
Two-Hearts, line 106: Based on PRD, this one is planned in `v1.2.0` right?
```suggestion
```

specs/commandline/plugin.md
```diff
@@ -41,51 +42,87 @@ Aliases:
...
 Usage:
-  notation plugin remove [flags] <plugin name>
+  notation plugin uninstall [flags] <plugin name>
 
 Flags:
   -h, --help          help for remove
 
 Aliases:
-  remove, rm, uninstall, delete
+  remove, rm, uninstall
```
Two-Hearts, line 71: The `notation plugin uninstall` command needs to be script friendly as well. Thus, we still need a `--yes` flag to `not prompt for confirmation`.

specs/commandline/plugin.md
```diff
@@ -41,51 +42,87 @@ Aliases:
...
+
+```shell
+$ notation plugin install --source registry <artifact reference>
+```
 
 ### Uninstall a plugin
 
 ```shell
-notation plugin remove <plugin name>
+notation plugin uninstall <plugin name>
```
Two-Hearts, line 111: Need to add an example of:
`notation plugin uninstall <plugin name> --yes`

specs/commandline/plugin.md
FeynmanZhou: Yes. I marked it as "for future iteration"

specs/commandline/plugin.md
FeynmanZhou: Having a `--source` flag with different options"file", "url","registry" (default "file") seems simpler than providing multiple flags here.

specs/commandline/plugin.md
```diff
@@ -41,51 +42,145 @@ Aliases:
...
-### Install a plugin
+## Install a plugin 
+
+### Install a plugin from file system
+
+Install a Notation plugin from file system and verify the plugin checksum.
 
 ```shell
-notation plugin install <plugin path>
+$ notation plugin install --source file <file_path> --checksum <digest>
```
Two-Hearts, line 97: Although it's the same, this might look better:
```suggestion
$ notation plugin install --source file --checksum <digest> <file_path>
```
Same applies to other examples.

specs/commandline/plugin.md
```diff
@@ -41,51 +42,145 @@ Aliases:
...
+$ notation plugin install <file_path> --checksum <digest>
+```
+
+Upon successful execution, the plugin is copied to Notation's plugin directory. The name and version of the installed plugin is displayed as follows. 
+
+```console
+Successfully installed plugin <plugin name>, version <x.y.z>
+```
+
+If the plugin directory does not exist, it will be created. When an existing plugin is detected and the version is the same as the installing plugin, it fails to install and returns the error as follows. Users can use a flag `--force` to skip existence check and force the installation with a specified version.
```
Two-Hearts, line 112: Only existence is checked, version is not checked here.
```suggestion
If the plugin directory does not exist, it will be created. When an existing plugin is detected, it fails to install and returns the error as follows. Users can use a flag `--force` to skip existence check and force the installation.
```

specs/commandline/plugin.md
```diff
@@ -41,51 +42,145 @@ Aliases:
...
+```shell
+$ notation plugin install --source url <URL> --checksum <digest>
+```
+
+### Install a plugin as an OCI artifact from a registry (for future iteration)
+
+Install a Notation plugin from a registry. Users can verify the plugin's signature with `notation verify` before the plugin installation.
+
+```shell
+$ notation plugin install --source registry <artifact_reference>
```
Two-Hearts, line 137: Checksum check is always required:
```suggestion
$ notation plugin install --source registry --checksum <digest> <artifact_reference>
```

specs/commandline/plugin.md
```diff
@@ -41,51 +42,145 @@ Aliases:
...
+```
+
+### Upgrade a plugin to a higher version from file system
```
Two-Hearts, line 140: Need an example of using `notation plugin upgrade`, but the plugin does NOT exist. In this case, we should return an error.

specs/commandline/plugin.md
```diff
@@ -41,51 +42,145 @@ Aliases:
...
 Usage:
-  notation plugin install [flags] <plugin path>
+  notation plugin install [flags] <plugin_source>
 
 Flags:
   -h, --help                    help for install
   -f, --force                   force the installation of a plugin
-
+      --checksum string         must match SHA256 of the plugin source
+      --source string           the location of plugin installation file, options: "file", "url","registry" (default "file")
```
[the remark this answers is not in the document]
yizha1: It is possible that the type of source can be detected automatically, so that we don't need specify the option "file", "url", "registry". It is not a common practice.
- filepath
- http URL
- repo URL

A [guideline ](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md#argument-naming-conventions) for your reference

yizha1 (commented):
Thanks @FeynmanZhou . I did a first round review, will take another pass later.

specs/commandline/plugin.md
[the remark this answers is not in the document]
FeynmanZhou: I was hoping to omit the plugin source options either and make the source type detect by Notation automatically.  @Two-Hearts mentioned there is a special case that if a file path is as same as a repo URL, it will be hard for Notation to distinguish that case.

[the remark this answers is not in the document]
Two-Hearts: > @yizha1 I was hoping to omit the plugin source options either and make the source type detect by Notation automatically. @Two-Hearts mentioned there is a special case that if a file path is as same as a repo URL, it will be hard for Notation to distinguish that case.

To correct, a file path can be as same as a valid oci artifact reference in a registry. If we are going to support oci registry in the future, then we should consider this situation in the initial design. @yizha1 @FeynmanZhou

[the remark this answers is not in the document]
yizha1: @FeynmanZhou @Two-Hearts Maybe we can refer to `helm install` command, which supports installation from a file/url/repo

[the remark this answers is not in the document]
Two-Hearts: @yizha1 That's a good idea, from helm, I could see:
`By OCI registries: helm install mynginx --version 1.2.3 oci://example.com/charts/nginx`
They ask user to add a prefix `oci://` to the input and they explicitly check for this prefix, if it exists, then it's by oci registry. @FeynmanZhou What do you think?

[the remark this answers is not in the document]
Two-Hearts: Another option might be:
1. We check the user input in the file system first. If it's a valid, existed, file path. Then we take the input as a file path and try install from there.
2. If it does not exist in the file system, then we check if it's a valid URL. If yes, install from the URL.
3. For the current iteration, if not file path nor URL, we will return an error. For future iteration, if we support oci registry, then we treat the input as an oci reference. If it's a valid reference, then we install from oci registry.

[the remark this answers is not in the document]
yizha1: Helm also supports installation from a repo, how does helm distinguish it is a file or a repo? For example, 

```
$ helm repo ls
NAME                    URL
harbor                  https://helm.goharbor.io

$ helm install myharbor harbor/harbor

$ ls
<no directory harbor/harbor>
```

specs/commandline/plugin.md
FeynmanZhou: Why an OCI image requires checksum? Maybe verify the signature and its image digest are enough to ensure the plugin's integrity and authenticity.

specs/commandline/plugin.md
```diff
@@ -41,51 +42,145 @@ Aliases:
...
 
 Usage:
-  notation plugin remove [flags] <plugin name>
+  notation plugin upgrade [flags] <plugin_source>
 
 Flags:
-  -h, --help          help for remove
+  -h, --help                    help for install
+      --checksum string         must match SHA256 of the plugin source
+      --source string           the location of plugin installation file, options: "file", "url","registry" (default "file")
```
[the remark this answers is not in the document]
FeynmanZhou: @yizha1  In general, a feature spec can have features in the roadmap or future milestones. The Notary Project spec also has some points and feature requirements that are planned for future releases. 
If we have a feature defined in the spec earlier, it is possible to be implemented anytime. I think we can mark this kind of features as "for future milestone" in spec to avoid confusion.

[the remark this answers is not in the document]
yizha1: My point is that this section will be used for implementation, however, we won't support registry, so we cannot display it to users, right?

specs/commandline/plugin.md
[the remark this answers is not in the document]
Two-Hearts: > Helm also supports installation from a repo, how does helm distinguish it is a file or a repo? For example,
> 
> ```
> $ helm repo ls
> NAME                    URL
> harbor                  https://helm.goharbor.io
> 
> $ helm install myharbor harbor/harbor
> 
> $ ls harbor/harbor
> ls: cannot access 'harbor/harbor': No such file or directory
> ```

@yizha1 That's a good question. The answer is, helm treats `harbor/harbor` as a file path first, they check if this is a valid file path. If the path does not exist, then they take it as a repo. So the process looks automatically on the external user's point of view.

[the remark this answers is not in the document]
Two-Hearts: > @Two-Hearts If the edge case can be covered by this option, it sounds good to me.

Yes, the edge case can be covered. No matter what user input it is, Notation always check by order: file path -> URL [-> OCI reference, in the future].

specs/commandline/plugin.md
```diff
@@ -15,9 +15,11 @@ Usage:
   notation plugin [command]
 
 Available Commands:
+  install     Install a plugin
   list        List installed plugins
-  install     Installs a plugin
-  remove      Removes a plugin
+  uninstall   Uninstall a plugin
+  upgrade     Upgrade a plugin
+
```
shizhMSFT, line 22: remove extra line

specs/commandline/plugin.md
```diff
@@ -41,51 +43,151 @@ Aliases:
...
-  notation plugin install [flags] <plugin path>
+  notation plugin install [flags] <plugin_source>
 
 Flags:
   -h, --help                    help for install
   -f, --force                   force the installation of a plugin
-
+      --checksum string         must match SHA256 of the plugin source
+      --source string           the plugin installation source, options: "file", "url","registry" (default "file")                  
 Aliases:
```
shizhMSFT, line 56: missing extra line

specs/commandline/plugin.md
```diff
@@ -41,51 +43,151 @@ Aliases:
...
 Usage:
-  notation plugin install [flags] <plugin path>
+  notation plugin install [flags] <plugin_source>
 
 Flags:
   -h, --help                    help for install
   -f, --force                   force the installation of a plugin
-
+      --checksum string         must match SHA256 of the plugin source
+      --source string           the plugin installation source, options: "file", "url","registry" (default "file")
```
shizhMSFT, line 55: should we default to `registry`? or even `auto`?

specs/commandline/plugin.md
```diff
@@ -41,51 +43,151 @@ Aliases:
...
+
+Usage:
+  notation plugin uninstall [flags] <plugin_name>
 
+Flags:
+  -h, --help          help for remove
+  -y, --yes           do not prompt for confirmation
 Aliases:
-  remove, rm, uninstall, delete
+  remove, rm, uninstall
```
shizhMSFT, line 86: nit:
```suggestion
  uninstall, remove, rm
```

specs/commandline/plugin.md
```diff
@@ -41,51 +43,151 @@ Aliases:
...
 
-### Install a plugin
+## Install a plugin 
+
+### Install a plugin from file system
+
+Install a Notation plugin from file system and verify the plugin checksum.
+
+```shell
+$ notation plugin install --source file --checksum <digest> <file_path>
```
shizhMSFT, line 98: When install from a local file, the checksum should be optional. Otherwise, people will just do
```bash
notation plugin install --source file --checksum $(sha256sum $file | cut -f1 -d' ') $file
```
which is not really useful.

specs/commandline/plugin.md
```diff
@@ -41,51 +43,151 @@ Aliases:
...
 
 Usage:
-  notation plugin install [flags] <plugin path>
+  notation plugin install [flags] <plugin_source>
 
 Flags:
   -h, --help                    help for install
   -f, --force                   force the installation of a plugin
-
+      --checksum string         must match SHA256 of the plugin source
```
shizhMSFT, line 54: What's the format of `checksum`?

specs/commandline/plugin.md
```diff
@@ -41,51 +43,151 @@ Aliases:
...
+Error: failed to install the plugin, input checksum does not match the published checksum, expected <digest>
+```
+
+### Install a plugin from URL
+
+Install a Notation plugin from a remote shared address and verify the plugin checksum.
 
 ```shell
-notation plugin install <plugin path>
+$ notation plugin install --source url <URL> --checksum <digest>
```
shizhMSFT, line 130: What kind of URL does it support?

specs/commandline/plugin.md
shizhMSFT: That's why an example is required.

The `artifact_reference` contains the digest.

specs/commandline/plugin.md
```diff
@@ -41,51 +43,151 @@ Aliases:
...
+Uninstall the plugin without prompt for confirmation.
+
+```shell
+notation plugin uninstall <plugin_name> --yes
+```
+
+If the plugin is not found, an error is returned showing the syntax for the plugin list command to show the installed plugins.
+
+```shell
+Error: <plugin_name> does not exist. Please install the plugin using 'notation plugin install' first.
```
shizhMSFT, line 189: Should we suggest `notation plugin ls` instead?

specs/commandline/plugin.md
FeynmanZhou: It's the encoded value without `sha256:`.

FeynmanZhou:
@rgnote @priteshbandi @iamsamirzon  Would you mind reviewing the notation plugin mgmt spec updates? Thanks

specs/commandline/plugin.md
```diff
@@ -41,51 +42,144 @@ Aliases:
...
+Successfully installed plugin <plugin name>, version <x.y.z>
+```
+
+If the plugin directory does not exist, it will be created. When an existing plugin is detected, it fails to install and returns the error as follows. Users can use a flag `--force` to skip existence check and force the installation.
+
+```console
+Error: failed to install the plugin, <plugin_name> already installed
+```
+
+If the entered plugin checksum digest doesn't match the published checksum, Notation will return an error message and will not start installation.
```
Two-Hearts, line 111: One thing might be worth to point out in the spec:
`when user is installing a plugin from file system, the checksum is optional.`

specs/commandline/plugin.md
```diff
@@ -41,51 +42,144 @@ Aliases:
...
+Error: failed to install the plugin, input checksum does not match the published checksum, expected <digest>
+```
+
+### Install a plugin from URL
+
+Install a Notation plugin from a remote shared address and verify the plugin checksum. Notation only supports installing plugins from an HTTPS URL.
 
 ```shell
-notation plugin install <plugin path>
+$ notation plugin install --checksum <digest>
```
Two-Hearts, line 122: This should be:
```suggestion
$ notation plugin install --checksum <digest> <URL>
```

specs/commandline/plugin.md
```diff
@@ -41,51 +42,144 @@ Aliases:
...
+```shell
+$ notation plugin install <registry>/<repository>@<digest>
+```
+
+### Upgrade a plugin to a higher version from file system 
+
+Upgrade a Notation plugin to a higher version from file system and verify the plugin checksum.
+
+```shell
+$ notation plugin upgrade <file_path> --checksum <digest>
```
Two-Hearts, line 138: When upgrading from a file system, `--checksum` should be optional? And should we also provide examples when upgrading from URL and OCI registry?

specs/commandline/plugin.md
FeynmanZhou: Agree. Added this note

specs/commandline/plugin.md
FeynmanZhou: Good catch. Fixed

specs/commandline/plugin.md
FeynmanZhou: Yes. Removed checksum validation here. Added two examples for upgrading plugin from URL and OCI registry

Two-Hearts (dismissed):
LGTM

specs/commandline/plugin.md
```diff
@@ -99,6 +209,6 @@ An example of output from `notation plugin list`:
 
 ```text
 NAME                                   DESCRIPTION                                   VERSION       CAPABILITIES                                                                                            ERROR
-azure-kv                               Sign artifacts with keys in Azure Key Vault   v0.5.0-rc.1   [SIGNATURE_GENERATOR.RAW]                                                                                <nil>
-com.amazonaws.signer.notation.plugin   AWS Signer plugin for Notation                1.0.290       [SIGNATURE_GENERATOR.ENVELOPE SIGNATURE_VERIFIER.TRUSTED_IDENTITY SIGNATURE_VERIFIER.REVOCATION_CHECK]   <nil>
+azure-kv                               Sign artifacts with keys in Azure Key Vault   v1.0.0        Signature generation                                                                                <nil>
+com.amazonaws.signer.notation.plugin   AWS Signer plugin for Notation                1.0.290       Signature envelope generation, Trusted Identity validation, Certificate chain revocation check   <nil>
```
priteshbandi, line 213: Should we keep capabilities as enum instead of text as they are more for plugin publisher as compared to consumer?

specs/commandline/plugin.md
```diff
@@ -41,51 +42,160 @@ Aliases:
...
+
+```shell
+notation plugin uninstall <plugin_name> --yes
 ```
 
-Upon successful execution, the plugin is removed from the plugins directory. If the plugin is not found, an error is returned showing the syntax for the plugin list command to show the installed plugins.
+If the plugin is not found, an error is returned showing the syntax for the plugin list command to show the installed plugins.
+
+```shell
+Error: <plugin_name> does not exist. Please check the plugin information using 'notation plugin list' first.
```
priteshbandi, line 197:
```suggestion
Error: unable to find plugin plugin_name. To view a list of installed plugins, use `notation plugin list`.
```

specs/commandline/plugin.md
```diff
@@ -41,51 +42,160 @@ Aliases:
...
-### Install a plugin
+## Install a plugin 
+
+### Install a plugin from file system
+
+Install a Notation plugin from file system. The checksum validation is optional for this case.
 
 ```shell
-notation plugin install <plugin path>
+$ notation plugin install <file_path>
```
priteshbandi, line 96: Can we please specify the file supported file formats and install behavior for them ?
For Example:

1. `.zip`: The ZIP file should house the plugin in the root directory. If the ZIP archive contains multiple files, Notation will locate an executable file associated with the plugin and copy it, ignoring any other files?
2. `.tar.gz`: Similar for .tar.gz
3. `.tar`: Should we even support?
4. `.gz`: Should we even support?
5. All other format, Notation will treat them as executable?

^ Notation will use file extension to determine how to unpack file/archive ?

specs/commandline/plugin.md
```diff
@@ -41,51 +42,160 @@ Aliases:
...
+```console
+Error: failed to install the plugin, input checksum does not match the published checksum, expected <digest>
+```
+
+### Install a plugin from URL
+
+Install a Notation plugin from a remote shared address and verify the plugin checksum. Notation only supports installing plugins from an HTTPS URL.
+
+```shell
+$ notation plugin install --checksum <digest> <URL>
```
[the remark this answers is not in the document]
Two-Hearts: @priteshbandi I could answer this one, here's the order (similar to [helm](https://github.com/helm/helm/blob/d70e293e5fea032e843c29d15c7e85e78b0aca6b/pkg/action/install.go#L718)):
1. We check the user input in the file system first. If it's a valid, existed, file path. Then we take the input as a file path and try install from there. (Notation would validate if it's indeed a valid Notation plugin before installation.)
2. If it does not exist in the file system, then we check if it's a valid URL (with prefix `https://`). If yes, try install from the URL.

specs/commandline/plugin.md
```diff
@@ -41,51 +42,160 @@ Aliases:
...
+```
+
+### Install a plugin as an OCI artifact from a registry (for future iteration)
+
+Install a Notation plugin from a registry. Users can verify the plugin's signature with `notation verify` before the plugin installation.
+
+```shell
+$ notation plugin install <registry>/<repository>@<digest>
```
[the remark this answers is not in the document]
Two-Hearts: @FeynmanZhou Since OCI related logic is planned in future iteration, can we remove it from this PR for now?

specs/commandline/plugin.md
Two-Hearts: @priteshbandi These questions are related to implementation though. Since you asked, here are the current design:
1. For `.zip`, Notation goes through the zip file -> find an executable file -> validate its name (must be `notatin-{plugin-name}`) -> check if plugin already exists (if --force is not set) -> extract the executable file into a `tmp` dir (regardless the final result is a success or a failure, this `tmp` dir will be removed in the end) -> validate the executable file is indeed a valid Notation plugin by calling its `get-plugin-metada` -> at this point, copy the file into correct Notation dir with all parent dir created along the way. A special note for `.zip` file, a normal file such as `LICENSE` can be treated as executable too. That's why we cannot directly use whatever executable file Notation finds. 
2. For `.tar.gz`, Notation will first detect that a `.gz` file is passed in. Then, Notation will require this `.gz` file compressed a `.tar` file in it. If yes, then apply similar logic as `.zip`.
3. `.tar` alone is not supported.
4. `.gz` alone is not supported.
5. All other format is not very different from `.tar` or `.gz`, they are not supported.

And yes, Notation has the capability of telling apart `.zip` and `.tar.gz`, but not using file extension. Instead, it considers at most the first 512 bytes of data and determine the file type.

specs/commandline/plugin.md
```diff
@@ -41,51 +42,160 @@ Aliases:
...
 Aliases:
   install, add
 ```
 
-### notation plugin remove
+### notation plugin upgrade
 
 ```text
-Removes a plugin
+Upgrade a plugin
```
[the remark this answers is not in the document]
FeynmanZhou: I also expect allow users to upgrade without specifying the URL. It seems users still need to manually get the sha256sum from the plugin download page

specs/commandline/plugin.md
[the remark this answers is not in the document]
FeynmanZhou: I am okay with removing it from this PR. In order to reserve the design doc, I think we can add the original PRD doc to another place (e.g. notaryproject/specification) to archive our design and discussion centrally.

specs/commandline/plugin.md
FeynmanZhou: The original enum could be used and kept on the [plugin-extensibility.md#plugin-metadata)](https://github.com/notaryproject/specifications/blob/main/specs/plugin-extensibility.md#plugin-metadata). For the output of `notation plugin list`, I think it should be user-friendly so using text would be much straightforward.

specs/commandline/plugin.md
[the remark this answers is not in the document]
yizha1: @FeynmanZhou we can create an issue for the future backlog. The future backlog will be visit during planning

specs/commandline/plugin.md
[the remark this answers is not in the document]
FeynmanZhou: > How does the proposed behavior differ from using the install command?

I expect upgrade a plugin without specifying the plugin URL. This differs from the `notation plugin install` experience. I will update the CLI spec to describe this experience. 

> Could the install command be used to handle the following scenarios:

For 1, it will be handled by current design.
For 2 and 3,  the logic is a little different. `notation plugin install` only checks the plugin existence instead of version.  
the current design and behavior are described at [this line](https://github.com/notaryproject/notation/pull/809/files#diff-7637fbf27d1c8c25ef1a9806613d4aa17aaa61d99aae29f8c802b5e50bcd2efdR105). In short, when an existing plugin is detected, it fails to install and returns the error. Users can use a flag `--force` to skip existence check and force the installation.

Two-Hearts:
@FeynmanZhou Can you create an issue in https://github.com/notaryproject/specifications to keep track that we are also going to update spec there? (spec related to what and how should a vendor publish a plugin).  And link this PR to the issue. Thanks.

specs/commandline/plugin.md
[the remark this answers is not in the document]
FeynmanZhou: @priteshbandi I have adopted your suggestions above and update the `notation plugin install` usage and examples accordingly. Thanks

specs/commandline/plugin.md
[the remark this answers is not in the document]
FeynmanZhou: Removed "Install a Notation plugin from a registry"

FeynmanZhou:
> @FeynmanZhou Can you create an issue in https://github.com/notaryproject/specifications to keep track that we are also going to update spec there? (spec related to what and how should a vendor publish a plugin). And link this PR to the issue. Thanks.

Thanks @patrickzheng200 . Issue created: https://github.com/notaryproject/specifications/issues/284

FeynmanZhou:
Hi @notaryproject/notaryproject-notation-maintainers 

I have major updates in a recent commit based on the community meeting discussion on Nov 13:
- Removed `notation plugin upgrade` from this iteration. This will be considered in the next iteration
- Added two flags `--url` and `file` to `notation plugin install`, added more usage examples and error messages accordingly
- Removed "Install a Notation plugin from a registry (for future iteration)". This will be considered in the next iteration

specs/commandline/plugin.md
```diff
@@ -41,51 +41,115 @@ Aliases:
 ### notation plugin install
 
 ```text
-Installs a plugin
+Install a plugin
 
 Usage:
-  notation plugin install [flags] <plugin path>
+  notation plugin install [flags] <plugin_source>
```
Two-Hearts, line 47: Since one and only one of [--file, --url] is required,
```suggestion
  notation plugin install [flags] <--file|--url> <plugin_source>
```

specs/commandline/plugin.md
```diff
@@ -41,51 +41,115 @@ Aliases:
...
 
 Flags:
+      --file                    install a plugin from a local file
```
Two-Hearts, line 50:
```suggestion
      --file                    if set, install plugin from a file in file system
```

specs/commandline/plugin.md
```diff
@@ -41,51 +41,115 @@ Aliases:
...
   -h, --help                    help for install
   -f, --force                   force the installation of a plugin
+      --sha256sum string        must match SHA256 of the plugin source
+      --url                     install a plugin from a remote HTTPS address
```
Two-Hearts, line 54:
```suggestion
      --url                     if set, install plugin from a HTTPS URL
```

specs/commandline/plugin.md
```diff
@@ -41,51 +41,115 @@ Aliases:
...
+Upon successful execution, the plugin is copied to Notation's plugin directory. If the plugin directory does not exist, it will be created. The name and version of the installed plugin are displayed as follows. 
+
+```console
+Successfully installed plugin <plugin name>, version <x.y.z>
+```
+
+If the entered plugin checksum digest doesn't match the published checksum, Notation will return an error message and will not start installation.
+
+```console
+Error: failed to install the plugin, input checksum does not match the published checksum, expected <digest>
```
Two-Hearts, line 96:
```suggestion
Error: failed to install the plugin: plugin checksum does not match user input. Expecting <sha256sum>
```

specs/commandline/plugin.md
```diff
@@ -41,51 +41,115 @@ Aliases:
...
+```
+
+If the plugin version is higher than the existing plugin, Notation will start installation and overwrite the existing plugin.
+
+```console
+Successfully installed plugin <plugin name>, updated the version from <x.y.z> to <a.b.c>
```
Two-Hearts, line 102: The success message can be the same as before,
```suggestion
Successfully installed plugin <plugin name>, version <a.b.c>
```

specs/commandline/plugin.md
```diff
@@ -41,51 +41,115 @@ Aliases:
...
+```
+
+If the plugin version is equal to the existing plugin, Notation will not start installation and return the following message. Users can use a flag `--force` to skip plugin version check and force the installation.
+
+```console
+Plugin <plugin name>-<x.y.z> already exists.
```
Two-Hearts, line 108:
```suggestion
Plugin <plugin name> with version <x.y.z> already installed.
```

specs/commandline/plugin.md
```diff
@@ -41,51 +41,115 @@ Aliases:
...
+```console
+Successfully installed plugin <plugin name>, updated the version from <x.y.z> to <a.b.c>
+```
+
+If the plugin version is equal to the existing plugin, Notation will not start installation and return the following message. Users can use a flag `--force` to skip plugin version check and force the installation.
+
+```console
+Plugin <plugin name>-<x.y.z> already exists. 
+To view a list of installed plugins, use "notation plugin list".
+If you want to install the same version, use "notation plugin uninstall" to uninstall it before re-installing the same version or use "--force" to force the installation
```
Two-Hearts, line 110: we don't need to explicitly say it,
```suggestion
```

specs/commandline/plugin.md
```diff
@@ -41,51 +41,115 @@ Aliases:
...
+```
+
+If the plugin version is lower than the existing plugin, Notation will return an error message and will not start installation. Users can use a flag `--force` to skip plugin version check and force the installation.
+
+```console
+Error: failed to install the plugin, plugin version is lower than the existing plugin.
```
Two-Hearts, line 116:
```suggestion
Error: failed to install the plugin: <plugin-name> current version <x.y.z> is larger than the installing version <a.b.c>
```

specs/commandline/plugin.md
```diff
@@ -41,51 +41,115 @@ Aliases:
...
+To view a list of installed plugins, use "notation plugin list".
+If you want to install the same version, use "notation plugin uninstall" to uninstall it before re-installing the same version or use "--force" to force the installation
+```
+
+If the plugin version is lower than the existing plugin, Notation will return an error message and will not start installation. Users can use a flag `--force` to skip plugin version check and force the installation.
+
+```console
+Error: failed to install the plugin, plugin version is lower than the existing plugin. 
+To view a list of installed plugins, use "notation plugin list".
+If you want to install an old version, use "notation plugin uninstall" to uninstall the existing plugin before installation or use "--force" to force the installation
```
Two-Hearts, line 118: same here,
```suggestion
```

specs/commandline/plugin.md
yizha1: @patrickzheng200 larger-->higher

specs/commandline/plugin.md
yizha1: Agree with @Two-Hearts

specs/commandline/plugin.md
yizha1: For older version, it is worth mentioning, but it's better we highlights not recommended, but in rare case, if users want to do that.

FeynmanZhou: Installing an older version is an downgrade operation. Users may consider downgrade for some reasons. I agree with @yizha1 that mentioned it but state that installing an older version is not recommended.

specs/commandline/plugin.md
FeynmanZhou: I think it's necessary to tell users about the version update if users install a higher version and overwrite an older version. @yizha1 @shizhMSFT what's your suggestion for this case?

yizha1: I agree with you @FeynmanZhou

Two-Hearts: > I think it's necessary to tell users about the version update if users install a higher version and overwrite an older version.

SGTM

specs/commandline/plugin.md
FeynmanZhou: Updated

specs/commandline/plugin.md
FeynmanZhou: Updated to `Error: failed to install the plugin: <plugin-name>. The installing version <a.b.c> is lower than the existing plugin. `

specs/commandline/plugin.md
```diff
@@ -41,51 +41,113 @@ Aliases:
...
+```console
+Plugin <plugin name> with version <x.y.z> already exists. 
+To view a list of installed plugins, use "notation plugin list"
+```
+
+If the plugin version is lower than the existing plugin, Notation will return an error message and will not start installation. Users can use a flag `--force` to skip plugin version check and force the installation.
+
+```console
+Error: failed to install the plugin: <plugin-name>. The installing version <a.b.c> is lower than the existing plugin. 
+To view a list of installed plugins, use "notation plugin list". It is not recommended to install older versions. Use "--force" to force the installation if you want to install an old version in a certain scenario.
```
Two-Hearts, line 116: This is still a bit too verbose. And logically, I think we should move `To view a list of installed plugins, use "notation plugin list".` to the end of the message. How about,
```suggestion
It is not recommended to install an older version. Use "--force" to force the installation if you want to do so. To view a list of installed plugins, use "notation plugin list".
```

specs/commandline/plugin.md
```diff
@@ -41,51 +41,113 @@ Aliases:
...
 Usage:
-  notation plugin install [flags] <plugin path>
+  notation plugin install [flags] <--file|--url> <plugin_source>
 
 Flags:
+      --file                    install plugin from a file in file system
   -h, --help                    help for install
   -f, --force                   force the installation of a plugin
+      --sha256sum string        must match SHA256 of the plugin source
+      --url                     install plugin from an HTTPS URL
```
Two-Hearts, line 54: nit:
```suggestion
  -d, --debug              debug mode
      --file               install plugin from a file in file system
      --force              force the installation of a plugin
  -h, --help               help for install
      --sha256sum string   must match SHA256 of the plugin source
      --url                install plugin from an HTTPS URL
  -v, --verbose            verbose mode
```

specs/commandline/plugin.md
```diff
@@ -41,51 +41,113 @@ Aliases:
...
+
+```console
+Plugin <plugin name> with version <x.y.z> already exists. 
+To view a list of installed plugins, use "notation plugin list"
+```
+
+If the plugin version is lower than the existing plugin, Notation will return an error message and will not start installation. Users can use a flag `--force` to skip plugin version check and force the installation.
+
+```console
+Error: failed to install the plugin: <plugin-name>. The installing version <a.b.c> is lower than the existing plugin.
```
Two-Hearts, line 115:
```suggestion
Error: failed to install the plugin: <plugin-name>. The installing version <a.b.c> is lower than the existing plugin version <x.y.z>. 
```

FeynmanZhou: Updated

Two-Hearts (dismissed):
LGTM

yizha1 (approved):
LGTM but IANAM

specs/commandline/plugin.md
```diff
@@ -41,51 +41,115 @@ Aliases:
...
+
+```console
+Successfully installed plugin <plugin name>, updated the version from <x.y.z> to <a.b.c>
+```
+
+If the plugin version is equal to the existing plugin, Notation will not start installation and return the following message. Users can use a flag `--force` to skip plugin version check and force the installation.
+
+```console
+Plugin <plugin name> with version <x.y.z> already exists. 
+To view a list of installed plugins, use "notation plugin list"
```
priteshbandi, line 111: Why do we want to show user that they can use plugin list command?
```suggestion
```console
Error: Plugin <plugin name> with version <x.y.z> already exists. 
```

specs/commandline/plugin.md
```diff
@@ -41,51 +41,115 @@ Aliases:
...
+```
+
+If the plugin version is lower than the existing plugin, Notation will return an error message and will not start installation. Users can use a flag `--force` to skip plugin version check and force the installation.
+
+```console
+Error: failed to install the plugin: <plugin-name>. The installing version <a.b.c> is lower than the existing plugin version <x.y.z>.
+It is not recommended to install an older version. Use "--force" to force the installation if you want to do so. To view a list of installed plugins, use "notation plugin list"
```
priteshbandi, line 118:
```suggestion
Unable to install the plugin: <plugin-name>. The version you are trying to install (<a.b.c>) is older than the version that is already installed (<x.y.z>). To force the installation, use the "--force" option.
```

specs/commandline/plugin.md
```diff
@@ -41,51 +41,115 @@ Aliases:
...
 ```
+### Install a plugin from URL
 
-Upon successful execution, the plugin is copied to plugins directory and name+version of plugin is displayed. If the plugin directory does not exist, it will be created. When an existing plugin is detected, the versions are compared and if the existing plugin is a lower version then it is replaced by the newer version.
+Install a Notation plugin from a remote shared address and verify the plugin checksum. Notation only supports installing plugins from an HTTPS URL.
```
priteshbandi, line 122:
```suggestion
Installs a Notation plugin from a remote location and verify the plugin checksum. Notation only supports installing plugins from an HTTPS URL, which means that the URL must start with "https://".
```

specs/commandline/plugin.md
```diff
@@ -41,51 +41,115 @@ Aliases:
...
+
+```shell
+$ notation plugin install --sha256sum <digest> --url <HTTPS_URL>
```
priteshbandi, line 125: Should we keep `--sha256sum` as optional flag ?

FeynmanZhou: Notation as a security project, I think check `--sha256sum` provides security guarantee to end users. 
Alternatively, users can download a plugin file to local file system and install it from file since `notation plugin install --file` doesn't require checksum.

specs/commandline/plugin.md
FeynmanZhou: @priteshbandi  To align the behavior with the case when installing version is equal to exist plugin as you described in https://github.com/notaryproject/notation/pull/809#discussion_r1396788077, I think this output should be stderr either.

specs/commandline/plugin.md
priteshbandi: Plugin venders may sign the artifact(i.e. zip) to generate a signature (either GPG or platform specific), making sha256 mandatory doesn't looks right for this usecase.

specs/commandline/plugin.md
```diff
@@ -41,51 +41,114 @@ Aliases:
...
+If the plugin version is higher than the existing plugin, Notation will start installation and overwrite the existing plugin.
+
+```console
+Successfully installed plugin <plugin name>, updated the version from <x.y.z> to <a.b.c>
+```
+
+If the plugin version is equal to the existing plugin, Notation will not start installation and return the following message. Users can use a flag `--force` to skip plugin version check and force the installation.
+
+```console
+Error: Plugin <plugin name> with version <x.y.z> already exists.
```
Two-Hearts, line 110:
```suggestion
Error: failed to install the plugin: <plugin-name> with version <x.y.z> already exists.
```

specs/commandline/plugin.md
```diff
@@ -41,51 +41,114 @@ Aliases:
...
+Uninstall a plugin
 
 Usage:
-  notation plugin remove [flags] <plugin name>
+  notation plugin uninstall [flags] <plugin_name>
 
 Flags:
   -h, --help          help for remove
-
+  -y, --yes           do not prompt for confirmation
```
Two-Hearts, line 72: nit:
```suggestion
  -d, --debug     debug mode
  -h, --help      help for uninstall
  -v, --verbose   verbose mode
  -y, --yes       do not prompt for confirmation
```

specs/commandline/plugin.md
priteshbandi: I concur, updated messsage
```
Error: Unable to install the plugin: <plugin-name>. The version you are trying to install (<a.b.c>) is equal or lower than the version that is already installed (<x.y.z>). To force the installation, use the "--force" option.
```

priteshbandi (approved):
LGTM

Two-Hearts (approved):
LGTM

PR #811, opened 2023-10-28
docs: spec updates for arbitrary blob signing

rgnote:
CLI Spec updated for Arbitrary blob signing. Proposal https://hackmd.io/ewbJr2ZnT4a8U1ObDVXcSw?view#CLI-Spec

Signed-off-by: rgnote <5878554+rgnote@users.noreply.github.com>


specs/commandline/blob.md
```diff
@@ -0,0 +1,362 @@
...
+### notation blob sign
+
+```text
+Produce a detached signature for a given blob
+
+Usage:
+  notation blob sign [flags] <blob_path>
+
+Flags:
+  -s,  --signature path             output location of the detached signature
```
JeyJeyGao, line 35: Is this a required flag? If it is, we might need to update the usage schema. If it's not a required flag, what's the default behavior?

Two-Hearts:
Before diving into the detailed UX design, shall we have a PR in the https://github.com/notaryproject/specifications repo to add the new specs related to blob? For example, signature payload, trust policy, ect.

specs/commandline/blob.md
```diff
@@ -0,0 +1,362 @@
...
+### Inspect the given detached blob signature
+
+```shell
+# Prerequisites: Signatures is produced by notation blob sign command
+notation blob inspect /tmp/my-blob-signature.sig
+```
+
+An example output:
+```shell
+Inspecting /tmp/my-blob-signature.sig
```
shizhMSFT, line 187: How does `notation` detect the signature format? Should `notation` print out the signature format (`jws` / `cose`)?

shizhMSFT: Same question for `verify`.

specs/commandline/policy.md
```diff
@@ -67,6 +67,110 @@ The `notation policy` command provides a user-friendly way to manage trust polic
...
 
+Policy language version 1.1 added support for verifying signatures associated with blob artifacts. User can use `scopes` field as a Policy selector string to decide which Policy gets applied to which blob. The `--policy-scope` argument provided in `notation blob verify` command will dictate which Policy gets picked from the policy configuration file and applied for verification. To get started with verifying blob signatures, users can refer to the following trust policy configuration sample. In this sample, there are three policies configured for different requirements:
+
+- The Policy named "blob-verification-policy" is for verifying blob artifacts signed by Wabbit Networks and scoped to `blob-verification-selector`.
+- Policy named "skip-blob-verification-policy" is for skipping verification on blob artifacts scoped to `skip-blob-verification-selector`.
+- Policy "wildcard-blob-verification-policy" is for auditing verification results when user wants to apply a wildcard policy by not providing `--policy-scope` argument in `notation blob verify` command.
+
+```jsonc
+{
+    "version": "1.1",
```
Two-Hearts, line 78: This is a new version of trust policy, it should be included in the https://github.com/notaryproject/specifications right?

specs/commandline/blob.md
rgnote: It is a required flag. I will updated the schema.

specs/commandline/policy.md
rgnote: It will be. I wanted to start from the UX side and work backwards. You can see the policy updates [here](https://hackmd.io/ewbJr2ZnT4a8U1ObDVXcSw?view#CLI-Spec). I will be raising a PR for policy spec updates shortly.

specs/commandline/blob.md
rgnote: Good question. I suggest Notation gets the signature format from the file extension. Something like `my-blob-signature.sig.cose`, `my-blob-signature.sig.jws`

For OCI, the format is an unsigned attribute that is present in the signature manifest (in other words, outside the signature envelope). Similarly, for Blobs, the format will come from outside the envelope i.e. file extension.

An alternative is encoding the format within the signature envelope file. Either json encoding or some sort of armoring like below.

```
{
    "mediaType": "application/cose",
    "signatureEnvelope": "<Base64 encoded envelope>"
}
``` 

or

```
-----START COSE SIGNATURE----
<Base64 encoded envelope>
-----END COSE SIGNATURE----
```

However, these two options will require Notation to parse the signature file in order to read the envelope. And also, tools other than Notation will not work out of the box as they do not understand the wrapping method,

Two-Hearts: @rgnote Quick question on the file extension method:
Say a verifier receives a signature with name `my-blob-signature.sig`, how would notation tell which signature envelope format should be used? Because a signer can always rename their signature files after signing. 
Basically, my concern here is that Notation verification is going to be dependent on the signature's file name, which is not strictly controlled by Notation.
/cc: @priteshbandi

specs/commandline/policy.md
Two-Hearts: I see. Then, could you create an issue in the https://github.com/notaryproject/specifications repo and link this PR there, as a reminder to add the spec.

specs/commandline/blob.md
JeyJeyGao: @rgnote I have a question about the UX. 

```sh
SIGNATURE_NAME=blob.sig
notation blob sign --signature blob.sig ./filename
```
Then, when the user wants to verify the signature:
```sh
notation blob verify $SIGNATURE_NAME.jws
```
How does the user determine the suffix of the signature? I believe this design elevates the complexity of using the Notation CLI. Users will either have to use `ls` to scan the directory to find the actual name or delve into our documentation to understand the naming suffix.

rgnote: @Two-Hearts 

We discussed about scenario on the call. Yes, a user can change the file name (including its extension) however they like. We can't assume that users won't change extensions. Since extension is only a hint about the format, this is not a big problem for Notation. The proposed solution is
1. If there is a valid extension, Notation will use it
2. If there is no valid extension, Notation will guess the format by bruteforcing all the available formats it knows (this is already being done in notation-core-go)
3. If none of the above methods work, Notation errors out saying unknown format.

rgnote: @JeyJeyGao 

We picked `--signature` flag when we had no idea about how to pass signature format data with the signature itself. Now that we decided to go with file extensions, we need to update the flag name. I suggest we change it to `--signature-name`.

For example, let's say user gives `xyz` for the `--signature-name` field. 
On the signing side, Notation will write the signature to `xyz.sig.cose` or `xyz.sig.jws` to the CWD. 

On the verification side, notation will expect `--signature-path` argument, to which a user need to provide the location of the signature. Like `notation blob verify ./xyz.sig.cose`.

specs/commandline/policy.md
rgnote: Good idea. https://github.com/notaryproject/specifications/issues/281

specs/commandline/blob.md
Two-Hearts: > 2. If there is no valid extension, Notation will guess the format by bruteforcing all the available formats it knows (this is already being done in notation-core-go)

I see, so we are keeping step 2. I thought we were going to remove it on the call. SGTM.

JeyJeyGao: @rgnote Then the UX will be:
```sh
SIGNATURE_NAME=xyz
notation blob sign --signature-name $SIGNATURE_NAME ./filename
```
Verify the signature:
```sh
notation blob verify $SIGNATURE_NAME.sig.jws
```
In other words, the user must know the complete name of the signature; otherwise, the notation blob verify command cannot locate the signature file. How can the user determine the full name of the signature? They can either obtain it from the output of the `notation blob sign command` or figure it out on their own.

@yizha1 @FeynmanZhou Could you provide some insights from the perspective of a PM?

specs/commandline/blob.md
```diff
@@ -0,0 +1,363 @@
...
+```text
+Error: signature verification failed: The blob is not of media type `application/my-media-octet-stream`.
+```
+
+### Verify the signature using a policy scope
+
+Use the `--policy-scope` flag to select a Policy scope to verify the signature against.
+
+```shell
+notation blob verify --policy-scope my-blob-verification-selector --signature /tmp/my-blob-signature.sig.jws /tmp/my-blob.bin
```
yizha1, line 350: For consistency, maybe you can consider using the scope from the above policy example. my-blob-verification-selector doesn't exist.

yizha1 (commented):
Thanks @rgnote I left some comments

specs/commandline/blob.md
yizha1: From a verifier point of view, I don't see big problems, since verifier needs to put the full path of the signature file. And the sign command will output the full file path. But it's better to describe it in the document, that the signature name will be added with suffixes based on the signature format that users choose.

specs/commandline/blob.md
```diff
@@ -0,0 +1,363 @@
...
+# sign a blob
+notation blob sign --signature-name my-blob-signature /tmp/my-blob.bin
+```
+
+An example for a successful signing:
+
+```console
+$ notation blob sign --signature-name my-blob-signature /tmp/my-blob.bin
+Successfully signed /tmp/my-blob.bin
+Signature written to ./my-blob-signature.sig.jws
```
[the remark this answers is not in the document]
rgnote: Correct. `.sig` is technically not an extension. It is just there to indicate the user that it is a signature file.

specs/commandline/blob.md
```diff
@@ -0,0 +1,363 @@
...
+```shell
+# Prerequisites:
+# - A signing plugin is installed. See plugin documentation (https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md) for more details.
+# - Configure the signing plugin as instructed by plugin vendor.
+
+# Add a default signing key referencing the remote key identifier, and the plugin associated with it.
+notation key add --default --name <key_name> --plugin <plugin_name> --id <remote_key_id>
+
+# sign a blob
+notation blob sign --signature-name my-blob-signature /tmp/my-blob.bin
```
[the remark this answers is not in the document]
rgnote: That is correct. Is this functionality absolutely needed? Would do you suggest adding a flag for `signature-directory`?

specs/commandline/blob.md
```diff
@@ -0,0 +1,363 @@
...
+
+```text
+Sign, Inspect, and Verify signatures associates with arbitrary blobs.
+
+Usage:
+  notation blob [command]
+
+Available Commands:
+  sign      produce a detached signature for a given blob
+  inspect   inspect a signature associated with a blob
```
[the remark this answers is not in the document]
rgnote: I sorted them based on user workflows. I don't think we follow the convention of alphabetical sorting ([example](https://github.com/notaryproject/notation/blob/main/specs/notation-cli.md))

specs/commandline/blob.md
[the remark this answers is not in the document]
rgnote: cose example is present at line 122

specs/commandline/blob.md
rgnote: Good call.

specs/commandline/blob.md
```diff
@@ -0,0 +1,365 @@
...
+An example of output messages for a successful verification:
+
+```text
+Successfully verified signature /tmp/my-blob-signature.sig.jws using policy scope `blob-verification-selector`
+
+```
+An example of output messages for an unsuccessful verification:
+
+```text
+Error: signature verification failed for Policy scope `blob-verification-selector`
```
yizha1, line 364:
```suggestion
Error: signature verification failed for policy scope `blob-verification-selector`
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,365 @@
...
+
+An example of output messages for an unsuccessful verification:
+
+```text
+Error: signature verification failed: The blob is not of media type `application/my-media-octet-stream`.
+```
+
+### Verify the signature using a policy scope
+
+Use the `--policy-scope` flag to select a Policy scope to verify the signature against.
```
yizha1, line 349:
```suggestion
Use the `--policy-scope` flag to select a policy scope to verify the signature against.
```

specs/commandline/policy.md
```diff
@@ -67,6 +67,110 @@ The `notation policy` command provides a user-friendly way to manage trust polic
...
+            },
+            "trustStores": ["ca:acme-rockets"],
+            "trustedIdentities": ["*"]
+        }
+    ]
+}
+```
+
+Note: Policy language version 1.1 renamed the field `registryScopes` from version 1.0 to `scopes`. The new field accepts values with prefixes `oci` or `blob` to limit a scope value to either OCI signature verification or Blob signature verification. While scope values with `blob` prefix can be of free-form text, values with `oci` prefix must be valid OCI references. `notation` supports both policy language versions 1.0 and 1.1. However, `notation` rejects policy configuration files with mixed terminology i.e. both `registryScopes` and `scopes` defined in a single configuration file. Users migrating from 1.0 to 1.1 can simply rename `registryScopes` to `scopes` and prefix the values with `oci`.
+Below is a sample Policy configuration file that verifies OCI artifacts using `scopes` field.
```
yizha1, line 116: Maybe a policy file including policies for both OCI and blob is the common case?

```
{
    "version": "1.1",
    "trustPolicies": [
        {
            "name": "wabbit-networks-images",
            "scopes": [ 
              "oci:registry.acme-rockets.io/software/net-monitor",
            ],…
         },
         {
           "name": "global-policy-for-OCI",
            "scopes": [ 
              "oci:*",
            ],…
         },
        {
            "name": "blob-policy",
            "scopes": [ 
              "blob:blob-verification-selector",
            ],…
         },
         {
           "name": "global-policy-for-blob",
            "scopes": [ 
              "blob:*",
            ],…
         },
```

yizha1 (commented):
Thanks @rgnote. I left additional comments.

specs/commandline/blob.md
```diff
@@ -29,11 +29,11 @@ Flags:
 Produce a detached signature for a given blob
 
 Usage:
-  notation blob sign [flags] <blob_path>
+  notation blob sign [flags] -n my-blob-signature <blob_path>
 
 Flags:
-  -s,  --signature path             output location of the detached signature
-       --media-type string          media type of the blob. If not provided, notation uses `application/octet-stream`
+  -n,  --signature-name string      friendly name for the detached signature. Signature file will be written to the currently working directory with this name plus ".sig" plus signature format as the file extension
```
shizhMSFT, line 35: Is `.sig` necessary? Can it be `mySignature.jws` or `mySignature.cose`?

shizhMSFT, line 35: Can `--signature-name` be optional? Basically, the signature filename can be default to `<blob_path>.<type>`.

For example, `notation blob sign hello.tar` generates `hello.tar.jws` like `gzip hello.tar` generates `hello.tar.gz`.

specs/commandline/blob.md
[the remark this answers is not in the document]
shizhMSFT: I think the original `--signature` is good enough and it is consistent with `notation blob verify`.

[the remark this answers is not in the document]
shizhMSFT: or `--signature-prefix` can be a candidate if `.sig` is keep.
or `--signature-name` allows folders like `foo/bar/my-blob-signature`.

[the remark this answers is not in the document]
rgnote: @shizhMSFT 
IMO these two options can be confusing. `--signature-prefix` if fine if it is just accepting the prefix for the signature file name. But if it also accepts a path, then the flag name is not very apparent.
`--signature-name` accepting a path is also bit misleading and error prone.

If we want to support arbitrary paths, I recommend doing this

1. Accept an optional flag `--signature-directory`
2. If it is provided, notation will write the <source file name>+".sig.jws" to that directory. If not, then writes to the currently working directory.
This way, the name of the file is not configurable but the path is. I think this is better than providing two many flags for configuring both name and path. 

LMK your thoughts.

specs/commandline/blob.md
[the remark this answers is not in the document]
rgnote: I think you are right. Flipped the order.

specs/commandline/policy.md
rgnote: Added few blob policies

specs/commandline/blob.md
rgnote: Not necessary but good to have to indicate that it is a signature file. Not all users of notation know that notation uses jws or cose. If we do not add `.sig` and user doesn't know anything about jws/cose (which the most of the userbase is), then it can be confusing as to why notation created the files with jws/cose extension.

rgnote: See https://github.com/notaryproject/notation/pull/811#discussion_r1384163309

FeynmanZhou: If users don't define a unique signature name, they might be hard to find signatures if there are multiple signatures generated in the same file path.

specs/commandline/blob.md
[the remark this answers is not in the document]
FeynmanZhou: I think it would be better to enable users choose which folder they want to generate the signature since this is a common scenario. 
For 2, is it really necessary to make the signature envelop type ".sig.jws" as the file extension name?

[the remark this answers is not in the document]
yizha1: @rgnote Here is the discussion result from community meeting 11/7/2023

- Notation controls the signature file name, for example, use the file name of the blob file + ".sig.jws". Then by default, the signature is stored on the current directory.
- Provide an option for users to specify the signature directory
- If users sign the blob for the 2nd time, overwrite the existing signature by asking whether users are ok with it, if users are OK, then overwriting the existing signature; if not, users can specify a different directory for storing the signature.
- consider `cp` behavior for output file path resolution. 
        - Should notation create a fully nested path or not? 
        - If its directory then should we outupt {dir}/{pattern}.{ext} 
            - if not, should we output the file to the output path and consider that it's a fully qualified filename.
        - `cp` does not create nested paths so maybe to keep this consistent with other tools and avoid surprises. notation could follow the same pattern of not creating nested paths. 

/cc: @shizhMSFT @Two-Hearts @priteshbandi feel free to provide additional comments.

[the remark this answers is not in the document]
rgnote: Perfect.

Two comments on the above proposal though.
>  If users sign the blob for the 2nd time, overwrite the existing signature by asking whether users are ok with it, if users are OK, then overwriting the existing signature; if not, users can specify a different directory for storing the signature.

I'm not a big fan interactive CLI tools as they make automation hard to achieve. That being said, we should do this.
1. Let notation fail by default if there is a signature file already
2. Provide an optional `--overwrite` flag for suppressing the above failure

> consider cp behavior for output file path resolution.

I think we should keep this behavior simple. Notation being smarter to proactively create non-existing directories or writing signatures directly when the path is not a directory, etc will only add sharp edges. I would say, let's start with the basic functionality i.e. fail if the provided path is not a directory or does not exist. This is not a one-way door. If there is enough demand for handling non-existing directories or configuring signature file name, then we can add them later.

specs/commandline/blob.md
rgnote: @FeynmanZhou 

Per https://github.com/notaryproject/notation/pull/811#discussion_r1384262523, we are moving away from users choosing signature names.

specs/commandline/blob.md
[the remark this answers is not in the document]
yizha1: @rgnote  I am OK with it. @shizhMSFT @patrickzheng200 @priteshbandi any comments?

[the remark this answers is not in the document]
priteshbandi: For overwriting signature we can support two mode:
1. Without `--overwrite` flag; ask for confirmation from user
2. With `--overwrite` flag; overwrite file without any confirmation prompt.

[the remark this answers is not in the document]
rgnote: Decided on the notary call to go with `--force`, to be consisted with rest of Notation UX.

specs/commandline/blob.md
```diff
@@ -0,0 +1,365 @@
...
+
+Flags:
+  -s, --signature path        location of the detached signature
+      --media-type string     optional media type of the blob to verify
+      --policy-scope string   optional policy scope to verify against. If not provided, notation verifies against wildcard policy if it exists.
+  -m, --user-metadata stringArray   user defined {key}={value} pairs that must be present in the signature for successful verification if provided
+  -o, --output string         output format, options: 'json', 'text' (default "text")
+  -d, --debug                 debug mode
+  -v, --verbose               verbose mode
+  -h, --help                  help for inspect
```
Two-Hearts, line 80: You might want to add the `--plugin-config` flag because user could use a plugin for verification.

specs/commandline/blob.md
```diff
@@ -0,0 +1,365 @@
...
+
+```text
+notation blob inspect [flags] /tmp/my-blob-signature.sig.jws
+```
+
+### Inspect the given detached blob signature
+
+```shell
+# Prerequisites: Signatures is produced by notation blob sign command
+notation blob inspect /tmp/my-blob-signature.sig.jws
```
Two-Hearts, line 184: Can you add an example of the following command: `notation blob inspect /tmp/my-blob-signature.sig`, where `my-blob-signature.sig` is a valid signature following the Notation spec.

specs/commandline/blob.md
```diff
@@ -0,0 +1,365 @@
...
+```shell
+
+# Prerequisites: Signature is produced on the filesystem from `notation blob sign` command.
+# Configure trust store by adding a certificate file into trust store named "wabbit-network" of type "ca"
+notation certificate add --type ca --store wabbit-networks wabbit-networks.crt
+
+# Create a JSON file named "trustpolicy.json" under directory "{NOTATION_CONFIG}".
+
+# Verify the detached signature
+notation blob verify --signature /tmp/my-blob-signature.sig.jws /tmp/my-blob.bin
```
Two-Hearts, line 288: Same here, can you add an example of command `notation blob verify --signature /tmp/my-blob-signature.sig /tmp/my-blob.bin`?

specs/commandline/blob.md
rgnote: This is what we have at the moment, correct? Can you please clarify how the command you provided is different from the existing one?

If you are referring to verifying a signature without signature format extension, I do not think that is something worth calling out in the docs. Notation should recommend users to keep extensions. Notation supporting no-extension signature files is only a convenience but should not be encouraged as it increases verification latencies.

specs/commandline/blob.md
rgnote: Same comment as above.

specs/commandline/blob.md
```diff
@@ -0,0 +1,381 @@
...
+### notation blob sign
+
+```text
+Produce a detached signature for a given blob. Signature file will be written to the currently working directory with blob file name + ".sig" + signature format as the file extension. For example, signature file name for "myBlob.bin" will be "myBlob.bin.sig.jws" for JWS signature format or "myBlob.bin.sig.cose" for COSE signature format.
+
+Usage:
+  notation blob sign [flags] <blob_path>
+
+Flags:
+  -sd,  --signature-directory path  optional path where the detached signature needs to be placed (default: currently working directory)
```
yizha1, line 35: To alignment with flags below
```suggestion
  -sd, --signature-directory path  optional path where the detached signature needs to be placed (default: currently working directory) 
```

yizha1 (commented):
LGTM with a nit, but I am not a maintainer of `notation` repo

specs/commandline/policy.md
```diff
@@ -67,6 +67,126 @@ The `notation policy` command provides a user-friendly way to manage trust polic
...
+              "level" : "audit"
+            },
+            "trustStores": ["ca:acme-rockets"],
+            "trustedIdentities": ["*"]
+        }
+    ]
+}
+```
+
+Note: Policy language version 1.1 renamed the field `registryScopes` from version 1.0 to `scopes`. The new field accepts values with prefixes `oci` or `blob` to limit a scope value to either OCI signature verification or Blob signature verification. While scope values with `blob` prefix can be of free-form text, values with `oci` prefix must be valid OCI references. `notation` supports both policy language versions 1.0 and 1.1. However, `notation` rejects policy configuration files with mixed terminology i.e. both `registryScopes` and `scopes` defined in a single configuration file. Users migrating from 1.0 to 1.1 can simply rename `registryScopes` to `scopes` and prefix the values with `oci`.
```
priteshbandi, line 115:
```suggestion
Note: Policy version 1.1 replaced field `registryScopes` from version 1.0 with `scopes`. The`scope` field accepts values with prefixes `oci` or `blob` to limit a scope value to either OCI signature verification or Blob signature verification. While scope values with `blob` prefix can be alpha-numeric text with `-` and `_` characters, values with `oci` prefix must be valid OCI references. `notation` supports both policy language versions 1.0 and 1.1. However, `notation` rejects policy configuration files with mixed terminology i.e. both `registryScopes` and `scopes` defined in a single configuration file. Users migrating from 1.0 to 1.1 can simply rename `registryScopes` to `scopes` and prefix the values with `oci`.
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,381 @@
+# notation policy
```
priteshbandi, line 1:
```suggestion
# notation blob
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,381 @@
...
+
+## Description
+
+Use `notation blob` command to sign, verify, and inspect signatures associated with arbitrary blobs. Notation can sign and verify any arbitrary bag of bits like zip files, documents, executables, etc. When a user signs a blob, `notation` produces a detached signature, which the user can transport/distribute in any medium that the user prefers along with the original blob. On the verification side, Notation can verify the blob's signature and assert that the blob has not been tampered with during its transmission. For more details, see [trust policy specification and examples](https://github.com/notaryproject/specifications/blob/main/specs/signing-and-verification-workflow.md#blob-signing-workflow).
```
priteshbandi, line 5:
```suggestion
Use `notation blob` command to sign, verify, and inspect signatures associated with arbitrary blobs. Notation can sign and verify any arbitrary bag of bits like zip files, documents, executables, etc. When a user signs a blob, `notation` produces a detached signature, which the user can transport/distribute using any medium that the user prefers along with the original blob. On the verification side, Notation can verify the blob's signature and assert that the blob has not been tampered during transmission. For more details on how to create trust policy for verifying blob, see [trust policy specification and examples](https://github.com/notaryproject/specifications/blob/main/specs/signing-and-verification-workflow.md#blob-signing-workflow).
```

specs/commandline/blob.md
priteshbandi: I think the utility which we use to generate cli only allows for single character short flags.

specs/commandline/blob.md
```diff
@@ -0,0 +1,381 @@
...
+# sign a blob
+notation blob sign /tmp/my-blob.bin
+```
+
+An example for a successful signing:
+
+```console
+$ notation blob sign /tmp/my-blob.bin
+Successfully signed /tmp/my-blob.bin
+Signature file written to ./my-blob.bin.sig.jws
```
priteshbandi, line 108: nit: Should we just say 
```suggestion
Signature file written to ./my-blob.bin.sig.jws
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,381 @@
...
+```
+
+### Inspect the given detached blob signature with JSON Output
+
+```shell
+notation blob inspect -o json /tmp/my-blob.bin.sig.jws
+```
+
+## Verify detached blob signatures
+The `notation blob verify` command can be used to verify blob signatures. In order to verify signatures, user will need to setup a policy configuration file with Policies scoped to blobs. Below are three examples of how a policy configuration file can be setup for verifying blob signatures.
```
priteshbandi, line 246:
```suggestion
The `notation blob verify` command can be used to verify blob signatures. In order to verify signatures, user will need to setup a trust policy file with Policies scoped to blobs. Below are three examples of how a policy configuration file can be setup for verifying blob signatures.
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,381 @@
...
+```shell
+
+# Prerequisites: Signature is produced on the filesystem from `notation blob sign` command.
+# Configure trust store by adding a certificate file into trust store named "wabbit-network" of type "ca"
+notation certificate add --type ca --store wabbit-networks wabbit-networks.crt
+
+# Create a JSON file named "trustpolicy.json" under directory "{NOTATION_CONFIG}".
+
+# Verify the detached signature
+notation blob verify --signature /tmp/my-blob.bin.sig.jws /tmp/my-blob.bin
```
priteshbandi, line 304: qq: will we allow multiple value for --signature?

priteshbandi (commented):
Should we add some note, for help of top level sign, verify, inspect and list that they only applies to oci.

specs/commandline/blob.md
rgnote: From UX perspective, `sd` makes sense. If the library doesn't support, we can revisit this.

specs/commandline/blob.md
rgnote: As an engineer, I love to see outputs that have "Successfully" in it. :)
I think we can keep it to indicate that signing had completed successfully.

specs/commandline/blob.md
rgnote: Probably. When that happens, the `--signature` flag can accept comma separated signature files.

rgnote:
> Should we add some note, for help of top level sign, verify, inspect and list that they only applies to oci.

They were already updated to note that they work with OCI artifacts.

priteshbandi (dismissed):
Approving this pull request, and we can address the following comments in the next PR:

1. Support 'blob sign' and 'verify' commands to function with standard input and output?
1. In my opinion, merely mentioning 'oci' in the command description (or in help) isn't sufficient; it should be more prominent

specs/commandline/policy.md
```diff
@@ -112,7 +112,7 @@ Policy language version 1.1 added support for verifying signatures associated wi
...
+Note: Policy version 1.1 replaced the field `registryScopes` from version 1.0 with `scopes`. The new field accepts values with prefixes `oci` or `blob` to limit a scope value to either OCI signature verification or Blob signature verification. While scope values with `blob` prefix can be alpha-numeric text with `-` and `_` characters, values with `oci` prefix must be valid OCI references. `notation` supports both policy language versions 1.0 and 1.1. However, `notation` rejects policy configuration files with mixed terminology i.e. both `registryScopes` and `scopes` defined in a single configuration file. Users migrating from 1.0 to 1.1 can simply rename `registryScopes` to `scopes` and prefix the values with `oci`.
```
yizha1, line 115: I think we need to update this spec: https://github.com/notaryproject/specifications/blob/main/specs/trust-store-trust-policy.md, and here just refer to the specification

yizha1 (commented):
LGTM with some minor comments.

specs/commandline/sign.md
```diff
@@ -2,7 +2,7 @@
 
 ## Description
 
-Use `notation sign` to sign artifacts.
+Use `notation sign` to sign OCI artifacts.
```
yizha1, line 5: Per Pritesh's comment maybe we can update it to 
```suggestion
Use `notation sign` to sign artifacts stored in OCI compliant registries
```

specs/commandline/policy.md
rgnote: https://github.com/notaryproject/specifications/pull/283

specs/commandline/blob.md
```diff
@@ -0,0 +1,381 @@
...
+An example of output messages for a successful verification:
+
+```text
+Successfully verified signature /tmp/my-blob.bin.sig.jws using policy scope `blob-verification-selector`
+
+```
+An example of output messages for an unsuccessful verification:
+
+```text
+Error: signature verification failed for Policy scope `blob-verification-selector`
```
Two-Hearts, line 380: nit:
```suggestion
Error: signature verification failed for policy scope `blob-verification-selector`
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,381 @@
...
+
+```text
+notation blob inspect [flags] /tmp/my-blob.bin.sig.jws
+```
+
+### Inspect the given detached blob signature
+
+```shell
+# Prerequisites: Signatures is produced by notation blob sign command
+notation blob inspect /tmp/my-blob.bin.sig.jws
```
Two-Hearts, line 200: Just curious, how does Notation tell apart signatures generated by `blob sign` and `sign` again? In other words, is there a fundamental difference that's ALWAYS making `detached signature` and `OCI signature` mutually exclusive? 

Because from the spec [PR](https://github.com/notaryproject/specifications/pull/283/files#diff-a4742f16cbae7ff1d39bafc711c835c1133dea5722c1922e522022ebdefac9f5), it seems that given a signature file, it is possible that it's both a valid `detached signature` and a valid `OCI signature`. If I pull an `OCI signature` from a registry and save it as a file, then I use `notation blob inspect` to inspect it, would Notation return an error?

rgnote: > If I pull an OCI signature from a registry and save it as a file, then I use notation blob inspect to inspect it, would Notation return an error?

No. And that is the right behavior and both OCI and Blob signatures follow the same scheme. 

>  I think for inspect, we could just add a new flag to cover all the scenarios: notation inspect --blob /tmp/my-blob.bin.sig.jws

The signature scheme is not different but aspects like signature format type, OCI signature manifest vs no manifest for blobs, etc. are couple of differences. It was decided in the Notary meeting to introduce `blob` sub-command to tackle these differences.

specs/commandline/inspect.md
```diff
@@ -24,6 +24,8 @@ Upon successful execution, both the digest of the signed artifact and the digest
         └── <signed artifact>
 ```
 
+NOTE: This command is for inspecting signatures associated with OCI artifacts only. Use `notation blob inspect` command for inspecting signatures associated with arbitrary blobs.
```
shizhMSFT, line 27: nit:
```suggestion
> [!NOTE]
> This command is for inspecting signatures associated with OCI artifacts only. Use `notation blob inspect` command for inspecting signatures associated with arbitrary blobs.
```
> [!NOTE]
> This command is for inspecting signatures associated with OCI artifacts only. Use `notation blob inspect` command for inspecting signatures associated with arbitrary blobs.

specs/commandline/blob.md
```diff
@@ -26,20 +26,21 @@ Flags:
...
-Produce a detached signature for a given blob
+Produce a detached signature for a given blob. Signature file will be written to the currently working directory with blob file name + ".sig" + signature format as the file extension. For example, signature file name for "myBlob.bin" will be "myBlob.bin.sig.jws" for JWS signature format or "myBlob.bin.sig.cose" for COSE signature format.
 
 Usage:
-  notation blob sign [flags] -n my-blob-signature <blob_path>
+  notation blob sign [flags] <blob_path>
 
 Flags:
-  -n,  --signature-name string      friendly name for the detached signature. Signature file will be written to the currently working directory with this name plus ".sig" plus signature format as the file extension
+  -sd, --signature-directory string optional path where the blob signature needs to be placed (default: currently working directory)
```
shizhMSFT, line 35: The short-hand `-sd` is not valid. It will be parsed as `-s` and `-d`.

yizha1:
@rgnote Would you mind resolving comments of this PR? Thanks.

specs/commandline/blob.md
rgnote: Will fix it. Thanks

specs/commandline/blob.md
```diff
@@ -0,0 +1,381 @@
...
+  verify    verify a signature associated with a blob
+
+Flags:
+  -h, --help   help for blob
+```
+
+### notation blob sign
+
+```text
+Produce a detached signature for a given blob. Signature file will be written to the currently working directory with blob file name + ".sig" + signature format as the file extension. For example, signature file name for "myBlob.bin" will be "myBlob.bin.sig.jws" for JWS signature format or "myBlob.bin.sig.cose" for COSE signature format.
```
gokarnm, line 29:
```suggestion
Produce a signature for a given blob. A detached signature file will be written to the currently working directory with blob file name + ".sig" + signature format as the file extension. For example, signature file name for "myBlob.bin" will be "myBlob.bin.sig.jws" for JWS signature format or "myBlob.bin.sig.cose" for COSE signature format.
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,381 @@
...
+      --plugin-config stringArray   {key}={value} pairs that are passed as it is to a plugin, if the verification is associated with a verification plugin, refer plugin documentation to set appropriate values
+  -o, --output string         output format, options: 'json', 'text' (default "text")
+  -d, --debug                 debug mode
+  -v, --verbose               verbose mode
+  -h, --help                  help for inspect
+```
+
+## Usage
+
+## Produce detached blob signatures
```
gokarnm, line 87:
```suggestion
## Produce blob signatures
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,381 @@
...
+```
+
+### Sign a blob and skip user confirmations when overwriting existing signature
+```shell
+$ notation blob sign --force /tmp/my-blob.bin
+Successfully signed /tmp/my-blob.bin
+Signature file overwritten to ./my-blob.bin.sig.jws
+```
+
+### Sign a blob with on-demand remote key
```
gokarnm, line 125: What does on-demand mean here? Can we just say sign with (named) plugin?

specs/commandline/blob.md
```diff
@@ -0,0 +1,381 @@
...
+```
+
+### Inspect the given blob signature with JSON Output
+
+```shell
+notation blob inspect -o json /tmp/my-blob.bin.sig.jws
+```
+
+## Verify blob signatures
+The `notation blob verify` command can be used to verify blob signatures. In order to verify signatures, user will need to setup a trust policy file with Policies scoped to blobs. Below are three examples of how a policy configuration file can be setup for verifying blob signatures.
```
gokarnm, line 246:
```suggestion
The `notation blob verify` command can be used to verify blob signatures. In order to verify signatures, user will need to setup a trust policy file with Policies for blobs. Below are three examples of how a policy configuration file can be setup for verifying blob signatures.
```

gokarnm (changes requested):

specs/commandline/blob.md
```diff
@@ -0,0 +1,381 @@
...
+### notation blob sign
+
+```text
+Produce a detached signature for a given blob. Signature file will be written to the currently working directory with blob file name + ".sig" + signature format as the file extension. For example, signature file name for "myBlob.bin" will be "myBlob.bin.sig.jws" for JWS signature format or "myBlob.bin.sig.cose" for COSE signature format.
+
+Usage:
+  notation blob sign [flags] <blob_path>
+
+Flags:
+  -d,  --signature-directory string optional path where the blob signature needs to be placed (default: currently working directory)
```
priteshbandi, line 35: `-d` conflicts with debug flag. I would suggest dropping short flag.

specs/commandline/blob.md
```diff
@@ -0,0 +1,381 @@
...
+# Verify the signature and verify that io.wabbit-networks.buildId=123 is present in the signed payload
+notation blob verify --user-metadata io.wabbit-networks.buildId=123 --signature /tmp/my-blob.bin.sig.jws /tmp/my-blob.bin
+```
+
+An example of output messages for a successful verification:
+
+```text
+Successfully verified signature /tmp/my-blob.bin.sig.jws
+
+The blob signature is having the following user metadata.
```
priteshbandi, line 327:
```suggestion
The signature contains the following user metadata:
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,381 @@
...
+Successfully verified signature /tmp/my-blob.bin.sig.jws
+
+The blob is of media type `application/my-media-octet-stream`.
+
+```
+
+An example of output messages for an unsuccessful verification:
+
+```text
+Error: signature verification failed: The blob's media type `application/xyz` and not matching `application/my-media-octet-stream`.
```
priteshbandi, line 360:
```suggestion
Error: Signature verification failed due to a mismatch in the blob's media type 'application/xyz' and the expected type 'application/my-media-octet-stream'.
```

specs/commandline/blob.md
Two-Hearts: +1

specs/commandline/blob.md
```diff
@@ -0,0 +1,381 @@
...
+```text
+Verify a signature associated with a blob
+
+Usage:
+  notation blob verify [flags] --signature <signature_path> <blob_path>
+
+Flags:
+  -s, --signature string      location of the blob signature file
+      --media-type string     optional media type of the blob to verify
+      --policy-scope string   optional policy scope to verify against. If not provided, notation verifies against wildcard policy if it exists.
```
Two-Hearts, line 76: We need to be more clear here. Because a valid version 1.1 trust policy can have two different trust policy statements,  one of them has scope `[oci:*]`, and the other one has scope `[blob:*]`. For `notation blob verify` command, when `--policy-scope` is not set, only `[blob:*]` will be used right? In other words, if `[blob:*]` is missing, even if the user set up `[oci:*]`, the verification should fail?

specs/commandline/policy.md
```diff
@@ -67,6 +67,126 @@ The `notation policy` command provides a user-friendly way to manage trust polic
...
+        {
+            "name": "wildcard-blob-verification-policy",
+            "scopes": [ "blob:*" ],
+            "signatureVerification": {
+              "level" : "audit"
+            },
+            "trustStores": ["ca:acme-rockets"],
+            "trustedIdentities": ["*"]
+        }
+    ]
```
Two-Hearts, line 186: Same as above, we need to clarify that:
1. `[blob:*]` only applies to `notation blob` commands.
2. `[oci: *]` only applies to non `notation blob` commands.

Two-Hearts (commented):
The current UX design LGTM. Left a couple of comments on trust policy wildcard scopes.

specs/commandline/blob.md
```diff
@@ -0,0 +1,381 @@
...
+Inspect a signature associated with a blob
+
+Usage:
+  notation blob inspect [flags] <signature_path>
+
+Flags:
+  -o, --output string         output format, options: 'json', 'text' (default "text")
+  -d, --debug                 debug mode
+  -v, --verbose               verbose mode
+  -h, --help                  help for inspect
```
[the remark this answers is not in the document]
Two-Hearts: @rgnote Maybe you forgot to update this one?

specs/commandline/blob.md
shizhMSFT: +1

rgnote:
Closing this PR as this is stale. Will raise a new one with the updated CX as described in https://hackmd.io/@-KPyDkW6QfGA-pldFa13pA/ByuHffALa

specs/commandline/blob.md
rgnote: The proposal has evolved from scopes to separate policy documents. See https://hackmd.io/@-KPyDkW6QfGA-pldFa13pA/ByuHffALa

I will update this PR with the new CX

rgnote:
Latest revision addresses the new CX proposed as part of https://hackmd.io/@-KPyDkW6QfGA-pldFa13pA/ByuHffALa

specs/notation-cli.md
```diff
@@ -28,15 +29,16 @@ Usage:
...
+  list        List signatures of a signed OCI artifact
+  login       Log into OCI registries
+  logout      Log out from the logged in OCI registries
   plugin      Manage plugins
   policy      Manage trust policy configuration for signature verification
-  sign        Sign artifacts
-  verify      Verify artifacts
+  sign        Sign OCI artifacts
+  verify      Verify OCI artifacts
+  blobs       Sign, verify and inspect singatures associated with blobs
```
whalelines, line 40: Elsewhere in the changes this command is stated to be `blob` not `blobs`. Which is correct? Also "signature" appears to have a typo.

specs/notation-cli.md
```diff
@@ -28,15 +29,16 @@ Usage:
...
+  inspect     Inspect all signatures associated with a signed OCI artifact
```
whalelines, line 41: Prior to these changes, the list of commands was sorted alphabetically, why reorder?

rgnote: Fixed

specs/notation-cli.md
```diff
@@ -7,15 +7,16 @@ This spec contains reference information on using notation commands. Each comman
...
 | [plugin](./commandline/plugin.md)           | Manage plugins                                                         |
-| [policy](./commandline/policy.md)           | Manage trust policy configuration for signature verification |
-| [sign](./commandline/sign.md)               | Sign artifacts                                                         |
-| [verify](./commandline/verify.md)           | Verify artifacts                                                       |
+| [policy](./commandline/policy.md)           | Manage trust policy configuration for signature verification           |
+| [sign](./commandline/sign.md)               | Sign OCI artifacts                                                     |
+| [verify](./commandline/verify.md)           | Verify OCI artifacts                                                   |
+| [inspect](./commandline/inspect.md)         | Inspect OCI signatures                                                 |
+| [blob](./commandline/blob.md)               | Sign, verify and inspect singatures associated with blobs                                |
```
whalelines, line 19: Same as below, these used to be sorted too, typo in "signatures", and last `|` does not line up.

rgnote: Fixed in the latest revision. Apologies for the churn. I somehow screwed up my local dev environment. I thought I fixed this in my earlier commit.

specs/commandline/blob.md
```diff
@@ -0,0 +1,504 @@
...
+            "trustStores": [ 
+              "ca:wabbit-networks",
+            ],
+            "trustedIdentities": [
+                "x509.subject: C=US, ST=WA, L=Seattle, O=wabbit-networks.io, OU=Security Tools"
+            ]
+        },
+        {
+            "name": "wildcard-verification-policy",
+            "wildcardPolicy": true,
```
JeyJeyGao, line 403: Will there only be one wildcard policy?

specs/commandline/blob.md
```diff
@@ -0,0 +1,504 @@
...
+### notation blob sign
+
+```text
+Produce a signature for a given blob. A detached signature file will be written to the currently working directory with blob file name + ".sig" + signature format as the file extension. For example, signature file name for "myBlob.bin" will be "myBlob.bin.sig.jws" for JWS signature format or "myBlob.bin.sig.cose" for COSE signature format.
+
+Usage:
+  notation blob sign [flags] <blob_path>
+
+Flags:
+  --signature-directory string optional path where the blob signature needs to be placed (default: currently working directory)
```
JeyJeyGao, line 80: Could we simplify the flag name since it may be used very frequently? For example, `-D`.

yizha1 (commented):
Thanks @rgnote . I have two questions:
- Regarding trust policy configuration, should we complete PR https://github.com/notaryproject/specifications/pull/283 first?
- Why do we need a new command `notation blob policy` for managing trust policy?

specs/commandline/blob.md
rgnote: Correct. Just like OCI policies, there can only be one wildcard policy for blob policies

specs/commandline/blob.md
rgnote: `-d` conflicts with debug flag. We decided in one of the earlier threads to drop the short-hand flag.

Two-Hearts:
@rgnote Looks like this PR is still WIP? https://github.com/notaryproject/specifications/pull/283
I think it's better to complete it with this PR together? Because https://github.com/notaryproject/specifications/pull/283 contains the new trust policy which is one of the major changes in this design.
It will be much easier to review the implementation PRs with the whole design in mind.

specs/commandline/blob.md
```diff
@@ -0,0 +1,504 @@
...
+User can refer to the below trust policy configuration sample `trustpolicy.blob.json` that is applicable for verifying signed arbitrary blobs using `notation blob verify` command. In this sample, there are three policies configured for different requirements:
+
+- The Policy named "wabbit-networks-policy" is for verifying blob artifacts signed by Wabbit Networks.
+- Policy named "skip-verification-policy" is for skipping verification on blob artifacts.
+- Policy "wildcard-verification-policy" is for auditing verification results when user wants to apply a wildcard policy by not providing `--policy-name` argument in `notation blob verify` command.
+
+```jsonc
+{
+    "version": "1.0",
+    "policyType": "blob",
```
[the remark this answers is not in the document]
rgnote: 1. I included `policyType` so that `policy export`, `policy import` commands can understand what policy types they are dealing with. But in retrospect, this flag causes some confusion (as we do not want to make any breaking changes to the OCI policies). That being said, I will move this field.
2. Added a bunch of policy examples in https://github.com/notaryproject/specifications/pull/283/files

specs/commandline/blob.md
```diff
@@ -0,0 +1,504 @@
...
+    ]
+}
+```
+
+## Outline
+
+### notation blob command
+
+```text
+Sign, inspect, and verify signatures associated with arbitrary blobs.
```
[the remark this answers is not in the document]
rgnote: Sure

rgnote:
Signing Scheme and Trust Policy updates at https://github.com/notaryproject/specifications/pull/283

yizha1:
> @yizha1
> 
> You can see the proposal at https://hackmd.io/@-KPyDkW6QfGA-pldFa13pA/ByuHffALa
> 
> * I don't think this PR blocks on [Support arbitrary blob signing specifications#283](https://github.com/notaryproject/specifications/pull/283). Both PRs are implementing the same proposal. If there is a benefit in completing the policy spec PR first, please let me know.
> * According to the new proposal (https://hackmd.io/@-KPyDkW6QfGA-pldFa13pA/ByuHffALa), there will be two policy configuration files in Notation. One for Blobs and one for OCI artifacts. Just like how we are separating the Sign/Verify/Inspect commands for blobs, we are doing the same for "policy" commands as well as that gives a nice separation between Blob verification and OCI verification workflows.

Thanks @rgnote. I got your point.

priteshbandi (commented):
Added couple of nitp picks; will continue reviewing today.

specs/commandline/blob.md
```diff
@@ -0,0 +1,502 @@
...
+
+   ```shell
+   notation blob policy import ./blob_trust_policy.json
+   ```
+
+## Verify blob signatures
+The `notation blob verify` command can be used to verify blob signatures. In order to verify signatures, user will need to setup a trust policy file `trustpolicy.blob.json` with Policies for blobs. Below are two examples of how a policy configuration file can be setup for verifying blob signatures.
+
+- The Policy named "wabbit-networks-policy" is for verifying blob artifacts signed by Wabbit Networks.
+- Policy "wildcard-verification-policy" is for auditing verification results when user wants to apply a wildcard policy by not providing `--policy-name` argument in `notation blob verify` command.
```
priteshbandi, line 381:
```suggestion
- Policy named "wildcard-verification-policy" is for auditing verification results when user doesn't not provides `--policy-name` argument in `notation blob verify` command.
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,502 @@
...
+}
+```
+
+### Verify the signature of a blob
+
+Configure trust store and trust policy properly before using `notation blob verify` command.
+
+```shell
+
+# Prerequisites: Signature is produced on the filesystem from `notation blob sign` command.
```
priteshbandi, line 418: the signature might have been generate at other place or could have been generated using another notation-compatible tool.
```suggestion
# Prerequisites: Blob and its associated signature is present on the filesystem.
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,502 @@
...
+# Configure trust store by adding a certificate file into trust store named "wabbit-network" of type "ca"
+notation certificate add --type ca --store wabbit-networks wabbit-networks.crt
+
+# Create a JSON file named "trustpolicy.blob.json" under directory "{NOTATION_CONFIG}".
```
priteshbandi, line 422: Missing command for this?

specs/commandline/blob.md
priteshbandi: Could we please document this and outline its expected output?

In my opinion, to start we should allow only single signature verification.

rgnote: Let's differ it to when we actually support multiple signatures.

specs/commandline/blob.md
```diff
@@ -0,0 +1,502 @@
...
+### notation blob verify
+
+```text
+Verify a signature associated with a blob
+
+Usage:
+  notation blob verify [flags] --signature <signature_path> <blob_path>
+
+Flags:
+  -s, --signature string      location of the blob signature file
```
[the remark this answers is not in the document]
rgnote: Resolving as I think this is not a big concern.

specs/commandline/blob.md
```diff
@@ -0,0 +1,502 @@
...
+# Prerequisites: Signatures is produced by notation blob sign command
+notation blob inspect /tmp/my-blob.bin.sig.jws
+```
+
+An example output:
+```shell
+Inspecting /tmp/my-blob.bin.sig.jws
+/tmp/my-blob.bin.sig.jws
+└── application/octet-stream
+    ├── sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```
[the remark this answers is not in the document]
rgnote: Good catch

specs/commandline/blob.md
```diff
@@ -0,0 +1,502 @@
...
+
+   ```shell
+   notation blob policy show > ./blob_trust_policy.json
+   ```
+
+2. Edit the exported JSON file "blob_trust_policy.json", update trust policy configuration and save the file.
+3. Import trust policy configuration from the file.
+
+   ```shell
+   notation blob policy import ./blob_trust_policy.json
```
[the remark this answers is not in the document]
rgnote: Not necessarily. If force is not given, notation will just prompt the user.

specs/commandline/blob.md
rgnote: Not a command. This is setting up the policy file, which is out of scope for this example. I clarified the language a bit so that it is clear.

specs/commandline/blob.md
```diff
@@ -0,0 +1,502 @@
...
+    ]
+}
+```
+
+## Outline
+
+### notation blob command
+
+```text
+Sign, inspect, and verify signatures and setup trust policies.
```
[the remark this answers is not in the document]
rgnote: Better

specs/commandline/blob.md
```diff
@@ -0,0 +1,502 @@
...
+
+Usage:
+  notation blob [command]
+
+Available Commands:
+  inspect   inspect a signature associated with a blob
+  policy    manage trust policy configuration file for signed blobs
```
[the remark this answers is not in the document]
rgnote: Sure

specs/commandline/blob.md
rgnote: To clarify, we are not allowing multiple signatures in this spec.

specs/commandline/blob.md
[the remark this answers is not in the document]
priteshbandi: Do we truly require a shorthand for --signature string? If not, IMO we should lean towards dropping it.

priteshbandi (commented):
LGTM with above 2 comments addressed.

vaninrao10 (approved):
LGTM

rgnote:
@Two-Hearts @JeyJeyGao @yizha1 
Can you approve this PR?

Two-Hearts (dismissed):
LGTM

specs/commandline/policy.md
```diff
@@ -72,7 +74,7 @@ The `notation policy` command provides a user-friendly way to manage trust polic
 ### notation policy command
 
 ```text
-Manage trust policy configuration for signature verification.
+Manage trust policy configuration for OCI image signature verification.
```
yizha1, line 77:
```suggestion
Manage trust policy configuration for OCI artifact signature verification.
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,501 @@
...
+```text
+Verify a signature associated with a blob
+
+Usage:
+  notation blob verify [flags] --signature <signature_path> <blob_path>
+
+Flags:
+      --signature string      location of the blob signature file
+      --media-type string     optional media type of the blob to verify
+      --policy-name string   optional policy name to verify against. If not provided, notation verifies against the global policy if it exists.
```
yizha1, line 161:
```suggestion
      --policy-name string    optional policy name to verify against. If not provided, notation verifies against the global policy if it exists.
```

yizha1 (approved):
LGTM with nits

specs/commandline/blob.md
```diff
@@ -0,0 +1,501 @@
...
+# sign a blob
+notation blob sign /tmp/my-blob.bin
+```
+
+An example for a successful signing:
+
+```console
+$ notation blob sign /tmp/my-blob.bin
+Successfully signed /tmp/my-blob.bin
+Signature file written to ./my-blob.bin.sig.jws
```
shizhMSFT, line 193:
```suggestion
Signature file written to my-blob.bin.sig.jws
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,501 @@
...
+```
+
+### Sign a blob by generating the signature in a particular directory
+```shell
```
shizhMSFT, line 197:
```suggestion
```console
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,501 @@
...
+$ notation blob sign --signature-directory /tmp/xyz/sigs /tmp/my-blob.bin
+Successfully signed /tmp/my-blob.bin
+Signature file written to /tmp/xyz/sigs/my-blob.bin.sig.jws
+```
+
+### Sign a blob and skip user confirmations when overwriting existing signature
+```shell
```
shizhMSFT, line 204:
```suggestion
```console
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,501 @@
...
+$ notation blob sign --force /tmp/my-blob.bin
+Successfully signed /tmp/my-blob.bin
+Signature file overwritten to ./my-blob.bin.sig.jws
```
shizhMSFT, line 207:
```suggestion
Signature file overwritten to my-blob.bin.sig.jws
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,501 @@
...
+Flags:
+       --signature-directory string optional path where the blob signature needs to be placed (default: currently working directory) 
+       --media-type string          optional media type of the blob (default: "application/octet-stream")
+  -e,  --expiry duration            optional expiry that provides a "best by use" time for the blob. The duration is specified in minutes(m) and/or hours(h). For example: 12h, 30m, 3h20m
+       --id string                  key id (required if --plugin is set). This is mutually exclusive with the --key flag
+  -k,  --key string                 signing key name, for a key previously added to notation's key list. This is mutually exclusive with the --id and --plugin flags
+       --plugin string              signing plugin name. This is mutually exclusive with the --key flag
+       --plugin-config stringArray  {key}={value} pairs that are passed as it is to a plugin, refer plugin's documentation to set appropriate values.
+       --signature-format string    signature envelope format, options: "jws", "cose" (default "jws")
+       --force                      skip user confirmation and force overwrite the existing blob signature file
```
shizhMSFT, line 87: What's the real-world scenario of overwriting an existing blob signature file?

specs/commandline/blob.md
```diff
@@ -0,0 +1,501 @@
...
+```
+
+### Sign a blob using COSE signature format
+
+```shell
+# Prerequisites:
+# A default signing key is configured using CLI "notation key"
+
+# Use option "--signature-format" to set the signature format to COSE.
+notation blob sign --signature-format cose /tmp/my-blob.bin
```
shizhMSFT, line 223:
```suggestion
$ notation blob sign --signature-format cose /tmp/my-blob.bin
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,501 @@
...
+
+### Sign a blob with a plugin
+
+```shell
+notation blob sign --plugin <plugin_name> --id <remote_key_id> /tmp/my-blob.bin
+```
+
+### Sign a blob using COSE signature format
+
+```shell
```
shizhMSFT, line 218:
```suggestion
```console
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,501 @@
...
+notation key add --default --name <key_name> --plugin <plugin_name> --id <remote_key_id>
+
+# sign a blob
+notation blob sign /tmp/my-blob.bin
+```
+
+An example for a successful signing:
+
+```console
+$ notation blob sign /tmp/my-blob.bin
```
shizhMSFT, line 191: I'm wondering why all examples are signing a file with an _absolute_ path? Can `notation blob sign/verify` take a _relative_ path?

specs/commandline/blob.md
rgnote: An example : User provided wrong media-type or plugin first time and they want to correct it by re-running the `sign` command.

specs/commandline/blob.md
rgnote: Relative paths should be supported. Its just that there were not called out explicitly. Added couple of examples for sign and verify.

priteshbandi (approved):
LGTM

specs/commandline/blob.md
rgnote: @shizhMSFT Do you have any follow up questions on this? If not, can you please approve the PR?

specs/commandline/blob.md
```diff
@@ -0,0 +1,508 @@
...
+Successfully signed ./relative/path/my-blob.bin
+Signature file written to ./relative/path/my-blob.bin.sig.jws
+```
+
+### Sign a blob and skip user confirmations when overwriting existing signature
+```console
+$ notation blob sign --force /tmp/my-blob.bin
+Successfully signed /tmp/my-blob.bin
+Signature file overwritten to /tmp/my-blob.bin.sig.jws
+```
```
shizhMSFT, line 215: Here comes the inconsistency.

The definition above
>        --signature-directory string optional path where the blob signature needs to be placed (default: currently working directory) 

states that the signature should be default written to the current working directory.

However, the above two examples are not (so as other examples).

shizhMSFT: Shouldn't it be

------

### Sign a blob using a relative path
```console
$ notation blob sign ./relative/path/my-blob.bin
Successfully signed ./relative/path/my-blob.bin
Signature file written to my-blob.bin.sig.jws
```

### Sign a blob and skip user confirmations when overwriting existing signature
```console
$ notation blob sign --force /tmp/my-blob.bin
Successfully signed /tmp/my-blob.bin
Signature file overwritten to my-blob.bin.sig.jws
```

------

or print the absolute path given the current working directory `/test`

### Sign a blob using a relative path
```console
$ notation blob sign ./relative/path/my-blob.bin
Successfully signed ./relative/path/my-blob.bin
Signature file written to /test/my-blob.bin.sig.jws
```

### Sign a blob and skip user confirmations when overwriting existing signature
```console
$ notation blob sign --force /tmp/my-blob.bin
Successfully signed /tmp/my-blob.bin
Signature file overwritten to /test/my-blob.bin.sig.jws
```

specs/commandline/blob.md
shizhMSFT: > User provided wrong media-type or plugin first time and they want to correct it by re-running the sign command.

In that scenario, customers should delete the previous signature securely. Instead of overwriting, customers should use `shred` command to securely delete it (so that it cannot be recovered from file system) or even revoke the signature if it is already spread.

rgnote: What is your recommendation here?

specs/commandline/blob.md
rgnote: I like the absolute path option. Makes it less ambiguous. Will update with this.

specs/commandline/blob.md
rgnote: @shizhMSFT ^

shizhMSFT: I'd recommend treating "overwriting the existing blob signature file" as an error.

@priteshbandi @yizha1 @FeynmanZhou We also need a doc for guidance in case the customer mis-signed an artifact or a blob.

rgnote: I think this requires more conversation. I removed the "overwriting" feature for now. We can add it back if needed in the future.

specs/commandline/blob.md
```diff
@@ -0,0 +1,500 @@
...
+# sign a blob
+notation blob sign /tmp/my-blob.bin
+```
+
+An example for a successful signing:
+
+```console
+$ notation blob sign /tmp/my-blob.bin
+Successfully signed /tmp/my-blob.bin
+Signature file written to /tmp/my-blob.bin.sig.jws
```
shizhMSFT, line 192:
```suggestion
Signature file written to /absolute/path/to/cwd/my-blob.bin.sig.jws
```

specs/commandline/blob.md
```diff
@@ -0,0 +1,500 @@
...
+### Sign a blob using COSE signature format
+
+```console
+# Prerequisites:
+# A default signing key is configured using CLI "notation key"
+
+# Use option "--signature-format" to set the signature format to COSE.
+$ notation blob sign --signature-format cose /tmp/my-blob.bin
+Successfully signed /tmp/my-blob.bin
+Signature file written to /tmp/my-blob.bin.sig.cose
```
shizhMSFT, line 224:
```suggestion
Signature file written to /absolute/path/to/cwd/my-blob.bin.sig.cose
```

shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

PR #827, opened 2023-11-08
feat: notation plugin install command

Two-Hearts:
This PR is based on the new spec: https://github.com/notaryproject/notation/pull/809

It supports following notation plugin installations:
1. Install from `.tar.gz/.zip` file in file system
2. Install from `HTTPS URL`, SHA256 checksum is required
3. Install a plugin executable file in file system

Test: Corresponding E2E test cases are added in this PR as well.

<del>Note: This PR depends on the notation-go PR: https://github.com/notaryproject/notation-go/pull/364

The notation-go PR: https://github.com/notaryproject/notation-go/pull/364 has been Merged.


cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,315 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create notationPluginTmpDir, %w", err)
+	}
+	defer os.RemoveAll(tmpDir)
+	tmpFilePath := filepath.Join(tmpDir, fileName)
+	pluginFile, err := os.OpenFile(tmpFilePath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, fmode)
+	if err != nil {
+		return err
+	}
+	if _, err := io.Copy(pluginFile, fileReader); err != nil {
```
shizhMSFT, line 287: `pluginFile` is never closed if `io.Copy` fails. Due to file lock, line 281 will fail silently.

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// ValidatePluginMetadata validates plugin metadata given plugin name and path,
+// returns the plugin version on success
+func ValidatePluginMetadata(ctx context.Context, pluginName, path string) (string, error) {
+	plugin, err := plugin.NewCLIPlugin(ctx, pluginName, path)
+	if err != nil {
+		return "", err
+	}
+	metadata, err := plugin.GetMetadata(ctx, &proto.GetMetadataRequest{})
```
shizhMSFT, line 94: `pluginName` is not checked against the metadata.

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+	// Check server response
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("bad status: %s", resp.Status)
+	}
+	// Writer the body to file
+	_, err = io.Copy(out, resp.Body)
```
shizhMSFT, line 122: There is no limit on the `resp.Body`. Therefore, the code is vulnerable to excessive downloading attack. A malicious server can blow up your disk.

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,138 @@
...
+		return err
+	}
+	defer r.Close()
+	dgst, err := digest.FromReader(r)
+	if err != nil {
+		return err
+	}
+	enc := dgst.Encoded()
+	if enc != checkSum {
+		return fmt.Errorf("plugin checksum does not match user input. Expecting %s", checkSum)
```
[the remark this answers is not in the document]
Two-Hearts: `checkSum` is the user input, and it's the expected value.
The real checksum computed by Notation is not displayed here by design. The reason is that if we display it, the user could just copy and paste the value and bypass the check. If they do want to bypass the check, they should find another way to get it, which is not in Notation's scope.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,315 @@
...
+	source = strings.TrimSpace(source)
+	// check file path
+	_, fileError := os.Stat(source)
+	if fileError == nil {
+		return notationplugin.TypeFile, nil
+	}
+	// check url
+	if strings.HasPrefix(source, httpsPrefix) {
+		return notationplugin.TypeURL, nil
+	}
```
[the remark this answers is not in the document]
Two-Hearts: The order actually matters here. This was discussed in the [spec PR](https://github.com/notaryproject/notation/pull/809#discussion_r1386052599). It's similar to [helm](https://github.com/helm/helm/blob/d70e293e5fea032e843c29d15c7e85e78b0aca6b/pkg/action/install.go#L718).
So if the input is both a valid file path and URL, we take it as file path.

cmd/notation/internal/plugin/plugin.go
Two-Hearts: It is checked here: https://github.com/notaryproject/notation-go/blob/1bc5a3f8c4d834440deeb3dc0f6855403a6d730d/plugin/plugin.go#L110

cmd/notation/internal/plugin/plugin.go
Two-Hearts: The size of the response could vary greatly depending on the size of the plugin. Do we have a suggested size limit here?

cmd/notation/plugin/install.go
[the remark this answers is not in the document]
Two-Hearts: Updated the logic.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+		return nil
+	default:
+		return errors.New("failed to install the plugin, invalid file format. Only support .tar.gz and .zip")
+	}
+}
+
+// installPluginFromZip extracts a plugin zip file, validates and
+// installs the plugin
+func installPluginFromZip(ctx context.Context, zipPath string, force bool) error {
```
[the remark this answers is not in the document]
Two-Hearts: In the current code, `zip` and `tar.gz` are sharing the same core process of plugin installation. Since [tar.Reader](https://pkg.go.dev/archive/tar#NewReader) does not implement io/fs.FS, there's going to be many duplicated codes. 
The current core process takes in an `io.Reader`, which is very general and covers both of our cases. So I think we should keep the current way?

cmd/notation/internal/plugin/plugin.go
shizhMSFT: 256 MB should be large enough as it is 10x of a compressed plugin.

Two-Hearts: > 256 MB should be large enough as it is 10x of a compressed plugin.

I was about to propose 64MB, but sure, 256MB sounds good to me as well.

cmd/notation/plugin/install.go
Two-Hearts: Code updated.

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,182 @@
...
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("bad status: %s", resp.Status)
+	}
+	// Write the body to file
+	lr := io.LimitReader(resp.Body, maxPluginSourceBytes)
+	_, err = io.Copy(tmpFile, lr)
+	if err != nil {
+		return err
+	}
+	return nil
```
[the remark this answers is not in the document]
JeyJeyGao: We have checksum validation so we may don't need to check more bytes here.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+	// get plugin metadata
+	pluginMetadata, err := notationplugin.GetPluginMetadata(ctx, pluginName, tmpFilePath)
+	if err != nil {
+		return err
+	}
+	pluginVersion := pluginMetadata.Version
+	// check plugin existence and version
+	var currentPluginVersion string
+	currentPluginMetadata, err := notationplugin.GetPluginMetadataIfExist(ctx, pluginName)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
```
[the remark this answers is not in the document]
Two-Hearts: Yes, it is possible that a non-executable plugin gets installed `manually` by the user. That's why a user should ALWAYS use `notation plugin install` to install a plugin because the command validates the plugin for the user, and only valid plugins can be installed.
In your case, the user needs to `manually` remove the malformed plugin. And always use `notation plugin install` in the future.

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,182 @@
...
+// maxPluginSourceBytes specifies the limit on how many response
+// bytes are allowed in the server's response to the download from URL request
+var maxPluginSourceBytes int64 = 256 * 1024 * 1024 // 256 MiB
+
+// PluginSourceType is an enum for plugin source
+type PluginSourceType int
+
+const (
+	// PluginSourceTypeFile means plugin source is file
+	PluginSourceTypeUnknown PluginSourceType = 1 + iota
```
JeyJeyGao, line 47: Do we need this type?

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+			}
+			return err
+		}
+		if !header.FileInfo().Mode().IsRegular() || strings.Contains(header.Name, "..") {
+			continue
+		}
+		// validate and get plugin name from file name
+		pluginName, err := notationplugin.ExtractPluginNameFromFileName(header.Name)
+		if err != nil {
+			logger.Infoln(err)
```
JeyJeyGao, line 246: It is possible that the tar contains a LICENSE file. In that case, the log will show
> Info: invalid plugin file name. Plugin file name requires format notation-{plugin-name}, but got LICENSE 

It is an expected file so can we update the log to make it clear.

Two-Hearts: There could be various expected files in a tar, for example, `root.cert`, `LICENSE`, and other user customized files. So only checking the `LICENSE` file doesn't seem to be very helpful? That's why I used a `logger.Infoln` to log the error instead of treating it as a real error.

Two-Hearts: Updated the logging message content to be clearer.

cmd/notation/internal/plugin/plugin.go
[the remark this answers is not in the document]
shizhMSFT: We probably need to document the format of `checksum` somewhere.

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	enc := dgst.Encoded()
+	if enc != strings.ToLower(checkSum) {
+		return fmt.Errorf("plugin checksum does not match user input. Expecting %s", checkSum)
+	}
+	return nil
+}
+
+// DownloadPluginFromURL downloads plugin file from url to a tmp directory
+func DownloadPluginFromURL(ctx context.Context, url string, tmpFile *os.File) error {
```
shizhMSFT, line 117: nit: consider `*os.File` to `io.Writer`.

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+		return err
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	// Check server response
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("bad status: %s", resp.Status)
```
shizhMSFT, line 131: This error message is not clear. Could you make it more readable?

cmd/notation/internal/plugin/plugin.go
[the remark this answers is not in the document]
shizhMSFT: We do need to check if there are more bytes. If so, we need to return proper error messages.

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+// getClient returns an *auth.Client
+func getClient(ctx context.Context) *auth.Client {
+	client := &auth.Client{
+		Cache:    auth.NewCache(),
+		ClientID: "notation",
+	}
+	client.SetUserAgent("notation/" + version.GetVersion())
+	setHttpDebugLog(ctx, client)
+	return client
+}
```
shizhMSFT, line 151: Is it possible to reuse 
https://github.com/notaryproject/notation/blob/b75827cc02aecbde8ec7a6b6b13f10c223ff47e6/cmd/notation/registry.go#L146
?

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+		return
+	}
+	if authClient.Client == nil {
+		authClient.Client = http.DefaultClient
+	}
+	if authClient.Client.Transport == nil {
+		authClient.Client.Transport = http.DefaultTransport
+	}
+	authClient.Client.Transport = trace.NewTransport(authClient.Client.Transport)
+}
```
shizhMSFT, line 165: Is it possible to refactor this code as it is duplicated with

https://github.com/notaryproject/notation/blob/b75827cc02aecbde8ec7a6b6b13f10c223ff47e6/cmd/notation/registry.go#L125

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+
+const (
+	notationPluginTmpDir = "notationPluginTmpDir"
+)
+
+type pluginInstallOpts struct {
+	cmd.LoggingFlagOpts
+	pluginSourceType notationplugin.PluginSourceType
+	pluginSource     string
+	inputCheckSum    string
```
shizhMSFT, line 46: Replace all `checkSum` to `checksum` with lower-case `s`. See https://en.wikipedia.org/wiki/Checksum

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+	"github.com/notaryproject/notation-go/dir"
+	"github.com/notaryproject/notation-go/log"
+	notationplugin "github.com/notaryproject/notation/cmd/notation/internal/plugin"
+	"github.com/notaryproject/notation/internal/cmd"
+	"github.com/notaryproject/notation/internal/osutil"
+	"github.com/spf13/cobra"
+)
+
+const (
+	notationPluginTmpDir = "notationPluginTmpDir"
```
shizhMSFT, line 39: nit: The suffix `TmpDir` is not necessary.

shizhMSFT: I think you can have the value `notation-plugin`.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+		Use:     "install [flags] <--file|--url> <plugin_source>",
+		Aliases: []string{"add"},
+		Short:   "Install plugin",
+		Long: `Install a plugin
+
+Example - Install plugin from file system:
+  notation plugin install --file wabbit-plugin-v1.0.zip
+
+Example - Install plugin from file system with user input SHA256 checksum:
+  notation plugin install --file wabbit-plugin-v1.0.zip --checksum abcdef
```
shizhMSFT, line 66: Could you use a real checksum instead of `abcdef`?

shizhMSFT: Otherwise, people may think about short digests.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+  notation plugin install --file wabbit-plugin-v1.0.zip --checksum abcdef 
+
+Example - Install plugin from file system regardless if it's already installed:
+  notation plugin install --file wabbit-plugin-v1.0.zip --force
+
+Example - Install plugin from file system with .tar.gz:
+  notation plugin install --file wabbit-plugin-v1.0.tar.gz
+
+Example - Install plugin from URL, SHA256 checksum is required:
+  notation plugin install --url https://wabbit-networks.com/intaller/linux/amd64/wabbit-plugin-v1.0.tar.gz --checksum abcxyz
```
shizhMSFT, line 75: Same here

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if opts.isFile {
+				opts.pluginSourceType = notationplugin.PluginSourceTypeFile
+				return nil
+			}
+			if opts.isUrl {
+				opts.pluginSourceType = notationplugin.PluginSourceTypeURL
+				return nil
+			}
+			return errors.New("must choose one and only one flag from [--file, --url]")
```
shizhMSFT, line 93: nit: You may want to reword it for a better prompt.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+			return errors.New("install from URL requires non-empty SHA256 checksum of the plugin source")
+		}
+		url, err := url.Parse(opts.pluginSource)
+		if err != nil {
+			return fmt.Errorf("failed to install from URL: %v", err)
+		}
+		if url.Scheme != "https" {
+			return fmt.Errorf("failed to install from URL: %q scheme is not HTTPS", opts.pluginSource)
+		}
+		tmpFile, err := os.CreateTemp(".", "notationPluginDownloadTmp")
```
shizhMSFT, line 126: Should `notationPluginDownloadTmp` be constant?

shizhMSFT, line 126: Should we create temp in `/tmp` or local folder? I guess it should be `/tmp` in the most case.

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+// and gets plugin name from it based on spec: https://github.com/notaryproject/specifications/blob/main/specs/plugin-extensibility.md#installation
+func ExtractPluginNameFromFileName(fileName string) (string, error) {
+	fname := osutil.FileNameWithoutExtension(fileName)
+	if !strings.HasPrefix(fname, proto.Prefix) {
+		return "", notationerrors.ErrorInvalidPluginFileName{Msg: fmt.Sprintf("invalid plugin file name. Plugin file name requires format notation-{plugin-name}, but got %s", fname)}
+	}
+	_, pluginName, found := strings.Cut(fname, "-")
+	if !found {
+		return "", notationerrors.ErrorInvalidPluginFileName{Msg: fmt.Sprintf("invalid plugin file name. Plugin file name requires format notation-{plugin-name}, but got %s", fname)}
+	}
```
shizhMSFT, line 177: Since prefix always has `-` (see https://github.com/notaryproject/notation-go/blob/1bc5a3f8c4d834440deeb3dc0f6855403a6d730d/plugin/proto/proto.go#L19), we don't need to check `found` as it is always `true`.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+	ctx := opts.LoggingFlagOpts.InitializeLogger(command.Context())
+	// core process
+	switch opts.pluginSourceType {
+	case notationplugin.PluginSourceTypeFile:
+		return installFromFileSystem(ctx, opts.pluginSource, opts.inputCheckSum, opts.force)
+	case notationplugin.PluginSourceTypeURL:
+		if opts.inputCheckSum == "" {
+			return errors.New("install from URL requires non-empty SHA256 checksum of the plugin source")
+		}
+		url, err := url.Parse(opts.pluginSource)
```
shizhMSFT, line 119: It is not suggested to use `url` as the variable name as some linter will give a warning saying it overrides the package name.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to install from URL: %v", err)
+		}
+		if url.Scheme != "https" {
```
shizhMSFT, line 123: case insensitive comparison?

cmd/notation/plugin/install.go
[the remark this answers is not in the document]
shizhMSFT: Using `fs.FS` will make the logic more readable and it opens possibilities in the future like installing plugins from a folder, or `embed.FS`, `fstest.MapFS` for unit testing, etc.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+
+// installPluginExecutable extracts, validates, and installs a plugin file
+func installPluginExecutable(ctx context.Context, fileName string, pluginName string, fileReader io.Reader, force bool) error {
+	// sanity check
+	if runtime.GOOS == "windows" && filepath.Ext(fileName) != ".exe" {
+		return fmt.Errorf("on Windows, plugin executable file %s is missing the '.exe' extension", fileName)
+	}
+	if runtime.GOOS != "windows" && filepath.Ext(fileName) == ".exe" {
+		return fmt.Errorf("on %s, plugin executable file %s cannot have the '.exe' extension", runtime.GOOS, fileName)
+	}
```
shizhMSFT, line 262: nit: should you use [build constraints](https://pkg.go.dev/cmd/go#hdr-Build_constraints) for those checks?

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+		return fmt.Errorf("failed to create notationPluginTmpDir: %w", err)
+	}
+	defer os.RemoveAll(tmpDir)
+	tmpFilePath := filepath.Join(tmpDir, fileName)
+	pluginFile, err := os.OpenFile(tmpFilePath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0700)
+	if err != nil {
+		return err
+	}
+	if _, err := io.Copy(pluginFile, fileReader); err != nil {
+		defer pluginFile.Close()
```
shizhMSFT, line 275: nit:
```suggestion
		_ = pluginFile.Close()
```

internal/osutil/file.go
```diff
@@ -94,3 +96,27 @@ func IsRegularFile(path string) (bool, error) {
...
+
+// DetectFileType returns a file's content type given path
+func DetectFileType(path string) (string, error) {
+	rc, err := os.Open(path)
+	if err != nil {
+		return "", err
+	}
+	defer rc.Close()
+	var header [512]byte
+	_, err = io.ReadFull(rc, header[:])
```
shizhMSFT, line 108: It is better to have a `LimitedReader` instead of a 512-byte array as a file may be smaller than 512 bytes

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+		return err
+	}
+	pluginVersion := pluginMetadata.Version
+	// check plugin existence and version
+	var currentPluginVersion string
+	currentPluginMetadata, err := notationplugin.GetPluginMetadataIfExist(ctx, pluginName)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return err
+	}
+	if err == nil { // plugin already exists
```
shizhMSFT, line 293: nit: it is more readable and less error prone.
```suggestion
	if err != nil {
		if !errors.Is(err, os.ErrNotExist) {
			return err
		}
	} else { // plugin already exists
```

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+			comp, err := notationplugin.ComparePluginVersion(pluginVersion, currentPluginVersion)
+			if err != nil {
+				return err
+			}
+			if comp < 0 {
+				return fmt.Errorf("%s. The installing version %s is lower than the existing plugin version %s.\nIt is not recommended to install an older version. To force the installation, use the \"--force\" option", pluginName, pluginVersion, currentPluginVersion)
+			}
+			if comp == 0 {
+				return fmt.Errorf("%s with version %s already exists", pluginName, currentPluginVersion)
+			}
```
shizhMSFT, line 305: Suggest using a golang-style `switch` case.

shizhMSFT, line 305: Should we return error on this case?

cmd/notation/internal/plugin/plugin.go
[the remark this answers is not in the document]
Two-Hearts: I agree, added examples in this command's long message so that user would see them when using `--help` option.

priteshbandi (commented):
Are we missing functionality to copy/install plugin executable not zip or tar.gz ?

cmd/notation/internal/plugin/plugin.go
Two-Hearts: `getAuthClient` is an unexported function in the `main` package, whereas `internal/plugin/plugin.go` is under pakcage `plugin`. In addition, `getAuthClient` is registry specific. 
Therefore, instead of refactoring the `main` package, I think creating plugin's own client might be a more feasible way here.

cmd/notation/internal/plugin/plugin.go
Two-Hearts: Yeah, `setHttpDebugLog` should really belong to the `github.com/notaryproject/notation/internal/trace` package. Moved it there and reusing it in `main` and `plugin` packages.

cmd/notation/plugin/install.go
JeyJeyGao: `/tmp` may not good for Windows? How about keep it empty.

cmd/notation/plugin/install.go
Two-Hearts: Oh yeah, and the flag name needs to be updated as well (it's called `sha256sum` now).

cmd/notation/plugin/install.go
Two-Hearts: > Can we use [`MarkFlagsMutuallyExclusive`](https://github.com/spf13/cobra/issues/1216) ?

@priteshbandi yes, I have done that. The prompt here is for none of the above flags is set.

Two-Hearts: > nit: You may want to reword it for a better prompt.

In Cobra's new version (v1.8.0), there's the new [MarkFlagsOneRequired](https://pkg.go.dev/github.com/spf13/cobra@v1.8.0#Command.MarkFlagsOneRequired). Updated to use it along with `MarkFlagsMutuallyExclusive`.

cmd/notation/plugin/install.go
Two-Hearts: Keeping as empty so that the os default tmp folder is used.

cmd/notation/plugin/install.go
Two-Hearts: The `url.Parse` function always returns `url.Scheme` in lower case: https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/net/url/url.go;l=522

cmd/notation/plugin/install.go
Two-Hearts: Yes, it is returning an error. This was discussed in the spec PR: https://github.com/notaryproject/notation/pull/809#discussion_r1396788077 suggested by Pritesh.

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+	mgr := plugin.NewCLIManager(dir.PluginFS())
+	plugin, err := mgr.Get(ctx, pluginName)
+	if err != nil {
+		return nil, err
+	}
+	return plugin.GetMetadata(ctx, &proto.GetMetadataRequest{})
+}
+
+// GetPluginMetadata returns plugin's metadata given plugin path
+func GetPluginMetadata(ctx context.Context, pluginName, path string) (*proto.GetMetadataResponse, error) {
```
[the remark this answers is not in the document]
Two-Hearts: This is because, `GetPluginMetadataIfExist` only checks if `plugin with name <pluginName>` is already installed in Notation.
However, `GetPluginMetadata` gets metadata of a plugin `candidate` at location `path`. At the same time, it expects the plugin name returned from metadata matches the input `<pluginName>`. If any step failed, `GetPluginMetadata` would return an error. This prevents installing a malformed plugin to Notation.

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return plugin.GetMetadata(ctx, &proto.GetMetadataRequest{})
+}
+
+// ComparePluginVersion validates and compares two plugin semantic versions
+func ComparePluginVersion(v, w string) (int, error) {
+	// semantic version strings must begin with a leading "v"
+	if !strings.HasPrefix(v, "v") {
```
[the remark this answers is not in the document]
Two-Hearts: We are using golang's built in package: [semver](https://pkg.go.dev/golang.org/x/mod/semver), and this package requires: `semantic version strings must begin with a leading "v"`, reference: https://pkg.go.dev/golang.org/x/mod/semver#pkg-overview.
Therefore, if the released version does not have the prefix "v", it is added just for comparing purpose. Note: the plugin's version is not changed, i.e., if it is `1.0.0`, this function won't update it to `v1.0.0`.

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+		v = "v" + v
+	}
+	if !semver.IsValid(v) {
```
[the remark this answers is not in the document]
Two-Hearts: Same as above, we are using golang's built in library here.

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+	rc, err := os.Open(path)
+	if err != nil {
+		return err
+	}
+	defer rc.Close()
+	dgst, err := digest.FromReader(rc)
+	if err != nil {
+		return err
+	}
+	enc := dgst.Encoded()
```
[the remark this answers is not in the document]
Two-Hearts: Good idea, updating the code.

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+		Cache:    auth.NewCache(),
+		ClientID: "notation",
+	}
+	client.SetUserAgent("notation/" + version.GetVersion())
+	setHttpDebugLog(ctx, client)
+	return client
+}
+
+// setHttpDebugLog sets http debug logger
+func setHttpDebugLog(ctx context.Context, authClient *auth.Client) {
```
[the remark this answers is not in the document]
Two-Hearts: This is a suggestion from @shizhMSFT. The main purpose here is to reuse the code we already have. I will need to refactor `getAuthClient` under `main` package so that it can be reused in other packages. I also moved `setHttpDebugLog` into the `trace` package since it's not really registry specific.

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+		authClient.Client.Transport = http.DefaultTransport
+	}
+	authClient.Client.Transport = trace.NewTransport(authClient.Client.Transport)
+}
+
+// ExtractPluginNameFromFileName checks if fileName is a valid plugin file name
+// and gets plugin name from it based on spec: https://github.com/notaryproject/specifications/blob/main/specs/plugin-extensibility.md#installation
+func ExtractPluginNameFromFileName(fileName string) (string, error) {
+	fname := osutil.FileNameWithoutExtension(fileName)
+	if !strings.HasPrefix(fname, proto.Prefix) {
```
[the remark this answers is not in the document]
Two-Hearts: Nice suggestion! I could see this is added in go 1.20. Updating the code to use it.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+			return errors.New("must choose one and only one flag from [--file, --url]")
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return installPlugin(cmd, opts)
+		},
+	}
+	opts.LoggingFlagOpts.ApplyFlags(command.Flags())
+	command.Flags().BoolVar(&opts.isFile, "file", false, "install plugin from a file in file system")
+	command.Flags().BoolVar(&opts.isUrl, "url", false, "install plugin from an HTTPS URL")
+	command.Flags().StringVar(&opts.inputCheckSum, "sha256sum", "", "must match SHA256 of the plugin source")
```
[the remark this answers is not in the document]
Two-Hearts: hmm, this is not really the case. It can also be used with `--file` flag, it's optional though.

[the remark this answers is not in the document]
Two-Hearts: Updated to `must match SHA256 of the plugin source, required when `--url` flag is set`

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+func installPlugin(command *cobra.Command, opts *pluginInstallOpts) error {
+	// set log level
+	ctx := opts.LoggingFlagOpts.InitializeLogger(command.Context())
+	// core process
+	switch opts.pluginSourceType {
+	case notationplugin.PluginSourceTypeFile:
+		return installFromFileSystem(ctx, opts.pluginSource, opts.inputCheckSum, opts.force)
+	case notationplugin.PluginSourceTypeURL:
+		if opts.inputCheckSum == "" {
+			return errors.New("install from URL requires non-empty SHA256 checksum of the plugin source")
```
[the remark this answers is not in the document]
Two-Hearts: There's one thing about `PreRunE`, it is evaluated before the check of `ValidateFlagGroups`, i.e. before the flags check set by `MarkFlagsMutuallyExclusive` and `MarkFlagsOneRequired`. Since `checksum` is required by `--url` and optional by `--file`, checking `checksum` in `PreRunE` complicates the logic comparing to checking it after the validation of the flags.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+		}
+		url, err := url.Parse(opts.pluginSource)
+		if err != nil {
+			return fmt.Errorf("failed to install from URL: %v", err)
+		}
+		if url.Scheme != "https" {
+			return fmt.Errorf("failed to install from URL: %q scheme is not HTTPS", opts.pluginSource)
```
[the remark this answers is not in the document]
Two-Hearts: Updated to:
`Error: the plugin download failed: only the HTTPS scheme is supported, but got <pluginURL.Scheme>`

cmd/notation/plugin/install.go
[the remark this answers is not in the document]
Two-Hearts: Updated the code.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+		defer tmpFile.Close()
+		err = notationplugin.DownloadPluginFromURL(ctx, opts.pluginSource, tmpFile)
+		if err != nil {
+			return fmt.Errorf("failed to download plugin from URL %s with error: %w", opts.pluginSource, err)
+		}
+		downloadPath, err := filepath.Abs(tmpFile.Name())
+		if err != nil {
+			return err
+		}
+		return installFromFileSystem(ctx, downloadPath, opts.inputCheckSum, opts.force)
```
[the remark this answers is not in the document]
Two-Hearts: It's been done in the `ValidateChecksum` function. We always compare lower case checksums.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+	default:
+		return errors.New("failed to install the plugin: plugin source type is unknown")
+	}
+}
+
+// installFromFileSystem install the plugin from file system
+func installFromFileSystem(ctx context.Context, inputPath string, inputCheckSum string, force bool) error {
```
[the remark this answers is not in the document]
Two-Hearts: The name `installPlugin` is already been taken. Since `installFromFileSystem` is an unexported function, I think we can keep the name?

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+	logger := log.GetLogger(ctx)
+	archive, err := zip.OpenReader(zipPath)
+	if err != nil {
+		return err
+	}
+	defer archive.Close()
+	// require one and only one file with name in the format
+	// `notation-{plugin-name}`
+	for _, f := range archive.File {
+		if !f.Mode().IsRegular() || strings.Contains(f.Name, "..") {
```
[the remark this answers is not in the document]
Two-Hearts: This is required by `github-advanced-security`: https://github.com/notaryproject/notation/security/code-scanning/44

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+func installPluginExecutable(ctx context.Context, fileName string, pluginName string, fileReader io.Reader, force bool) error {
+	// sanity check
+	if runtime.GOOS == "windows" && filepath.Ext(fileName) != ".exe" {
+		return fmt.Errorf("on Windows, plugin executable file %s is missing the '.exe' extension", fileName)
+	}
+	if runtime.GOOS != "windows" && filepath.Ext(fileName) == ".exe" {
+		return fmt.Errorf("on %s, plugin executable file %s cannot have the '.exe' extension", runtime.GOOS, fileName)
+	}
+	// extract to tmp dir
+	tmpDir, err := os.MkdirTemp(".", notationPluginTmpDir)
```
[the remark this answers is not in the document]
Two-Hearts: Code updated.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+	pluginFile, err := os.OpenFile(tmpFilePath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0700)
+	if err != nil {
+		return err
+	}
+	if _, err := io.Copy(pluginFile, fileReader); err != nil {
+		defer pluginFile.Close()
+		return err
+	}
+	if err := pluginFile.Close(); err != nil {
+		return err
```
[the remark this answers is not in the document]
Two-Hearts: The previous close happens when there's an error during `io.Copy`, if we do not close `pluginFile` there, `defer os.RemoveAll(tmpDir)` would fail silently.
The reason we need to manually close `pluginFile` here is that the file is used again in `GetPluginMetadata`, we need to close it before using it again.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+	// get plugin metadata
+	pluginMetadata, err := notationplugin.GetPluginMetadata(ctx, pluginName, tmpFilePath)
+	if err != nil {
+		return err
+	}
+	pluginVersion := pluginMetadata.Version
+	// check plugin existence and version
+	var currentPluginVersion string
+	currentPluginMetadata, err := notationplugin.GetPluginMetadataIfExist(ctx, pluginName)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
```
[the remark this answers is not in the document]
Two-Hearts: If `force==true` AND there's already an existing plugin, we still need to get the current plugin's version from its metadata, because according to our spec, on success, the final message is:
`Succussefully installed plugin <plugin-name>, updated the version from <current-version> to <new-version>`.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+	if err == nil { // plugin already exists
+		currentPluginVersion = currentPluginMetadata.Version
+		if !force {
+			comp, err := notationplugin.ComparePluginVersion(pluginVersion, currentPluginVersion)
+			if err != nil {
+				return err
+			}
+			if comp < 0 {
+				return fmt.Errorf("%s. The installing version %s is lower than the existing plugin version %s.\nIt is not recommended to install an older version. To force the installation, use the \"--force\" option", pluginName, pluginVersion, currentPluginVersion)
```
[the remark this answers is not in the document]
Two-Hearts: The current final error message displayed to user is:
`Error: plugin installation failed: <plugin-name>. The installing version <new-version> is lower than the existing plugin version <current-version>.`
`It is not recommended to install an older version. To force the installation, use the \"--force\" option`
It's been discussed in the spec PR: https://github.com/notaryproject/notation/pull/809.
@FeynmanZhou

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+	if err != nil {
+		return err
+	}
+	_, err = osutil.CopyToDir(tmpFilePath, pluginPath)
+	if err != nil {
+		return err
+	}
+	// plugin is always executable
+	pluginFilePath := filepath.Join(pluginPath, fileName)
+	err = os.Chmod(pluginFilePath, 0700)
```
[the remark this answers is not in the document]
Two-Hearts: We are reusing the function `osutil.CopyToDir` to move the plugin binary file into Notation, in this process, we are giving the file with permission `0600`: https://github.com/notaryproject/notation/blob/4569f3f2cfe723b38ea7d7a26ac007281b6f0a34/internal/osutil/file.go#L81. 
Therefore, after the move, the plugin binary file is not executable, but since we have already validated the plugin before the move, we are safe to set its permission back to `0700` in this case.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,329 @@
...
+	if err != nil {
+		return err
+	}
+	if currentPluginVersion != "" {
+		fmt.Printf("Succussefully installed plugin %s, updated the version from %s to %s\n", pluginName, currentPluginVersion, pluginVersion)
```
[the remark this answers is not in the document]
Two-Hearts: I think the original message is better, because user used `notation plugin install` command, the message should begin with `Successfully installed...`. @FeynmanZhou

cmd/notation/internal/plugin/plugin.go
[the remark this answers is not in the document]
Two-Hearts: Code updated.

cmd/notation/plugin/install.go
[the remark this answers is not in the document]
Two-Hearts: Update the code.

cmd/notation/internal/plugin/plugin.go
Two-Hearts: Updated the code.

cmd/notation/plugin/install.go
Two-Hearts: Refactored this part to notation-go.

cmd/notation/internal/plugin/plugin.go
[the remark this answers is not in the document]
Two-Hearts: Updated the code.

cmd/notation/internal/plugin/plugin.go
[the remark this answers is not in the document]
Two-Hearts: Refactored the code to notation-go.

cmd/notation/internal/plugin/plugin.go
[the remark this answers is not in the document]
Two-Hearts: Refactored the code to notation-go.

cmd/notation/plugin/install.go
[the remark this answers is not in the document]
Two-Hearts: Updated the code.

Two-Hearts:
> Are we missing functionality to copy/install plugin executable not zip or tar.gz ?

@priteshbandi This is a good question, it is tracked by this issue created by @FeynmanZhou : https://github.com/notaryproject/specifications/issues/284. Basically, we need to be clear in the `notaryproject spec` on how should a vendor release their plugins in what kind of format. 

Therefore, in this iteration, I think we could just support zip/tar.gz as a first step. We can add the support for executable in the future once we have a decision on the spec.

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,80 @@
...
+	lr := &io.LimitedReader{
+		R: resp.Body,
+		N: MaxPluginSourceBytes,
+	}
+	_, err = io.Copy(tmpFile, lr)
+	if err != nil {
+		return err
+	}
+	if lr.N == 0 {
+		return fmt.Errorf("https response reaches the %d MiB size limit", MaxPluginSourceBytes)
```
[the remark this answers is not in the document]
Two-Hearts: By design, yes. The error message states that `the response reaches the size limit`.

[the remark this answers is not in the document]
JeyJeyGao: Got it. Then can we update the description of `MaxPluginSourceBytes` to mention that the file size should be strictly less than the max value (not equal).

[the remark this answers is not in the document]
Two-Hearts: Updated.

JeyJeyGao (approved):
LGTM

priteshbandi:
> > Are we missing functionality to copy/install plugin executable not zip or tar.gz ?
> 
> @priteshbandi This is a good question, it is tracked by this issue created by @FeynmanZhou : [notaryproject/specifications#284](https://github.com/notaryproject/specifications/issues/284). Basically, we need to be clear in the `notaryproject spec` on how should a vendor release their plugins in what kind of format.
> 
> Therefore, in this iteration, I think we could just support zip/tar.gz as a first step. We can add the support for executable in the future once we have a decision on the spec.

I mean the functionality to copy plugin from a location on disk. This should be easy and straightforward.

FeynmanZhou:
> I mean the functionality to copy plugin from a location on disk. This should be easy and straightforward.
Hi @priteshbandi ,

Just want to clarify it: Do you mean support installing a plugin from an executable file (like `.msi`, `.exe`, `.pkg`)?

If so, this has not been included in the [plugin management spec](https://github.com/notaryproject/notation/pull/809/files#diff-7637fbf27d1c8c25ef1a9806613d4aa17aaa61d99aae29f8c802b5e50bcd2efdR83). Notation v1.1.0 will only support `.zip` and `.tar.gz` format of the plugin file as we aligned in the CLI spec. 

IMO, if there is an executable plugin file, users should be able to install the plugin without `notation plugin install`. But in that case it is not secured by Notation plugin system.

Have you seen any plugin vendor or customer needs relying on executable file format? How about we put the executable file format in the [notation plugin spec](https://github.com/notaryproject/specifications/blob/main/specs/plugin-extensibility.md) as its roadmap first and considering it until received vendor/customer needs on this format?

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,326 @@
...
+	defer os.RemoveAll(tmpDir)
+	tmpFilePath := filepath.Join(tmpDir, fileName)
+	pluginFile, err := os.OpenFile(tmpFilePath, os.O_WRONLY|os.O_CREATE, 0700)
+	if err != nil {
+		return err
+	}
+	lr := &io.LimitedReader{
+		R: fileReader,
+		N: notationplugin.MaxPluginSourceBytes,
+	}
```
priteshbandi, line 300: Are we saying the size of plugin executable cannot be more than 256 MB (compressed or uncompressed doesnt matters)?

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,83 @@
...
+	"io"
+	"net/http"
+
+	notationauth "github.com/notaryproject/notation/internal/auth"
+)
+
+// MaxPluginSourceBytes specifies the limit on how many bytes are allowed in the
+// server's response to the download from URL request.
+//
+// It also specifies the limit of a potentail plugin executable file in a
```
priteshbandi, line 28:
```suggestion
// It also specifies the maximum size of a plugin executable file in a
```

cmd/notation/plugin/install.go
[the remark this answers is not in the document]
FeynmanZhou: Hi @priteshbandi ,

I think it is worth reminding users that an older version is not recommended when users try to downgrade. The reason is that this behavior may introduce security risk because an older version may include unfixed security issue or may not be supported by the Notary Project community/vendor.

cmd/notation/plugin/install.go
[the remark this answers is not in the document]
Two-Hearts: Updated the logic.

Two-Hearts:
> yes, install command would just need to copy file from user provided location to notation's location. It is part of the spec with following flag ` --file install plugin from a file in file system`

@priteshbandi Added this logic.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,326 @@
...
+	cmd.LoggingFlagOpts
+	pluginSourceType notationplugin.PluginSourceType
+	pluginSource     string
+	inputChecksum    string
+	isFile           bool
+	isUrl            bool
+	force            bool
+}
+
+var ErrNoPluginExecutableFileWasFound = errors.New("no plugin executable file was found")
```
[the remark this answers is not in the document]
Two-Hearts: Removed it, since it's returned from notation-go now.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,326 @@
...
+  notation plugin install --file wabbit-plugin-v1.0.zip --force
+
+Example - Install plugin from file system with .tar.gz:
+  notation plugin install --file wabbit-plugin-v1.0.tar.gz
+
+Example - Install plugin from URL, SHA256 checksum is required:
+  notation plugin install --url https://wabbit-networks.com/intaller/linux/amd64/wabbit-plugin-v1.0.tar.gz --sha256sum f8a75d9234db90069d9eb5660e5374820edf36d710bd063f4ef81e7063d3810b
+`,
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) == 0 {
```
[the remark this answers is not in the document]
Two-Hearts: Added the check in `Install`. (We are not checking it in other Notation commands such as `Sign` and `Verify` though. )

go.mod
```diff
@@ -30,3 +30,5 @@ require (
 	golang.org/x/sync v0.4.0 // indirect
 	golang.org/x/sys v0.14.0 // indirect
 )
+
+replace github.com/notaryproject/notation-go => github.com/Two-Hearts/notation-go v0.0.0-20231206052238-c2f7ef881b94
```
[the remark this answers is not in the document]
Two-Hearts: Dependency is up to date.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,312 @@
...
+				if opts.isFile {
+					return errors.New("missing plugin file path")
+				}
+				if opts.isURL {
+					return errors.New("missing plugin URL")
+				}
```
shizhMSFT, line 86: nit: consider
```suggestion
				switch {
				case opts.isFile:
					return errors.New("missing plugin file path")
				case opts.isURL:
					return errors.New("missing plugin URL")
				}
```
The logic is the same but for readability.

shizhMSFT: Same comment for line 96-100

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,312 @@
...
+	}); err != nil {
+		return err
+	}
+	// install core process
+	installOpts := plugin.CLIInstallOptions{
+		PluginPath: tmpDir,
+		Overwrite:  force,
+	}
+	return installPluginWithOptions(ctx, installOpts)
+}
```
shizhMSFT, line 244: This part can be refactored into `notation-go` by having a new field like `FS fs.FS` in `CLIInstallOptions`.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,312 @@
...
+			Overwrite:  force,
+		}
+		return installPluginWithOptions(ctx, installOpts)
+	}
+}
+
+// installPluginFromFS extracts, validates and installs the plugin files
+// from a fs.FS
+//
+// Note: zip.ReadCloser implments fs.FS
```
shizhMSFT, line 199: info: just FYI that tar can also be mapped to `fs.FS`: https://github.com/oras-project/oras-go/blob/main/internal/fs/tarfs/tarfs.go but gzipped tar cannot.

cmd/notation/plugin/install.go
Two-Hearts: @shizhMSFT How about we move it to notation-go in the next iteration? (It won't be a breaking change.)

shizhMSFT: I think you can start a new PR in `notation-go`. If that PR merges first, we can update that in this PR. Otherwise, we can do it after merging this PR.

specs/commandline/plugin.md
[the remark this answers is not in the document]
Two-Hearts: Oops, nice catch! My IDE did not pop up an alert to resolve conflicts in this file.

internal/trace/transport.go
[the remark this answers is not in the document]
Two-Hearts: Create new issue to tack this: https://github.com/notaryproject/notation/issues/851

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,81 @@
...
+	MediaTypeZip = "application/zip"
+
+	// MediaTypeGzip means plugin file is gzip
+	MediaTypeGzip = "application/x-gzip"
+)
+
+// DownloadPluginFromURL downloads plugin file from url to a tmp directory
+func DownloadPluginFromURL(ctx context.Context, pluginURL string, tmpFile io.Writer) error {
+	// Get the data
+	client := httputil.NewAuthClient(ctx, &http.Client{Timeout: 100 * time.Second})
```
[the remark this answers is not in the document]
priteshbandi: Also if we are doing 100 seconds timeout; we need to communicate this to user otherwise user might think the command is stuck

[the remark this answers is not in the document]
Two-Hearts: @shizhMSFT The purpose of having a timeout is to avoid scripts and CI/CD pipelines hang at this step indefinitely. Golang provides a timeout with 0 as default value, I found the `100 seconds` from here: https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.timeout?view=net-8.0#remarks.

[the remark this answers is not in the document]
Two-Hearts: > Also if we are doing 100 seconds timeout; we need to communicate this to user otherwise user might think the command is stuck

@priteshbandi The default value is 0 which means no timeout. Adding this value actually prevent user from stuck.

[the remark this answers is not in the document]
shizhMSFT: Can you make it 20 minutes and document it somewhere?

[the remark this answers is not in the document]
Two-Hearts: Updated to 20 minutes, added the info in the command description.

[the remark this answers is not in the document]
priteshbandi: For all practical purpose waiting more than 2 mins without showing progress is basically stuck.

cmd/notation/plugin/install.go
Two-Hearts: @shizhMSFT Could you approve this PR first as the 1st iteration of the feature? Making the above changes is not a quick PR, it's actually the next iteration of the command. The update is in progress, created this issue to track: https://github.com/notaryproject/notation/issues/852

cmd/notation/internal/plugin/plugin.go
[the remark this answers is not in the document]
Two-Hearts: > For all practical purpose waiting more than 2 mins without showing progress means that process is stuck.

@priteshbandi I had a discussion with @shizhMSFT, updated the timeout to 10 minutes. Stated the value in the description of the `--url` flag. If user has a very slow network connection, their workaround is to download the file then install from file system.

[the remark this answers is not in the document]
priteshbandi: My concern still persist. Can we tell customers that we are downloading plugin, probably using progress bar(if possible) ?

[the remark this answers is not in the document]
Two-Hearts: > My concern still exists. Can we tell customers that we are downloading plugin, probably using progress bar(if possible) ?

@priteshbandi Adding progress bar is considered as a new feature which is not in this PR's scope. I added printouts before and after the download so that user could know what is happening.

cmd/notation/internal/plugin/plugin.go
```diff
@@ -51,7 +51,7 @@ const (
 // DownloadPluginFromURL downloads plugin file from url to a tmp directory
 func DownloadPluginFromURL(ctx context.Context, pluginURL string, tmpFile io.Writer) error {
 	// Get the data
-	client := httputil.NewAuthClient(ctx, &http.Client{Timeout: 100 * time.Second})
+	client := httputil.NewAuthClient(ctx, &http.Client{Timeout: 10 * time.Minute})
```
shizhMSFT, line 54: Can you make the timeout a constant variable at the top of the file?

cmd/notation/plugin/install.go
```diff
@@ -138,10 +139,12 @@ func install(command *cobra.Command, opts *pluginInstallOpts) error {
 		}
 		defer os.Remove(tmpFile.Name())
 		defer tmpFile.Close()
+		fmt.Printf("Downloading plugin from %s\n", opts.pluginSource)
 		err = notationplugin.DownloadPluginFromURL(ctx, opts.pluginSource, tmpFile)
 		if err != nil {
 			return fmt.Errorf("failed to download plugin from URL %s with error: %w", opts.pluginSource, err)
 		}
+		fmt.Printf("Download completed. Saved to tmp file %s\n", tmpFile.Name())
```
shizhMSFT, line 147:
```suggestion
		fmt.Printf("Download completed. Saved to temporary file %s\n", tmpFile.Name())
```

shizhMSFT: Do we really need to show the path of the temporary file as it will be deleted anyway?

Two-Hearts: Removed.

shizhMSFT (approved):
LGTM

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,85 @@
...
+	lr := &io.LimitedReader{
+		R: resp.Body,
+		N: MaxPluginSourceBytes,
+	}
+	_, err = io.Copy(tmpFile, lr)
+	if err != nil {
+		return err
+	}
+	if lr.N == 0 {
+		return fmt.Errorf("%s %q: https response reaches the %d MiB size limit", resp.Request.Method, resp.Request.URL, MaxPluginSourceBytes)
```
priteshbandi, line 82: nit: plugin size %s MB is more than max supported size of 256 MB.

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,315 @@
...
+				switch {
+				case opts.isFile:
+					return errors.New("missing plugin file path")
+				case opts.isURL:
+					return errors.New("missing plugin URL")
+				}
+				return errors.New("missing plugin source")
+			}
+			if len(args) > 1 {
+				return fmt.Errorf("can only insall one plugin at a time, but got %v", args)
```
priteshbandi, line 90:
```suggestion
				return fmt.Errorf("You can install only one plugin at a time, but got %v", args)
```

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,315 @@
...
+`,
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) == 0 {
+				switch {
+				case opts.isFile:
+					return errors.New("missing plugin file path")
+				case opts.isURL:
+					return errors.New("missing plugin URL")
+				}
+				return errors.New("missing plugin source")
```
priteshbandi, line 87:
```suggestion
				return errors.New("missing plugin location")
```

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,315 @@
...
+			case opts.isFile:
+				opts.pluginSourceType = notationplugin.PluginSourceTypeFile
+			case opts.isURL:
+				opts.pluginSourceType = notationplugin.PluginSourceTypeURL
+			}
+			return install(cmd, opts)
+		},
+	}
+	opts.LoggingFlagOpts.ApplyFlags(command.Flags())
+	command.Flags().BoolVar(&opts.isFile, "file", false, "install plugin from a file in file system")
```
priteshbandi, line 106:
```suggestion
	command.Flags().BoolVar(&opts.isFile, "file", false, "install plugin from a file on file system")
```

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,315 @@
...
+	command.Flags().BoolVar(&opts.isURL, "url", false, "install plugin from an HTTPS URL. The timeout of the download HTTPS request is set to 10 minutes")
```
priteshbandi, line 107:
```suggestion
	command.Flags().BoolVar(&opts.isURL, "url", false, "install plugin from an HTTPS URL. The default plugin download timeout is 10 minutes")
```

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,315 @@
...
+	// core process
+	switch opts.pluginSourceType {
+	case notationplugin.PluginSourceTypeFile:
+		if err := installPlugin(ctx, opts.pluginSource, opts.inputChecksum, opts.force); err != nil {
+			return fmt.Errorf("plugin installation failed: %w", err)
+		}
+		return nil
+	case notationplugin.PluginSourceTypeURL:
+		if opts.inputChecksum == "" {
+			return errors.New("installing from URL requires non-empty SHA256 checksum of the plugin source")
```
priteshbandi, line 127:
```suggestion
			return errors.New("installing from URL requires non-empty SHA256 checksum of the plugin archive")
```

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+		pluginURL, err := url.Parse(opts.pluginSource)
+		if err != nil {
+			return fmt.Errorf("failed to parse plugin download URL %s with error: %w", pluginURL, err)
```
priteshbandi, line 131:
```suggestion
			return fmt.Errorf("plugin download from %s failed with error: %w", pluginURL, err)
```

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+		if pluginURL.Scheme != "https" {
+			return fmt.Errorf("failed to download plugin from URL: only the HTTPS scheme is supported, but got %s", pluginURL.Scheme)
+		}
+		tmpFile, err := os.CreateTemp("", notationPluginDownloadTmpFile)
+		if err != nil {
+			return fmt.Errorf("failed to create notationPluginDownloadTmpFile: %w", err)
```
priteshbandi, line 138:
```suggestion
			return fmt.Errorf("failed to create temporary directory required for downloading plugin: %w", err)
```

internal/osutil/file.go
```diff
@@ -94,3 +102,62 @@ func IsRegularFile(path string) (bool, error) {
...
+	}
+	defer rc.Close()
+	sha256Hash := sha256.New()
+	if _, err := io.Copy(sha256Hash, rc); err != nil {
+		return err
+	}
+	sha256sum := sha256Hash.Sum(nil)
+	enc := hex.EncodeToString(sha256sum[:])
+	if !strings.EqualFold(enc, checksum) {
+		return fmt.Errorf("plugin checksum does not match user input. Expecting %s", checksum)
```
priteshbandi, line 160:
```suggestion
		return fmt.Errorf("plugin sha256sum %s does not match user input. Expecting %s", enc, checksum)
```

internal/osutil/file.go
```diff
@@ -94,3 +102,62 @@ func IsRegularFile(path string) (bool, error) {
...
+	lr := &io.LimitedReader{
+		R: src,
+		N: MaxFileBytes,
+	}
+	if _, err := io.Copy(dstFile, lr); err != nil || lr.N == 0 {
+		_ = dstFile.Close()
+		if err != nil {
+			return err
+		}
+		return fmt.Errorf("file reaches the %d MiB size limit", MaxFileBytes)
```
priteshbandi, line 122:
```suggestion
		return fmt.Errorf("file reached the %d MiB size limit", MaxFileBytes)
```

cmd/notation/plugin/install.go
```diff
@@ -0,0 +1,315 @@
...
+			return err
+		}
+	}
+	// install the plugin based on file type
+	fileType, err := osutil.DetectFileType(inputPath)
+	if err != nil {
+		return err
+	}
+	switch fileType {
+	case notationplugin.MediaTypeZip:
```
priteshbandi, line 179: Can we please add test for [zipbomb](https://en.wikipedia.org/wiki/Zip_bomb), [zipslip](https://security.snyk.io/research/zip-slip-vulnerability), ZIP filename spoofing ?

priteshbandi (approved):
LGTM.

We should track all the open comments in an issue and address them before 1.1 release

cmd/notation/internal/plugin/plugin.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+// DownloadPluginFromURLTimeout is the timeout when downloading plugin from a
+// URL
+const DownloadPluginFromURLTimeout = 10 * time.Minute
+
+// DownloadPluginFromURL downloads plugin file from url to a tmp directory
+func DownloadPluginFromURL(ctx context.Context, pluginURL string, tmpFile io.Writer) error {
+	// Get the data
+	client := httputil.NewAuthClient(ctx, &http.Client{Timeout: DownloadPluginFromURLTimeout})
```
priteshbandi, line 58: Please open an issue for this: We should show download progress to users.,

Two-Hearts:
@priteshbandi As discussed, issue created: https://github.com/notaryproject/notation/issues/853

cmd/notation/plugin/install.go
Two-Hearts: This is not always true, because we also support install from a single executable file (which is not in an archive).

cmd/notation/plugin/install.go
Two-Hearts: One thing to point out is that according to notation's threat model: https://github.com/notaryproject/specifications/blob/main/threatmodels/notation-threatmodel.md

malicious plugin is mitigated by `Verify plugin package before installation or plugin is installed by authorized users`.

PR #857, opened 2024-01-03
fix: fix bug in `SetHTTPDebugLog`

### LINKED ISSUE #851: Fix of setHttpDebugLog function

Creating an issue to track these comments:
https://github.com/notaryproject/notation/pull/827#discussion_r1430982227
https://github.com/notaryproject/notation/pull/827#discussion_r1430983979

Quote from @shizhMSFT: `When authClient.Client is nil, http.DefaultClient.Transport is changed to http.DefaultTransport if it is nil. Later, http.DefaultClient.Transport is changed to NewTransport. That's a global change and should be avoided.`

### How can we reproduce it?

nil

### Describe your environment

all environments

current main branch

Two-Hearts:
This PR fixes the `SetHTTPDebugLog` function.

Resolves #851.

### DIFF
--- a/internal/trace/transport.go
+++ b/internal/trace/transport.go
@@ -88,14 +88,14 @@ func logHeader(header http.Header, e log.Logger) {
 
 // SetHTTPDebugLog sets up http debug log with logrus.Logger
 func SetHTTPDebugLog(ctx context.Context, authClient *auth.Client) {
-	if logrusLog, ok := log.GetLogger(ctx).(*logrus.Logger); ok && logrusLog.Level != logrus.DebugLevel {
+	if logrusLog, ok := log.GetLogger(ctx).(*logrus.Logger); !ok || logrusLog.Level != logrus.DebugLevel {
 		return
 	}
 	if authClient.Client == nil {
-		authClient.Client = http.DefaultClient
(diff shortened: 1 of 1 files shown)


internal/trace/transport.go
```diff
@@ -88,14 +88,14 @@ func logHeader(header http.Header, e log.Logger) {
...
+	if logrusLog, ok := log.GetLogger(ctx).(*logrus.Logger); !ok || logrusLog.Level != logrus.DebugLevel {
 		return
 	}
 	if authClient.Client == nil {
-		authClient.Client = http.DefaultClient
+		authClient.Client = &http.Client{}
 	}
 	if authClient.Client.Transport == nil {
-		authClient.Client.Transport = http.DefaultTransport
+		authClient.Client.Transport = http.DefaultTransport.(*http.Transport).Clone()
```
shizhMSFT, line 98: The original one looks good as we are not changing anything in `http.DefaultTransport` (even at line 100). It is not necessary to clone a new transport as long as we have new instance at line 95.
```suggestion
		authClient.Client.Transport = http.DefaultTransport
```

priteshbandi (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #858, opened 2024-01-05
fix: `notation plugin install` fails to install a single plugin executable file

Two-Hearts:
This PR fixes the following bug:
When installing a single plugin executable file from URL, the downloaded file name is not preserved. Since Notation relies on file name and file extension to retrieve the plugin name ([spec](https://github.com/notaryproject/specifications/blob/main/specs/plugin-extensibility.md#installation)) and check whether it's executable on the OS, the downloaded file name has to be preserved.

This PR depends on PR: https://github.com/notaryproject/notation-go/pull/369.

### DIFF
--- a/cmd/notation/internal/plugin/plugin.go
+++ b/cmd/notation/internal/plugin/plugin.go
@@ -57,7 +57,7 @@ const (
 const DownloadPluginFromURLTimeout = 10 * time.Minute
 
 // DownloadPluginFromURL downloads plugin source from url to a tmp dir on file
-// system. On success, it returns the downloaded file.
+// system. On success, it returns the downloaded file path.
 func DownloadPluginFromURL(ctx context.Context, pluginURL, tmpDir string) (string, error) {
 	// Get the data
 	client := httputil.NewAuthClient(ctx, &http.Client{Timeout: DownloadPluginFromURLTimeout})
--- a/cmd/notation/internal/plugin/plugin.go
+++ b/cmd/notation/internal/plugin/plugin.go
@@ -17,7 +17,11 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"mime"
 	"net/http"
+	"os"
+	"path"
+	"path/filepath"
 	"time"
 
 	"github.com/notaryproject/notation/internal/httputil"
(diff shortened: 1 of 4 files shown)


cmd/notation/internal/plugin/plugin.go
```diff
@@ -52,35 +56,54 @@
...
+		if err == nil { // if there's an error, use the filename in URL
+			downloadedFilename = params["filename"]
+		}
+	}
+	if downloadedFilename == "" {
+		downloadedFilename = path.Base(req.URL.Path)
 	}
 	// Write the body to file
+	tmpFilePath := filepath.Join(tmpDir, downloadedFilename)
+	tmpFile, err := os.Create(tmpFilePath)
```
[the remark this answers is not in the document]
Two-Hearts: `path.Base` is used here to get the file name only. If the file path is `/a../../../myFile`, the result would still be `myFile`. The code does not follow `url.Path`.

cmd/notation/internal/plugin/plugin.go
```diff
@@ -52,35 +56,54 @@ const (
 // URL
 const DownloadPluginFromURLTimeout = 10 * time.Minute
 
-// DownloadPluginFromURL downloads plugin source from url to a tmp file on file
-// system
-func DownloadPluginFromURL(ctx context.Context, pluginURL string, tmpFile io.Writer) error {
+// DownloadPluginFromURL downloads plugin source from url to a tmp dir on file
+// system. On success, it returns the downloaded file.
+func DownloadPluginFromURL(ctx context.Context, pluginURL, tmpDir string) (*os.File, error) {
```
JeyJeyGao, line 61: The `*os.File` has been closed, so we cannot read or write it anymore. Looks like we only need the file path. Can we only return the path instead of a closed `*os.File`?

cmd/notation/plugin/install.go
```diff
@@ -136,14 +136,13 @@ func install(command *cobra.Command, opts *pluginInstallOpts) error {
 		if pluginURL.Scheme != "https" {
 			return fmt.Errorf("failed to download plugin from URL: only the HTTPS scheme is supported, but got %s", pluginURL.Scheme)
 		}
-		tmpFile, err := os.CreateTemp("", notationPluginDownloadTmpFile)
+		tmpDir, err := os.MkdirTemp("", pluginDownloadTmpDir)
 		if err != nil {
-			return fmt.Errorf("failed to create temporary file required for downloading plugin: %w", err)
+			return fmt.Errorf("failed to create temporary directory: %w", err)
```
JeyJeyGao, line 141: Is it better to provide the temp directory path in the error so the users can check whether they have proper permission or storage size for the directory?

Two-Hearts: Resolving as discussed offline.

JeyJeyGao (dismissed):
LGTM

cmd/notation/internal/plugin/plugin.go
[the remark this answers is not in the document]
shizhMSFT: It is possible since `downloadedFilename` can be from `Content-Disposition`.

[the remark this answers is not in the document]
shizhMSFT: Even if we remove the `..` in the file path, it is still insecure / improper to have a filename based on a user-provided value.

cmd/notation/plugin/install.go
```diff
@@ -136,19 +136,18 @@ func install(command *cobra.Command, opts *pluginInstallOpts) error {
...
 		if err != nil {
-			return fmt.Errorf("failed to create temporary file required for downloading plugin: %w", err)
+			return fmt.Errorf("failed to create temporary directory required for downloading plugin: %w", err)
 		}
-		defer os.Remove(tmpFile.Name())
-		defer tmpFile.Close()
+		defer os.RemoveAll(tmpDir)
 		fmt.Printf("Downloading plugin from %s\n", opts.pluginSource)
-		err = notationplugin.DownloadPluginFromURL(ctx, opts.pluginSource, tmpFile)
+		tmpFile, err := notationplugin.DownloadPluginFromURL(ctx, opts.pluginSource, tmpDir)
```
shizhMSFT, line 145: Here's a question: How do you know the filename of `tmpFile` is properly named as `notation-<plugin_name>[.exe]`?

Two-Hearts: @shizhMSFT This is because if the plugin is a `single executable file` and it follows the notary project [spec](https://github.com/notaryproject/specifications/blob/main/specs/plugin-extensibility.md#installation), the `filename` has to be in `notation-{plugin-name}` format. If it doesn't, then Notation would fail the installation because we check the filename format before the install (in notation-go).
If the plugin is a `zip/tar.gz`, then the `filename` is not used/validated. Instead, files inside the `zip/tar.gz` are used/validated.

shizhMSFT: @Two-Hearts What if I want to install `notation-my-plugin` (plugin name `my-plugin`) from `https://packages.contoso.com/notation/plugins/notation-my-plugin-linux-am64` where `notation-my-plugin-linux-am64` does not match `notation-my-plugin`? Is there a way to resolve it?

Two-Hearts: > @Two-Hearts What if I want to install `notation-my-plugin` (plugin name `my-plugin`) from `https://packages.contoso.com/notation/plugins/notation-my-plugin-linux-am64` where `notation-my-plugin-linux-am64` does not match `notation-my-plugin`? Is there a way to resolve it?

The http response header `Content-Description filename` is checked first. If that's also not available, then one cannot directly install this plugin from URL using `notation plugin install`. They can still download the file onto file system and install from there though.

Two-Hearts:
Based on offline disccusion, closing this PR.

PR #889, opened 2024-02-13
Adding verify blob commands

rrsemlani:
Adding verify blob commands.
Spec Pr used for ref: https://github.com/notaryproject/notation/pull/811


cmd/notation/blob/verify.go
```diff
@@ -0,0 +1,118 @@
...
+Example - Verify the signature on a BLOB artifact using a policy name:
+  notation blob verify --policy-name <policy_name> --signature <signaturePath> <blobPath>
+`
+	command := &cobra.Command{
+		Use:   "blob verify [flags] --signature <signaturePath> <blobPath>",
+		Short: "Verify BLOB artifacts",
+		Long:  longMessage,
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) == 0 {
+				return errors.New("missing signature/blob path to the artifact: use `notation blob verify --help` to see what parameters are required")
```
priteshbandi, line 62:
```suggestion
				return errors.New("missing signature/blob path to the artifact: use `notation blob verify --help` to see what parameters are required")
```
```suggestion
				return errors.New("missing blob path: use `notation blob verify --help` to see what parameters are required")
```

cmd/notation/blob/verify.go
```diff
@@ -0,0 +1,118 @@
...
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if opts.maxSignatureAttempts <= 0 {
+				return fmt.Errorf("max-signatures value %d must be a positive number", opts.maxSignatureAttempts)
+			}
+			return runVerify(cmd, opts)
+		},
+	}
+	command.Flags().StringArrayVar(&opts.pluginConfig, "plugin-config", nil, "{key}={value} pairs that are passed as it is to a plugin, if the verification is associated with a verification plugin, refer plugin documentation to set appropriate values")
```
priteshbandi, line 75: are we missing `opts.LoggingFlagOpts.ApplyFlags(command.Flags())` ?

also one for signature path and policy name?

PR #916, opened 2024-03-28
feat: upgrade to OCI 1.1

### LINKED ISSUE #893: bump: update to go 1.22

-

-

-

### Any additional context?

_No response_

### LINKED ISSUE #926: Improve output message for signing in the OCI v1.0 compliant registry

I signed an image with `--allow-referrers-api` in the [distribution registry](https://distribution.github.io/distribution/about/) (OCI v1.0 compliant) which doesn't support Referrers API. The signing operation executed successfully but fallback to referrers tag schema. 

From the output message, it indicates that I signed successfully with `--allow-referrers-api` and it reminds users must set the `--allow-referrers-api` flag to list, inspect, and verify the signature. But the fact is that Notation fallback to store the signature with referrers tag schema and no need to `--allow-referrers-api` flag when verify the signaute. It might be misleading to users.

Explicitly tell users that the registry doesn't support Referrers API and will be fallback to referrers tag schema. So users don't need to set the flag `--allow-referrers-api` to list, inspect, and verify the signature.

### How can we reproduce it?

Sign an image in the distribution registry:

```
$ notation sign localhost:5000/my-ubuntu@sha256:a3785f78ab8547ae2710c89e627783cfa7ee7824d3468cae6835c9f4eae23ff7 --key demo-key --allow-referrers-api

Warning: This feature is experimental and may not be fully tested or completed and may be deprecated. Report any issues to "https://github/notaryproject/notation"
Warning: using the Referrers API to store signature. On success, must set the `--allow-referrers-api` flag to list, inspect, and verify the signature.
Successfully signed localhost:5000/my-ubuntu@sha256:a3785f78ab8547ae2710c89e627783cfa7ee7824d3468cae6835c9f4eae23ff7
```

View referrers tag schema

```
$ oras repo tags localhost:5000/my-ubuntu
sha256-a3785f78ab8547ae2710c89e627783cfa7ee7824d3468cae6835c9f4eae23ff7
latest
```

### Describe your environment

Linux Ubuntu 22.06

Notation v1.1.0

Two-Hearts:
This PR upgrades Notation to OCI 1.1.

Major changes **[UPDATE as of 4/2/2024 after community meeting]**:
1. New flag `--force-referrers-tag` is introduced. And it is only applied to the `Sign` command. It's default to `true`, and it's NOT an `experimental` flag. The original `experimental` flag `--allow-referrers-api` will be hidden, i.e., description and example will be hidden from help page. It is kept only for backwards compatibility purpose, and a warning will be printed out when user sets it.
2. Sign: 
     ```
     # Default behavior: Use the referrers tag schema for backwards compatibility.
     notation sign ...
     notation sign --force-referrers-tag ...
     
    # With `--force-referrers-tag=false`: Check the Referrers API first, if not supported, automatically fallback to the referrers tag schema.
     notation sign --force-referrers-tag=false ...
     ```
3. Verify/List/Inspect: They will always use the Referrers API first, if not supported, automatically fallback to the referrers tag schema. Note: for backwards compatibility reason, the experimental flag `--allow-referrers-api` is still kept and hidden. When set, a warning will be printed out. 

Changes are tested in E2E and ACR.

This PR partially resolves issue https://github.com/notaryproject/notation/issues/892. Resolves #893. Resolves #926,


.github/workflows/build.yml
shizhMSFT, line 1: One PR one purpose. Can you split the bump up go version part to another PR?

sudo-bmitch:
This may be too late to change, but if not, I'd suggest renaming the tag to something more clear to end users:

```
--force-1-1-compatibility: Pushes content to registries required to support notation 1.1.0 and earlier clients
```

Default the value to true for now, and on a 2.x release, change the default to false.

From an outside view, `--allow-referrers-api` is confusing since the signing operation isn't calling the API and the client has no say in whether the registry will include the content in the referrers response, that's automatic based on the `subject` field.

priteshbandi:
> This may be too late to change, but if not, I'd suggest renaming the flag to something more clear to end users:
> 
> ```
> --force-1-1-compatibility: Pushes content to registries required to support notation 1.1.0 and earlier clients
> ```
> 
> Default the value to true for now, and on a 2.x release, change the default to false.
> 
> From an outside view, `--allow-referrers-api` is confusing since the signing operation isn't calling the API and the client has no say in whether the registry will include the content in the referrers response, that's automatic based on the `subject` field.

+1, can we add `--force-oci-1-1` and make `--allow-referrers-api` its alias ?

Two-Hearts:
I'm actually seeing two suggestions here:
1. Suggested by @sudo-bmitch  : `--force-1-1-compatibility`, here `1-1` means notation 1.1.0 and before. This flag is default to `true`. In this case:
```
Sign with the referrers tag schema by default:
notation sign ...
notation sign --force-1-1-compatibility ...

Sign with the referrers API:
notation sign --force-1-1-compatibility=false ...
```

2. Suggested by @priteshbandi  : `--force-oci-1-1`, an alias of the current `--allow-referrers-api`, default to `false`. Since `--allow-referrers-api` is currently under `experimental`, we could safely rename it to `--force-oci-1-1`. In this case:
```
Sign with the referrers tag schema by default:
notation sign ...

Sign with the referrers API:
notation sign --force-oci-1-1 ...
```

Maybe a vote? @shizhMSFT @yizha1 @FeynmanZhou @JeyJeyGao @rgnote

.github/workflows/build.yml
Two-Hearts: @shizhMSFT To upgrade to OCI 1.1, we need oras-go v2.5.0. Based on oras-go v2.5.0 release note: https://github.com/oras-project/oras-go/releases/tag/v2.5.0, it does not compile with Go 1.20.x or earlier.

Therefore, keeping our build to go 1.20 will cause this PR's CI failure.

shizhMSFT: Oh. I wasn’t aware of notation being two versions behind.

shizhMSFT:
> This may be too late to change, but if not, I'd suggest renaming the flag to something more clear to end users:
> 
> ```
> --force-1-1-compatibility: Pushes content to registries required to support notation 1.1.0 and earlier clients
> ```
> 
> Default the value to true for now, and on a 2.x release, change the default to false.
> 
> From an outside view, `--allow-referrers-api` is confusing since the signing operation isn't calling the API and the client has no say in whether the registry will include the content in the referrers response, that's automatic based on the `subject` field.
> 
> Edit: note this is a focus on the notation version, rather than the OCI version, since users would more likely know the version of their notation clients while they don't necessarily know the version of their registry.

It makes sense as pushing manifests with subjects does not really call the referrers API as defined in `distribution-spec v1.1.0` although it did call as defined in `distribution-spec` RC versions.

The `--force-1-1-compatibility` is good but confusing since it means more than registry interactions.
Therefore, I'd suggest `--force-referrers-tag`. In notation 1.x, the default value is `--force-referrers-tag=true` for backward compatibility. In notation 2.x, the default value can set to `false`.

yizha1:
I vote for the flag name `--force-referrers-tag`.

priteshbandi:
As discussed in Today's call, I'm fine with both `--force-referrers-tag` or --force-tag-schema` with defualt value of true in. (1.2 release). This serves the intent of using tags for storing signature and doesn't uses oci version.

FeynmanZhou:
I am kind of @shizhMSFT 's proposal, i.e., 

- `--force-referrers-tag` and its default value is true: **From end users' point of view, they care about how signature can be stored in the registry. Apparently, the signature will be pushed and stored as a "sha256-xxxxx" tag in the registry with the default `--force-referrers-tag=true`.**
- Make `--allow-referrers-api` as an alias: For some popular registries that already supported OCI v1.1 and integrated Notation, such as Harbor and Zot, `--allow-referrers-api` has been referenced in their documentation. Keep the original flag name will be friendly and non-breaking for those.

For @sudo-bmitch and @priteshbandi 's proposals above, `--force-1-1-compatibility` and `force-oci-1-1` sound a bit vague and technical because most of end users are not aware of the OCI spec and different versions. In general, users care about which signature format they can use to store the signature in registry.

yizha1:
I think `--force-referrers-tag` is better than `--force-tag-schema`, as it shows that the `tag` is related to `referrers`, and also a short version of the term `Referrers Tag Schema` used in OCI specification, see the screenshot below

image

JeyJeyGao:
I vote for `--force-referrers-tag` with `true` as the default value. I also want to mention that it conflicts with the original flag `--allow-referrers-api`. If we also need to keep the original flag, we should ensure they are mutually exclusive.

Two-Hearts:
Going for `--force-referrers-tag` as it gets the highest vote. 

In this PR:
Renaming `--allow-referrers-api` to `--force-referrers-tag`, moving it out of experimental.

sudo-bmitch:
> For @sudo-bmitch ... proposals above, `--force-1-1-compatibility` and `force-oci-1-1` sound a bit vague and technical because most of end users are not aware of the OCI spec and different versions.

This reason is why my suggestion wasn't about OCI, referrers, or a tag schema. Users may not  know what they are or why they would want it. The `1-1-compatibility` flag described that it was needed to support 1.1 versions of *notation clients*, which tells users when and why they would need to use the flag.

> ```
> --force-1-1-compatibility: Pushes content to registries required to support notation 1.1.0 and earlier clients
> ```

> The --force-1-1-compatibility is good but confusing since it means more than registry interactions.

My understanding here is that it got turned down because you have other flags that users will also need to use to support older versions of notation, which I was not aware of. That may be an indication your users won't be aware of that either.

building.md
```diff
@@ -4,7 +4,7 @@ The notation repo contains the following:
 
 - `notation` - A CLI for signing and verifying artifacts with Notation
 
-Building above binaries require [golang](https://golang.org/dl/) with version `>= 1.20`.
+Building above binaries require [golang](https://golang.org/dl/) with version `>= 1.21`.
```
JeyJeyGao, line 7: do we need >= 1.22?

cmd/notation/registry.go
```diff
@@ -57,17 +56,18 @@ func getRepository(ctx context.Context, inputType inputType, reference string, o
 }
 
 // getRemoteRepository returns a registry.Repository.
-// When experimental feature is disabled OR allowReferrersAPI is not set,
-// Notation always uses referrers tag schema to store and consume signatures
-// by default.
-// When experimental feature is enabled AND allowReferrersAPI is set, Notation
-// tries the Referrers API, if not supported, fallback to use the Referrers
-// tag schema.
+// When forceReferrersTag is set, Notation will always generate an image index
```
JeyJeyGao, line 59: for this function, we always need to set `forceReferrersTag`, so is it better to be:
```suggestion
// When forceReferrersTag is true, Notation will always generate an image index
```

cmd/notation/registry.go
```diff
@@ -57,17 +56,18 @@ func getRepository(ctx context.Context, inputType inputType, reference string, o
...
+// according to the Referrers tag schema to store signature.
+//
+// When forceReferrersTag is not set, Notation will first try to store the
```
JeyJeyGao, line 62:
```suggestion
// When forceReferrersTag is false, Notation will first try to store the
```

building.md
Two-Hearts: 1.21 should still be compatible so I included it in the range.

JeyJeyGao: The `go.mod` file has been updated to 1.22. Why can it still support 1.21?

Two-Hearts: Make sense. Updated.

JeyJeyGao (dismissed):
LGTM

cmd/notation/sign.go
```diff
@@ -102,6 +101,15 @@ Example - [Experimental] Sign an OCI artifact identified by a tag and referenced
 			return experimental.CheckFlagsAndWarn(cmd, "allow-referrers-api", "oci-layout")
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			// allow-referrers-api flag is set
+			if cmd.Flags().Changed("allow-referrers-api") {
+				if opts.allowReferrersAPI {
+					fmt.Fprintln(os.Stderr, "Warning: flag '--allow-referrers-api' is deprecated, use '--force-referrers-tag=false' instead.")
```
priteshbandi, line 107:
```suggestion
					fmt.Fprintln(os.Stderr, "warning: flag '--allow-referrers-api' flag is deprecated and will be removed in a future version, use '--force-referrers-tag=false' instead.")
```

cmd/notation/sign.go
```diff
@@ -102,6 +101,15 @@ Example - [Experimental] Sign an OCI artifact identified by a tag and referenced
...
+					opts.forceReferrersTag = false
+				} else {
+					fmt.Fprintln(os.Stderr, "Warning: flag '--allow-referrers-api' is deprecated, use '--force-referrers-tag' instead.")
```
priteshbandi, line 110: Why do we need this message ?

cmd/notation/sign_test.go
```diff
@@ -154,7 +159,9 @@ func TestSignCommmand_OnDemandKeyOptions(t *testing.T) {
 		"-u", expected.Username,
 		"--password", expected.Password,
 		"--id", expected.KeyID,
-		"--plugin", expected.PluginName}); err != nil {
+		"--plugin", expected.PluginName,
+		"--force-referrers-tag=false",
```
priteshbandi, line 163: Why are we making this change? Also, applies to below tests

cmd/notation/verify.go
```diff
@@ -90,15 +87,18 @@ Example - [Experimental] Verify a signature on an OCI artifact identified by a t
 			if opts.maxSignatureAttempts <= 0 {
 				return fmt.Errorf("max-signatures value %d must be a positive number", opts.maxSignatureAttempts)
 			}
+			if cmd.Flags().Changed("allow-referrers-api") {
+				fmt.Fprintln(os.Stderr, "Warning: flag '--allow-referrers-api' is deprecated and ignored.")
```
priteshbandi, line 91: Please update message to show it would be removed in future versions like major version bump.

test/e2e/suite/command/inspect.go
```diff
@@ -77,4 +77,58 @@ var _ = Describe("notation inspect", func() {
...
+		Host(BaseOptionsWithExperimental(), func(notation *utils.ExecOpts, artifact *Artifact, vhost *utils.VirtualHost) {
+			notation.Exec("sign", "--allow-referrers-api", artifact.ReferenceWithDigest()).
+				MatchKeyWords(SignSuccessfully)
+
+			notation.Exec("inspect", artifact.ReferenceWithDigest(), "-v").
+				MatchKeyWords(inspectSuccessfully...)
+
+			notation.Exec("inspect", artifact.ReferenceWithDigest(), "--allow-referrers-api", "-v").
+				MatchErrKeyWords(
+					"Warning: This feature is experimental and may not be fully tested or completed and may be deprecated.",
```
priteshbandi, line 111: This line should not be emitted. What am I missing?

cmd/notation/sign_test.go
Two-Hearts: > Why are we making this change? Also, applies to below tests

Oh, it's just an extra check, having `--force-referrers-tag=false` is equivalent to not setting the `--force-referrers-tag` flag at all.

test/e2e/suite/command/inspect.go
Two-Hearts: This is because `--allow-referrers-api` flag is still kept under `experimental`. Its behavior is not changed for backwards compatible purpose.

cmd/notation/inspect.go
```diff
@@ -106,6 +102,9 @@ Example - [Experimental] Inspect signatures on an OCI artifact identified by a d
 			if opts.maxSignatures <= 0 {
 				return fmt.Errorf("max-signatures value %d must be a positive number", opts.maxSignatures)
 			}
+			if cmd.Flags().Changed("allow-referrers-api") {
+				fmt.Fprintln(os.Stderr, "Warning: flag '--allow-referrers-api' is deprecated and ignored.")
```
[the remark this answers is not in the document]
Two-Hearts: To align with other commands such as `policy`, I'll keep the `Warning` here capitalized.

cmd/notation/sign.go
Two-Hearts: Updated.

test/e2e/suite/command/inspect.go
priteshbandi: Why does this flag needs to be behind experimental flag, when we are going to keep implementation and are emitting warning?

Two-Hearts: > Why does this flag needs to be behind experimental flag, when we are going to keep implementation and are emitting warning?

@priteshbandi `--allow-referrers-api` flag is already behind experimental as of today. I'm just keeping its behavior while hide the flag from new users only because we want to deprecate this experimental flag gracefully.

cmd/notation/inspect.go
```diff
@@ -115,7 +114,7 @@ Example - [Experimental] Inspect signatures on an OCI artifact identified by a d
 	cmd.SetPflagOutput(command.Flags(), &opts.outputFormat, cmd.PflagOutputUsage)
 	command.Flags().IntVar(&opts.maxSignatures, "max-signatures", 100, "maximum number of signatures to evaluate or examine")
 	cmd.SetPflagReferrersAPI(command.Flags(), &opts.allowReferrersAPI, fmt.Sprintf(cmd.PflagReferrersUsageFormat, "inspect"))
-	experimental.HideFlags(command, experimentalExamples, []string{"allow-referrers-api"})
+	experimental.HideFlags(command, "", []string{"allow-referrers-api"})
```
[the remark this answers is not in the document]
Two-Hearts: Updated to remove it.

shizhMSFT (approved):
:GTM

test/e2e/suite/command/inspect.go
yizha1: @priteshbandi, does the comment from Patrick make sense to you? We want to deprecate the experimental flag gracefully and remove it in v1.3.0 release completely. In v1.2.0 release note, we will introduce the new flag `--force-referrers-tag` and mention the existing one `--allow-referrers-api` is deprecated and will be completely removed in v1.3.0. This is a more user-friendly approach and gives existing users a time window to replace the old flag with the new flag.

JeyJeyGao (approved):
LGTM

priteshbandi (approved):
LGTM

PR #978, opened 2024-06-27
feat: Timestamp

### LINKED ISSUE #838: Support Time-stamping

Signature validation will fail if signing key/certificates are expired. In this case, signers need to rotate the key/certificates in time and re-sign container images using new key/certificates. The lifetime of a signature is limited by the lifetime of a corresponding signing key/certificate.

Time-stamping (https://www.rfc-editor.org/rfc/rfc3161) extends the trust of signature beyond the validity period of a certificate, thus signers do not need to regularly re-sign images before certificates are expired. RFC 3161 said that Time-stamping helps establish whether the image was signed before or after the certificate was compromised. However, in the real world, the compromised time is not the time of certificate revocation. The compromised time is not determined. So, the proposal is to not use Time-stamping for revocation scenarios.

Currently Notation does not support Time-stamping, so signers need to regularly rotate the key/certificates in time and re-sign container images using new key/certificates. However, this process requires signers to establish a re-sign workflow, which leads to usability problems and waste of resources.

### Any additional context?

There is an [existing issue](https://github.com/notaryproject/roadmap/issues/59) in roadmap repo, however this repo is not in active mode.

Two-Hearts:
This PR adds timestamping support including corresponding E2E tests and spec changes.

Resolves #838 

Example of sign with timestamping:
```
notation sign --timestamp-url <TSA_url> --timestamp-root-cert <TSA_root_certificate_filepath> <registry>/<repository>@<digest>
```

Example of verify with timestamping:
```
notation verify <registry>/<repository>@<digest>
```


cmd/notation/sign.go
```diff
@@ -120,9 +143,12 @@ Example - [Experimental] Sign an OCI artifact identified by a tag and referenced
 	cmd.SetPflagPluginConfig(command.Flags(), &opts.pluginConfig)
 	cmd.SetPflagUserMetadata(command.Flags(), &opts.userMetadata, cmd.PflagUserMetadataSignUsage)
 	cmd.SetPflagReferrersAPI(command.Flags(), &opts.allowReferrersAPI, fmt.Sprintf(cmd.PflagReferrersUsageFormat, "sign"))
+	command.Flags().StringVar(&opts.tsaServerURL, "tsa-url", "", "timestamp authority server URL")
+	command.Flags().StringVar(&opts.tsaRootCertificatePath, "tsa-root-cert", "", "filepath of timestamp authority root certificate")
```
JeyJeyGao, line 147: Do we need to mention the supported format for the root certificate?

cmd/notation/sign.go
```diff
@@ -189,5 +215,23 @@ func prepareSigningOpts(opts *signOpts) (notation.SignOptions, error) {
 		},
 		UserMetadata: userMetadata,
 	}
+	if opts.tsaServerURL != "" {
+		// timestamping
+		fmt.Printf("Timestamping with TSA %q\n", opts.tsaServerURL)
+		signOpts.Timestamper, err = tspclient.NewHTTPTimestamper(&http.Client{Timeout: TimestampingTimeout}, opts.tsaServerURL)
+		if err != nil {
+			return notation.SignOptions{}, fmt.Errorf("cannot get http timestamper for timestamping: %v", err)
```
JeyJeyGao, line 223:
```suggestion
			return notation.SignOptions{}, fmt.Errorf("cannot get http timestamper for timestamping: %w", err)
```

cmd/notation/sign.go
```diff
@@ -189,5 +215,23 @@ func prepareSigningOpts(opts *signOpts) (notation.SignOptions, error) {
...
+		}
+
+		rootCerts, err := corex509.ReadCertificateFile(opts.tsaRootCertificatePath)
+		if err != nil {
+			return notation.SignOptions{}, err
+		}
+		if len(rootCerts) == 0 {
+			return notation.SignOptions{}, fmt.Errorf("cannot read tsa root certificate from %q", opts.tsaRootCertificatePath)
```
JeyJeyGao, line 231: Should we mention that the file is empty?

JeyJeyGao: If the file contains more than one certs, should we remind the user that we will only use the first one.

Two-Hearts: The file is not necessarily empty though. It's just there's no valid x509 certificate in it. The root cert downloaded from TSA's webpage is usually a single cert file.

cmd/notation/sign.go
```diff
@@ -14,34 +14,44 @@
...
+	"github.com/notaryproject/tspclient-go"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/spf13/cobra"
 )
 
 const referrersTagSchemaDeleteError = "failed to delete dangling referrers index"
 
+// TimestampingTimeout is the timeout when requesting timestamp countersignature
+// from a TSA
+const TimestampingTimeout = 15 * time.Second
```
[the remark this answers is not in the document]
Two-Hearts: Unexported.

cmd/notation/sign.go
Two-Hearts: I think that's too detail for a description of a flag? I haven't seen a TSA root cert that cannot be read by Notation so far.

cmd/notation/sign.go
```diff
@@ -189,5 +215,23 @@ func prepareSigningOpts(opts *signOpts) (notation.SignOptions, error) {
 		},
 		UserMetadata: userMetadata,
 	}
+	if opts.tsaServerURL != "" {
+		// timestamping
+		fmt.Printf("Timestamping with TSA %q\n", opts.tsaServerURL)
```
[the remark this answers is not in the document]
Two-Hearts: Updated.

cmd/notation/sign.go
```diff
@@ -189,5 +215,23 @@ func prepareSigningOpts(opts *signOpts) (notation.SignOptions, error) {
...
+		signOpts.Timestamper, err = tspclient.NewHTTPTimestamper(&http.Client{Timeout: TimestampingTimeout}, opts.tsaServerURL)
```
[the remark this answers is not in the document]
Two-Hearts: Looks like we cannot use the `auth.Client` here, because the `tspclient.NewHTTPTimestamper` takes an `*http.Client` as input.

Since the OCSP client has the same issue, I will create an issue to track it. And we could fix it in the next iteration.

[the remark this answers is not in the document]
Two-Hearts: Issue created: https://github.com/notaryproject/notation/issues/988

cmd/notation/sign.go
JeyJeyGao: Yes, we need to mention the reason and expected format to help the user figure out the issue.

> cannot read tsa root certificate from %q

It doesn't explain the reason.

specs/commandline/sign.md
```diff
@@ -155,6 +157,20 @@ Warning: Always sign the artifact using digest(`@sha256:...`) rather than a tag(
...
+```shell
+# Prerequisites:
+# A default signing key is configured using CLI "notation key".
+# Signer knows the TSA url that they want to use to require a RFC 3161 timestamp.
+# Signer has downloaded the TSA's root certificate in their file system.
+
+# Use option "--tsa-url" to specify the timestamp authority URL.
+# Use option "--tsa-root-cert" to specify the filepath of the tsa root
+# certificate.
+notation sign --tsa-url <tsa_url> --tsa-root-cert <tsa_root_certificate_filepath> <registry>/<repository>@<digest>
```
FeynmanZhou, line 171: Does the flag `--tsa-root-cert` allow an URL when passing a root cert? 

For example, users are running notation in a CI/CD platform and its runner machine is a temporary VM resource, users have to download the root cert into the runner machine filesystem every time before every signing action. It might not be efficient and performant in this case.

specs/commandline/sign.md
```diff
@@ -42,6 +42,8 @@ Flags:
        --plugin string              signing plugin name. This is mutually exclusive with the --key flag
        --plugin-config stringArray  {key}={value} pairs that are passed as it is to a plugin, refer plugin's documentation to set appropriate values.
        --signature-format string    signature envelope format, options: "jws", "cose" (default "jws")
+       --tsa-root-cert string       filepath of timestamp authority root certificate
```
yizha1, line 45: How about `--tsa-cert-chain` filepath to the certificate chain file for the RFC3161 Timestamping Authority (TSA). Must contain the root CA certificate. Optionally it may contain intermediate CA certificates.

specs/commandline/sign.md
```diff
@@ -42,6 +42,8 @@ Flags:
...
+       --tsa-url string             timestamp authority server URL
```
yizha1, line 46:
```suggestion
       --tsa-url string             RFC3161 Timestamping Authority (TSA) server URL
```

specs/commandline/sign.md
```diff
@@ -155,6 +157,20 @@ Warning: Always sign the artifact using digest(`@sha256:...`) rather than a tag(
 Successfully signed localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
 ```
 
+### Sign an OCI artifact and timestamp the signature with user specified timestamp authority
```
yizha1, line 160: We can explain TSA in one place, then refer to the short name TSA directly.

```suggestion
### Sign an OCI artifact and timestamp the signature with user specified RFC3161 Timestamp Authority (TSA)
```

yizha1 (commented):
Thanks @Two-Hearts

Two-Hearts:
Based on 7/15/2024 community meeting, sharing my investigations on TSA servers:
image

As you can see, TSAs are including their intermediate certs in the response. 7 out of 10 are providing the root certificate as well. However, as we discussed earlier, due to potential MITM attack, we must require the user to provide their own trust anchor in the timestamping process. Therefore, in `beta.1` release, we propose to only require user to provide the TSA's root cert. If, in the future, there's use case asking for intermediate chain, we could always update the design.

@priteshbandi @shizhMSFT @toddysm @yizha1 @JeyJeyGao @FeynmanZhou

specs/commandline/sign.md
Two-Hearts: @yizha1 Please see my research on TSA servers below: https://github.com/notaryproject/notation/pull/978#issuecomment-2229929172

The TSAs are already including their intermediate certs in the response, I'm not seeing any extra benefit of user providing them again. Therefore, in this iteration, I would propose only require user to provide the TSA root cert.

specs/commandline/sign.md
Two-Hearts: Then that's another network roundtrip for Notation.

For a CI/CD pipeline signing, user does NOT have to download the root cert every time. They could simply download it once and save it in the repo, the CI/CD pipeline could retrieve the cert from the repo. For example, in GitHub workflow, this is done by [`actions/checkout`](https://github.com/actions/checkout).

cmd/notation/sign.go
Two-Hearts: Updated.

specs/commandline/sign.md
Two-Hearts: Updated.

specs/commandline/sign.md
Two-Hearts: Updated.

cmd/notation/sign.go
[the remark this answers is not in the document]
Two-Hearts: Based on offline discussion, updated the `http.Client` in this PR.

internal/httputil/client.go
```diff
@@ -22,14 +22,49 @@ import (
...
+func NewClient(ctx context.Context, client *http.Client) *http.Client {
+	client = trace.SetHTTPDebugLog(ctx, client)
+	return SetUserAgent(client)
+}
+
+type userAgentTransport struct {
+	base http.RoundTripper
+}
+
+func (t *userAgentTransport) RoundTrip(req *http.Request) (*http.Response, error) {
```
shizhMSFT, line 49: nit: request documentation

cmd/notation/sign.go
```diff
@@ -143,7 +145,7 @@ Example - [Experimental] Sign an OCI artifact identified by a tag and referenced
 	cmd.SetPflagPluginConfig(command.Flags(), &opts.pluginConfig)
 	cmd.SetPflagUserMetadata(command.Flags(), &opts.userMetadata, cmd.PflagUserMetadataSignUsage)
 	cmd.SetPflagReferrersAPI(command.Flags(), &opts.allowReferrersAPI, fmt.Sprintf(cmd.PflagReferrersUsageFormat, "sign"))
-	command.Flags().StringVar(&opts.tsaServerURL, "tsa-url", "", "timestamp authority server URL")
+	command.Flags().StringVar(&opts.tsaServerURL, "tsa-url", "", "RFC3161 Timestamping Authority (TSA) server URL")
```
shizhMSFT, line 148: nit:
```suggestion
	command.Flags().StringVar(&opts.tsaServerURL, "tsa-url", "", "RFC 3161 Timestamping Authority (TSA) server URL")
```

shizhMSFT (approved):
LGTM with suggestions

priteshbandi:
instead of tsa can we call it timestamp `--timestamp <tsa_url> timestamp-root-cert <location>` ?

---

> Therefore, in beta.1 release, we propose to only require user to provide the TSA's root cert. If, in the future, there's use case asking for intermediate cert, we could always update the design.

Since we are planning to do a GA release soon, updating name would be challenging. However, if we have suggestions on how we can support the 'intermediate cert' in future updates, we are good for beta.1.

Two-Hearts:
> instead of tsa can we call it timestamp `--timestamp <tsa_url> timestamp-root-cert <location>` ?

@yizha1 @FeynmanZhou do you have any comments on the new proposed flag names above?

> Since we are planning to do a GA release soon, updating name would be challenging. However, if we have suggestions on how we can support the 'intermediate cert' in future updates, we are good for beta.1.

Some options:
1. We encourage the user to switch to use root cert as trust anchor. (For those TSAs that NOT including their root in the response, a root cert MUST be provided by the user anyway.)
2. Introducing a new flag called `timestamp-intermediate-cert`. 

> --
> 
> I agree we should discourage intermediate pinning.

Yes, I also agree that we should discourage using intermediate as trust anchor.

yizha1:
@priteshbandi @Two-Hearts @shizhMSFT 

How about using `--timestamp-url`?

I am good with `--timestamp-root-cert`.

shizhMSFT:
> instead of tsa can we call it timestamp --timestamp <tsa_url> timestamp-root-cert <location> ?

The flags `--timestamp` and `--timestamp-root-cert` LGTM. I also agree with @yizha1 that `--timestamp-url` is a better flag name for `--timestamp`.

specs/commandline/sign.md
```diff
@@ -42,8 +42,8 @@ Flags:
        --plugin string              signing plugin name. This is mutually exclusive with the --key flag
        --plugin-config stringArray  {key}={value} pairs that are passed as it is to a plugin, refer plugin's documentation to set appropriate values.
        --signature-format string    signature envelope format, options: "jws", "cose" (default "jws")
-       --tsa-root-cert string       filepath of timestamp authority root certificate
-       --tsa-url string             RFC3161 Timestamping Authority (TSA) server URL
+       --timestamp-root-cert string       filepath of timestamp authority root certificate
+       --timestamp-url string             RFC 3161 Timestamping Authority (TSA) server URL
   -u,  --username string            username for registry operations (default to $NOTATION_USERNAME if not specified)
```
shizhMSFT, line 47: The rendering is broken: descriptions are not aligned.

Two-Hearts: Ah, nice catch.

shizhMSFT (approved):
LGTM

JeyJeyGao (dismissed):
LGTM

cmd/notation/sign.go
```diff
@@ -189,5 +217,24 @@ func prepareSigningOpts(opts *signOpts) (notation.SignOptions, error) {
...
+		if err != nil {
+			return notation.SignOptions{}, fmt.Errorf("cannot get http timestamper for timestamping: %w", err)
+		}
+
+		rootCerts, err := corex509.ReadCertificateFile(opts.tsaRootCertificatePath)
+		if err != nil {
+			return notation.SignOptions{}, err
+		}
+		if len(rootCerts) == 0 {
+			return notation.SignOptions{}, fmt.Errorf("cannot find any tsa root certificate from %q. Expecting x509 certificate in PEM or DER format from the file", opts.tsaRootCertificatePath)
```
priteshbandi, line 233:
```suggestion
			return notation.SignOptions{}, fmt.Errorf("%q doesnot contain certificate(s). Expecting x509 certificate in PEM or DER format from the file", opts.tsaRootCertificatePath)
```

Are we really checking tsa root cert or just that there are no certificates ?

cmd/notation/sign.go
```diff
@@ -189,5 +217,24 @@ func prepareSigningOpts(opts *signOpts) (notation.SignOptions, error) {
...
+		}
+		rootCAs := x509.NewCertPool()
+		rootCAs.AddCert(rootCerts[0])
```
priteshbandi, line 236: If we are going to use first cert, then we shouldn't allow multiple cert in the file. i.e throw error if there are multiple certs in file.

cmd/notation/sign.go
```diff
@@ -189,5 +217,24 @@ func prepareSigningOpts(opts *signOpts) (notation.SignOptions, error) {
 		},
 		UserMetadata: userMetadata,
 	}
+	if opts.tsaServerURL != "" {
+		// timestamping
+		fmt.Printf("Configured to timestamp with TSA %q\n", opts.tsaServerURL)
```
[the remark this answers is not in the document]
Two-Hearts: This message is shown during signing. It is the only print out that's showing the TSA url.

cmd/notation/sign.go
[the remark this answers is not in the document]
Two-Hearts: Good point. It's been checked in the tspclient-go library during verification. But yes, we could check it here and fail fast.

cmd/notation/sign.go
Two-Hearts: I think this is an implementation detail, because this helper function from `notation-core-go` returns a list of cert. This is not exposed to the external user. Based on the flag description, the user will only provide the root cert.

cmd/notation/sign.go
Two-Hearts: Tracking the comment above.

cmd/notation/sign.go
Two-Hearts: Updated.

cmd/notation/sign.go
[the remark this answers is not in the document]
Two-Hearts: Updated.

JeyJeyGao (dismissed):
LGTM

internal/x509/cert.go
```diff
@@ -0,0 +1,30 @@
...
+
+// IsRootCertificate returns true if cert is a root certificate.
+// A root certificate MUST be a self-signed and self-issued CA certificate with
+// valid BasicConstraints.
+func IsRootCertificate(cert *x509.Certificate) (bool, error) {
+	// CheckSignatureFrom also checks cert.BasicConstraintsValid
+	if err := cert.CheckSignatureFrom(cert); err != nil {
+		return false, err
+	}
+	return cert.IsCA && bytes.Equal(cert.RawSubject, cert.RawIssuer), nil
```
shizhMSFT, line 29: Is `cert.IsCA` required? Can it be just a self-signed certificate? Is there any spec reference?

Two-Hearts: Yeah, spec: https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#root-and-intermediate-ca-certificates

Basically, we require root and intermediate certs to be CA issued:
`The cA field MUST be set to true`

shizhMSFT: The spec states
> Valid certificate chain MUST contain a root certificate. If the certificate chain contains only a root certificate then the root certificate MUST meet [Leaf Certificates](https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#leaf-certificates) requirements and ignore [Root and Intermediate CA Certificates](https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#root-and-intermediate-ca-certificates) requirements.

Two-Hearts: Yes, but do we want to support the self-signed case? This is an `internal` helper function for `--tsa-root-cert`. Based on our testing, none of the TSA is using a chain of length 1 so far.

shizhMSFT: What about private TSAs? Of course, we can improve that in the next iteration.

Two-Hearts: Updated.

shizhMSFT (approved):
LGTM

cmd/notation/sign.go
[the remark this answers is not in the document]
priteshbandi: IMO showing this is redundant during sign also. I am fine emitting it under -v flag.

priteshbandi (approved):
LGTM with a nit

cmd/notation/sign.go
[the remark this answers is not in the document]
Two-Hearts: Makes sense to me, created an issue to track it: https://github.com/notaryproject/notation/issues/992

Will update in the next iteration.

PR #538, opened 2023-02-06
bump: go 1.19 to 1.20

mintbomb27:
Fixes #535 

I wasn't able to find a `Dockerfile`. If you could guide me, I will change it there too @shizhMSFT

PR #1019, opened 2024-08-26
bump: upgrade golang version to v1.23

Two-Hearts:


priteshbandi (dismissed):
LGTM

vaninrao10 (approved):
LGTM

test/e2e/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/notaryproject/notation/test/e2e
 
-go 1.21
+go 1.22
```
shizhMSFT, line 3: I'm curious why the golang version for e2e test is not the same as the main one?

shizhMSFT (commented):
The dependency bump up should also be cherry-picked or backported to `v1.2.0`.

shizhMSFT (changes requested):
It seems #538 did the right thing and this PR does not.

test/e2e/go.mod
Two-Hearts: Updated.

JeyJeyGao:
what about `building.md`?

> Building above binaries require [golang](https://golang.org/dl/) with version `>= 1.22`.

Two-Hearts:
> what about `building.md`?
> 
> > Building above binaries require [golang](https://golang.org/dl/) with version `>= 1.22`.

This should be correct? `>= 1.22` includes 1.23.

shizhMSFT:
> what about `building.md`?
> 
> > Building above binaries require [golang](https://golang.org/dl/) with version `>= 1.22`.

Good catch. Since `go.mod` requires `1.23`, we should update the `building.md` for `1.23`. This is also for security purposes.

JeyJeyGao (approved):
LGTM

shizhMSFT (approved):
LGTM

priteshbandi (approved):
LGTM

PR #1076, opened 2024-11-01
fix: discard crl cache error

Two-Hearts:
This PR discards crl cache errors as they are not considered as critical in the verification path. After the fix, verification won't be terminated on crl cache error.


cmd/notation/notation
shizhMSFT, line 1: This file should not be checked in.

shizhMSFT (changes requested):

cmd/notation/verify.go
```diff
@@ -246,6 +246,7 @@ func getVerifier(ctx context.Context) (notation.Verifier, error) {
 	if err != nil {
 		return nil, err
 	}
+	crlFetcher.DiscardCacheError = true // discard cache error
```
priteshbandi, line 249: Should we throw warning when we are discarding cache error? IMO user would like to know if there caching mechanism is not working

Two-Hearts: Added log.

shizhMSFT: We should not always log `Discard any crl cache error`. We should only log when there is an error and discarded.

Two-Hearts: Then this log needs to be added in notation-core-go, because from notation-go level the error has already been discarded. Removed from notation CLI.

shizhMSFT (approved):
LGTM

cmd/notation/verify.go
JeyJeyGao: I think for notation-core-go, we cannot use the logger. Should we add the warning log in notation-go?

JeyJeyGao (approved):
LGTM

cmd/notation/verify.go
priteshbandi: Can we please open issue to track this issue.

priteshbandi (approved):
LGTM

cmd/notation/verify.go
Two-Hearts: Issue created: https://github.com/notaryproject/notation-core-go/issues/242

PR #1078, opened 2024-11-08
feat: crl cache with log

Two-Hearts:
This PR adds a wrapper of [FileCache](https://github.com/notaryproject/notation-go/blob/f855f2552698173168413cfb0e4c746a3dffb0a9/verifier/crl/crl.go#L52) to log any cache errors during revocation check. It also prints out a one-time warning if the cache error is set to be discarded, which is true for Notation CLI.

### DIFF
--- a/internal/crl/crl.go
+++ b/internal/crl/crl.go
@@ -1,0 +1,64 @@
+// Copyright The Notary Project Authors.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package crl
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"sync"
+
+	corecrl "github.com/notaryproject/notation-core-go/revocation/crl"
+	"github.com/notaryproject/notation-go/log"
+)
+
+// CrlCacheWithLog implements corecrl.Cache with logging
+type CrlCacheWithLog struct {
+	corecrl.Cache
+
+	// logDiscardCrlCacheErrorOnce guarantees the discard crl cache error
+	// warning is logged only once
+	logDiscardCrlCacheErrorOnce sync.Once
+}
+
+// Get retrieves the CRL bundle with the given url
+func (c *CrlCacheWithLog) Get(ctx context.Context, url string) (*corecrl.Bundle, error) {
+	logger := log.GetLogger(ctx)
+
+	bundle, err := c.Cache.Get(ctx, url)
+	if err != nil && !errors.Is(err, corecrl.ErrCacheMiss) {
+		c.logDiscardCrlCacheErrorOnce.Do(c.logDiscardCrlCacheError)
+		logger.Debug(err.Error())
+	}
+	return bundle, err
+}
+
+// Set stores the CRL bundle with the given url
+func (c *CrlCacheWithLog) Set(ctx context.Context, url string, bundle *corecrl.Bundle) error {
+	logger := log.GetLogger(ctx)
+
+	err := c.Cache.Set(ctx, url, bundle)
+	if err != nil {
+		c.logDiscardCrlCacheErrorOnce.Do(c.logDiscardCrlCacheError)
+		logger.Debug(err.Error())
+	}
+	return err
+}
+
+// logDiscardCrlCacheError logs the warning when CRL cache error is
+// discarded
+func (c *CrlCacheWithLog) logDiscardCrlCacheError() {
+	fmt.Fprintln(os.Stderr, "Warning: CRL cache error discarded")
+}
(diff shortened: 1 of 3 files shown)


internal/crl/crl.go
```diff
@@ -0,0 +1,72 @@
...
+	"fmt"
+	"os"
+	"sync"
+
+	corecrl "github.com/notaryproject/notation-core-go/revocation/crl"
+	"github.com/notaryproject/notation-go/log"
+)
+
+// CrlCacheWithLog implements corecrl.Cache with logging
+type CrlCacheWithLog struct {
```
JeyJeyGao, line 28:
```suggestion
type CacheWithLog struct {
```
Is it simpler since the package is called crl?

internal/crl/crl.go
```diff
@@ -0,0 +1,72 @@
...
+	logger := log.GetLogger(ctx)
+
+	bundle, err := c.Cache.Get(ctx, url)
+	if err != nil && !errors.Is(err, corecrl.ErrCacheMiss) {
+		if c.DiscardCacheError {
+			c.logDiscardCrlCacheErrorOnce.Do(c.logDiscardCrlCacheError)
+		}
+		logger.Debug(err.Error())
+	}
+	return bundle, err
```
JeyJeyGao, line 51:
```suggestion
	if err != nil && !errors.Is(err, corecrl.ErrCacheMiss) {
		if c.DiscardCacheError {
			c.logDiscardCrlCacheErrorOnce.Do(c.logDiscardCrlCacheError)
		}
		logger.Debug(err.Error())
		return nil, err
	}
	return bundle, nil
```
nit

internal/crl/crl.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+		logger.Debug(err.Error())
+	}
+	return err
+}
+
+// logDiscardCrlCacheError logs the warning when CRL cache error is
+// discarded
+func (c *CrlCacheWithLog) logDiscardCrlCacheError() {
+	fmt.Fprintln(os.Stderr, "Warning: CRL cache error discarded")
```
JeyJeyGao, line 71: Should we mention that enabling the debug flag will show detailed errors?

internal/crl/crl.go
```diff
@@ -0,0 +1,72 @@
...
+	// logDiscardCrlCacheErrorOnce guarantees the discard cache error
+	// warning is logged only once
+	logDiscardCrlCacheErrorOnce sync.Once
+}
+
+// Get retrieves the CRL bundle with the given url
+func (c *CrlCacheWithLog) Get(ctx context.Context, url string) (*corecrl.Bundle, error) {
+	logger := log.GetLogger(ctx)
+
+	bundle, err := c.Cache.Get(ctx, url)
```
JeyJeyGao, line 44: If the `c.Cache` is nil, will it trigger a panic? How will the user know that they need to set the Cache of `CrlCacheWithLog`? Do we need a `NewCacheWithLog` function and make the Cache a required argument?

internal/crl/crl.go
Two-Hearts: I think that's a good idea. Let me add it.

internal/crl/crl.go
```diff
@@ -0,0 +1,72 @@
...
+	logger := log.GetLogger(ctx)
+
+	err := c.Cache.Set(ctx, url, bundle)
+	if err != nil {
+		if c.DiscardCacheError {
+			c.logDiscardCrlCacheErrorOnce.Do(c.logDiscardCrlCacheError)
+		}
+		logger.Debug(err.Error())
+	}
+	return err
```
JeyJeyGao, line 65: We may need to return `err` in the block that actually has an error and return `nil` at the end to make the code clearer.

internal/crl/crl.go
Two-Hearts: Added sanity check for `Get` and `Set` methods.

internal/crl/crl.go
```diff
@@ -0,0 +1,80 @@
...
+
+	bundle, err := c.Cache.Get(ctx, url)
+	if err != nil && !errors.Is(err, corecrl.ErrCacheMiss) {
+		if c.DiscardCacheError {
+			c.logDiscardCrlCacheErrorOnce.Do(c.logDiscardCrlCacheError)
+		}
+		logger.Debug(err.Error())
+		return nil, err
+	}
+	return bundle, err
```
JeyJeyGao, line 55:
```suggestion
	return bundle, nil
```

Two-Hearts: This doesn't seem to be correct? What happens when error is `cach miss`? when cache miss, the bundle should be nil, because it's not exist in the cache. In notation-core-go, if err is nil, it directly uses the bundle to check expiry which panics.

JeyJeyGao (approved):
LGTM

internal/crl/crl.go
```diff
@@ -0,0 +1,80 @@
...
+		logger.Debug(err.Error())
+		return err
+	}
+	return nil
+}
+
+// logDiscardCrlCacheError logs the warning when CRL cache error is
+// discarded
+func (c *CacheWithLog) logDiscardCrlCacheError() {
+	fmt.Fprintln(os.Stderr, "Warning: CRL cache error discarded. Enable debug log through '-d' for error details.")
```
shizhMSFT, line 79: Do you want to print this warning through the logger?

shizhMSFT (approved):
LGTM with suggestions

internal/crl/crl.go
Two-Hearts: Yeah, I think both would work. I have a slightly preferring to follow the pattern of warnings in the cli though: https://github.com/search?q=repo%3Anotaryproject%2Fnotation+os.Stderr&type=code

priteshbandi (approved):
LGTM. Thanks for completing this :)

PR #1094, opened 2024-12-02
fix: add timestamping cert chain revocation check during signing

Two-Hearts:
This PR adds tsa cert chain revocation check after signing to fail fast.


internal/revocation/revocation.go
JeyJeyGao, line 1: We will have both `internal/revocation` and `internal/crl` folders. Is it better to reorganize the folders and move the `crl` package to `internal/revocation/crl`?

Two-Hearts: Good idea!

JeyJeyGao (approved):
LGTM

shizhMSFT (approved):
LGTM

rgnote (approved):
LGTM

PR #1133, opened 2025-01-02
feat: `blob inspect` command

JeyJeyGao:
Feat:
- added `notation blob inspect` command
- added `signature envelope type` for `inspect command` to be consistent with `blob inspect command` spec

Fix:
- removed repeated `blob.Cmd()`

Fix docs:
- updated the command spec docs

Refactor:
- refactor `inspect handler` related code to be shared by `blob inspect handler`

Test:
- added E2E

Output examples
Text format:
```
/home/jj/download/LICENSE.jws.sig
├── signature algorithm: RSASSA-PSS-SHA-256
├── signature envelope type: application/jose+json
├── signed attributes
│   ├── content type: application/vnd.cncf.notary.payload.v1+json
│   ├── signing scheme: notary.x509
│   └── signing time: Tue Jan  7 08:42:43 2025
├── user defined attributes
│   └── (empty)
├── unsigned attributes
│   ├── signing agent: notation-go/1.3.0+unreleased
│   └── timestamp signature
│       ├── timestamp: [Tue Jan  7 08:42:43 2025, Tue Jan  7 08:42:44 2025]
│       └── certificates
│           ├── SHA256 fingerprint: 36e731cfa9bfd69dafb643809f6dec500902f7197daeaad86ea0159a2268a2b8
│           │   ├── issued to: CN=Microsoft Public RSA Timestamping CA 2020,O=Microsoft Corporation,C=US
│           │   ├── issued by: CN=Microsoft Identity Verification Root Certificate Authority 2020,O=Microsoft Corporation,C=US
│           │   └── expiry: Mon Nov 19 20:42:31 2035
│           └── SHA256 fingerprint: 3403d75002d22e2b8c49a8a113957d9eb225c901b946837fd61ff3ce32c51f65
│               ├── issued to: CN=Microsoft Public RSA Time Stamping Authority,OU=Microsoft America Operations+OU=Thales TSS ESN:45D6-96C5-5E63,O=Microsoft Corporation,L=Redmond,ST=Washington,C=US
│               ├── issued by: CN=Microsoft Public RSA Timestamping CA 2020,O=Microsoft Corporation,C=US
│               └── expiry: Sat Feb 15 20:35:56 2025
├── certificates
│   └── SHA256 fingerprint: 3678adce9daa3a82f4f55fd65e0c87c398b3d9bcd5338c06bbf8850df8c6641d
│       ├── issued to: CN=testcert3,O=Notary,L=Seattle,ST=WA,C=US
[...]

PR #1152, opened 2025-01-20
fix: remove repeated blob command

JeyJeyGao:


shizhMSFT (approved):
LGTM

cmd/notation/main.go
```diff
@@ -52,7 +52,6 @@ func main() {
 		},
 	}
 	cmd.AddCommand(
-		blob.Cmd(),
```
Two-Hearts, line 55: This fix should be in PR: https://github.com/notaryproject/notation/pull/1133, please have the fix there.

Two-Hearts (changes requested):
I'm actually seeing this https://github.com/notaryproject/notation/pull/1133#discussion_r1923004904 in #1133.

cmd/notation/main.go
JeyJeyGao: This is a minor fix. Either way is fine. Let's fix it in #1133.

JeyJeyGao:
As discussed with @Two-Hearts . I will include the fix in #1133

PR #1156, opened 2025-01-21
docs: fix inspect command spec

JeyJeyGao:
Fix:
- For `text` output, renamed `io.cncf.notary.signingAgent` to `signing agent` to be consistent with the JSON output `signingAgent`.
- For `json` output, rendered timestamp in RFC 3339 with nanoseconds. Timestamp [RFC 3161](https://www.rfc-editor.org/rfc/rfc3161#section-2.4.2) can have a fraction-of-second time value.

### DIFF
--- a/specs/commandline/inspect.md
+++ b/specs/commandline/inspect.md
@@ -78,14 +78,14 @@ localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac4efe37a5380ee
     │   ├── user defined attributes
     │   │   └── io.wabbit-networks.buildId: 123                             //user defined metadata
     │   ├── unsigned attributes
-    │   │   ├── timestamp signature                                         //TSA response
-    |   │   │   ├── timestamp: [Fri Jun 23 22:04:31 2023, Fri Jun 23 22:04:31 2023]
-    |   │   │   └── certificates
-    |   │   │       └── SHA256 fingerprint: d2f6e46ded7422ccd1d440576841366f828ada559aae3316af4d1a9ad40c7828
-    |   │   │           ├── issued to: wabbit-com Software Timestamp
-    |   │   │           ├── issued by: wabbit-com Software Trusted Timestamping
-    |   │   │           └── expiry: Fri Oct 13 23:59:59 2034
-    │   │   └── io.cncf.notary.signingAgent: notation/1.0.0                 //client version
+    │   │   ├── signing agent: notation/1.0.0                               //client version
+    │   │   └── timestamp signature                                         //TSA response
+    |   │       ├── timestamp: [Fri Jun 23 22:04:31 2023, Fri Jun 23 22:04:31 2023]
+    |   │       └── certificates
+    |   │           └── SHA256 fingerprint: d2f6e46ded7422ccd1d440576841366f828ada559aae3316af4d1a9ad40c7828
+    |   │               ├── issued to: wabbit-com Software Timestamp
+    |   │               ├── issued by: wabbit-com Software Trusted Timestamping
+    |   │               └── expiry: Fri Oct 13 23:59:59 2034
     │   ├── certificates
     │   │   ├── SHA256 fingerprint: E8C15B4C98AD91E051EE5AF5F524A8729050B2A
     │   │   │   ├── issued to: wabbit-com Software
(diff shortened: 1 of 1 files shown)


specs/commandline/inspect.md
Two-Hearts, line 1: Let's make this change in PR #1150. It would be much easier to review the changes if they are in the same PR.
Logically, we should update the spec before any implementation.

JeyJeyGao: I think it is better to be an isolated PR for PMs to review. Yes, #1150 depends on this PR.

Two-Hearts: Please combine them into one PR this time. 
Next time, we should create and discuss the spec changes first. Any implementation change shall be after the spec discussion.

shizhMSFT (approved):
LGTM

specs/commandline/inspect.md
JeyJeyGao: #1150 has been converted to be a draft. Let's merge #1156 first.

specs/commandline/inspect.md
```diff
@@ -78,14 +78,14 @@ localhost:5000/net-monitor@sha256:b94d27b9934d3e08a52e52d7da7dabfac4efe37a5380ee
...
-    │   │   └── io.cncf.notary.signingAgent: notation/1.0.0                 //client version
+    │   │   ├── signing agent: notation/1.0.0                               //client version
+    │   │   └── timestamp signature                                         //TSA response
+    │   │       ├── timestamp: [Fri Jun 23 22:04:31 2023, Fri Jun 23 22:04:31 2023]
+    │   │       └── certificates
+    │   │           └── SHA256 fingerprint: d2f6e46ded7422ccd1d440576841366f828ada559aae3316af4d1a9ad40c7828
+    │   │               ├── issued to: wabbit-com Software Timestamp
+    │   │               ├── issued by: wabbit-com Software Trusted Timestamping
+    │   │               └── expiry: Fri Oct 13 23:59:59 2034
     │   ├── certificates
```
Two-Hearts, line 89: Since we decide to update the spec of `inspect` command, it's a good chance to add the display of `io.cncf.notary.x509chain.thumbprint#S256` (https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#oci-signatures) in the output.
It's been a pain point for a while, and we have received quite a few Notation users asking for this information. This is true for both `notation inspect` command and `notation blob inspect` command. (The latter can be done in the blob related PR.)

JeyJeyGao: Yes, we can have another PR for this new feature request.

Two-Hearts: Since you already created a PR for the spec change. Why not include all changes related to `inspect` command here? Besides, it's just a one-line change in the output.

Two-Hearts: Discussed offline, we could save it in the blob inspect command spec.

Two-Hearts (approved):
LGTM

specs/commandline/inspect.md
```diff
@@ -169,14 +169,14 @@ localhost:5000/net-monitor@sha256:ca5427b5567d3e06a72e52d7da7dabfac484efe37a5380
...
-    |   │   │   ├── timestamp: [Fri Jun 23 22:04:31 2023, Fri Jun 23 22:04:31 2023]
-    |   │   │   └── certificates
-    |   │   │       └── SHA256 fingerprint: d2f6e46ded7422ccd1d440576841366f828ada559aae3316af4d1a9ad40c7828
-    |   │   │           ├── issued to: wabbit-com Software Timestamp
-    |   │   │           ├── issued by: wabbit-com Software Trusted Timestamping
-    |   │   │           └── expiry: Fri Oct 13 23:59:59 2034
-    │   │   └── io.cncf.notary.signingAgent: notation/1.0.0
+    │   │   ├── signing agent: notation/1.0.0                               //client version
+    │   │   └── timestamp signature
+    │   │       ├── timestamp: [Fri Jun 23 22:04:31 2023, Fri Jun 23 22:04:31 2023]
```
FeynmanZhou, line 174: I am wondering why this timestamp does not uses nanosecond but only timestamp under unsignedAttributes?

JeyJeyGao: `unsignedAttributes` is for JSON output. In `text` output, it is `unsigned attributes`. `timestamp` is under unsigned attributes, as this is a text result for humans, so nanoseconds may not be helpful. It follows a predefined format called `ANSIC`, which doesn't include nanoseconds.

FeynmanZhou (approved):
LGTM

vaninrao10 (approved):
LGTM

PR #1180, opened 2025-02-18
docs: add the proposal for Notation blob signing and verification

yizha1:
Add the proposal for https://github.com/notaryproject/notation/issues/741

### DIFF
--- a/specs/proposals/blob-signing.md
+++ b/specs/proposals/blob-signing.md
@@ -105,20 +105,20 @@ For registry-based distribution, such as using an OCI-compliant container regist
 1. Sign a file on filesystem with a specified signature envelope:
 
     ```shell
-    notation blob sign sbom.json --signature-format cose --id myKeyId --plugin myKMSPluginName
+    notation blob sign --id myKeyId --signature-format "cose" --media-type "application/spdx+json" sbom.json
     ```
 
-    This generates a signature named `sbom.json.cose.sig` in the same directory as `sbom.json`.
+    This generates a signature named `sbom.json.cose.sig` in the current working directory.
 
 2. Publish the file and signature to the OCI-compliant registry:
 
+    Use ORAS tool to push the blob file:
+
     ```shell
-    notation blob push --artifact-type "application/spdx+json" sbom.json:"application/json" --signature sbom.json.cose.sig --reference myregistry/mypath/mysboms:v1
+    oras push myregistry/myapp/sbom:v1 --artifact-type "application/spdx+json" sbom.json:application/json
     ```
-
-    This command stores `sbom.json` as an OCI image referenced by `myregistry/mypath/mysboms:v1` in the registry, which has the artifact type `application/spdx+json` and `sbom.json` data stored as a layer with the media type `application/json`. Additionally, this command stores the signature file `sbom.json.cose.sig` as another OCI image with the artifact type `application/vnd.cncf.notary.signature` and the signature data as a layer with media type `application/cose`. The artifact type should be `application/vnd.cncf.notary.signature` as it is Notary Project signature. The layer media type can be detected according to the file name, which is either `application/cose` or `application/jws`. The signature manifest's `subject` property refers to the `myregistry/mypath/mysboms:v1`. To distinguish these signatures from those generated during OCI artifact signing, the specific annotation `io.cncf.notary.blob.signature=true` is added for the signature manifest.
-
-    Below is an example of the manifest for the reference `myregistry/mypath/mysboms:v1`:
+    
+    This command stores `sbom.json` as an OCI artifact referenced by `myregistry/mypath/mysboms:v1` in the registry, which has the artifact type `application/spdx+json` and `sbom.json` data stored as a layer with the media type `application/json`. Below is an example of the manifest for the reference `myregistry/mypath/mysboms:v1`:
 
     ```json  
     {
@@ -147,6 +147,19 @@ For registry-based distribution, such as using an OCI-compliant container regist
     }
     ```
 
+    Use ORAS tool to attach the signature to the blob file:
+
+    ```shell
+    oras attach myregistry/myapp/sbom:v1 --artifact-type 'application/vnd.cncf.notary.signature' sbom.json.cose.sig:application/cose -a "io.cncf.notary.blob.signature=true"
+    ```
+     
+    This command stores the signature file `sbom.json.cose.sig` as another OCI artifact with the artifact type `application/vnd.cncf.notary.signature` as is Notary Project signature, and the signature data as a layer with media type `application/cose` to show it is a COSE signature. The layer media type can be detected according to the file name, which is either `application/cose` or `application/jws`. The signature manifest's `subject` property refers to the `myregistry/mypath/mysboms:v1`. 
+    
+    To distinguish these signatures from those generated during signing an artifact in the OCI registry, the specific annotation `io.cncf.notary.blob.signature=true` is added for the signature manifest. For example, users can push the SBOM file to the OCI registry first and then use `notation sign myregistry/myapp/sbom:v1`, which generates a signature attached to the SBOM in the registry, but this signature is not in the same format as the blob signature.
+    
+    > [!NOTE]
+    > While users can push both the blob and the signature file as a single OCI artifact, this is not recommended. Since there may already be existing blobs stored in the registry, you only need to attach the signature to the SBOM. Additionally, this approach is not practical if you need to sign the SBOM with multiple signatures using different keys at a later stage.
+
     Below is an example of the manifest for the blob signature in the registry:
 
     ```json
@@ -162,12 +175,12 @@ For registry-based distribution, such as using an OCI-compliant container regist
         },
         "layers": [
             {
-            "mediaType": "application/cose",
-            "digest": "sha256:6dc2a336b124d625f87deb3f539fa1465414ddad300a898adf4682e530d6f144",
-            "size": 1479,
-            "annotations": {
-                "org.opencontainers.image.title": "sbom.json.cose.sig"
-            }
+                "mediaType": "application/cose",
+                "digest": "sha256:6dc2a336b124d625f87deb3f539fa1465414ddad300a898adf4682e530d6f144",
+                "size": 1479,
+                "annotations": {
+                    "org.opencontainers.image.title": "sbom.json.cose.sig"
+                }
             }
         ],
         "subject": {
@@ -177,31 +190,45 @@ For registry-based distribution, such as using an OCI-compliant container regist
         },
         "annotations": {
             "io.cncf.notary.blob.signature": "true",
-            "io.cncf.notary.x509chain.thumbprint#S256": "[\"aaaa\", \"bbbb\"]",
             "org.opencontainers.image.created": "2025-02-18T06:08:45Z"
         }
     }
     ```
 
-    Alternatively, you can also use ORAS tool to push the blob file first, and then use Notation CLI command to push only the signature:
+3. Discover signatures in OCI registries
 
-    ```shell
-    oras push myregistry/mypath/mysboms:v1 sbom.json:application/octet-stream 
-    notation blob push --signature sbom.json.cose.sig --reference myregistry/mypath/mysboms:v1
-    ```
+   Use the `notation ls` command to discover the blob signature attached to a blob in the OCI registry. Since there may be OCI signatures generated for an OCI artifact, the `notation ls` command need to display the annotation `io.cncf.notary.blob.signature=true`, if present, to indicate that it is a blob signature. For example,
 
-    You can also create an oci image layout and then use ORAS tool to copy it to the target registry. 
+   ```shell
+   notation ls myregistry/myapp/sbom:v1
+   ```
 
-    ```shell
-    notation blob push --artifact-type "application/spdx+json" sbom.json:"application/json" --signature sbom.json.cose.sig --oci-layout sbom.json:v1
-    oras copy -r --from-oci-layout sbom.json:v1 myregistry/mypath/mysboms:v1
-    ```
+   Sample output:
 
-    > NOTE:
-    >
-    > Support for OCI image layout in Notation is currently an experimental feature.
+   ```text
+   myregistry/myapp/sbom@sha256:3472cf1b816e5615a5f1883bbdb8c5b91e3a11cd7a566d323bd429d72c667f2b
+   └── application/vnd.cncf.notary.signature
+       └── sha256:e8c3567e1f679dfd372ab97635329e404c5182864e29abd96acf852b1ac1ef6b
+           └── io.cncf.notary.blob.signature: "true"      
+   ```
 
-3. Verify the signature from the registry:
+   Similarly, users can use the ORAS tool to discover signatures. However, the additional `-v` flag should be used to list annotations. Unlike notation ls, the ORAS tool lists all annotations present in the manifest. For example,
+
+   ```shell
+   oras discover myregistry/myapp/sbom:v1
+   ```
+
+   Sample output:
+
+   ```text
+   myregistry/myapp/sbom@sha256:3472cf1b816e5615a5f1883bbdb8c5b91e3a11cd7a566d323bd429d72c667f2b
+   └── application/vnd.cncf.notary.signature
+       └── sha256:e8c3567e1f679dfd372ab97635329e404c5182864e29abd96acf852b1ac1ef6b
+           └── io.cncf.notary.blob.signature: "true"
+           └── org.opencontainers.image.created: "2025-02-28T18:06:08Z"
+   ```
+
+4. Verify the signature from the registry:
 
     - Add the root CA certificate and set up the trust store:
 
@@ -221,11 +248,24 @@ For registry-based distribution, such as using an OCI-compliant container regist
         notation blob policy init --name "myBlobPolicy" --trust-store "ca:myCACerts" --trust-identity "x509.subject:C=US,ST=WA,O=wabbit-network.io"
         ```
 
+    - Download the blob and signature using ORAS tool:
+
+      Follow the previous section to discover blog signatures, and get the digest of the siganture that you want to pull, and use the following command to pull the signature and the blob:
+
+        ```shell
+        oras pull --include-subject myregistry/myapp/sbom@sha256:e8c3567e1f679dfd372ab97635329e404c5182864e29abd96acf852b1ac1ef6b
+        ```
+    
+      This command pulls both the `sbom.json` and `sbom.json.cose.sig` to your filesystem.
+
     - Verify the signature:
 
         ```shell
-        notation blob verify --reference myregistry/mypath/mysboms:v1 --policy-name myBlobPolicy
+        notation blob verify --policy-name myBlobPolicy --signature sbom.json.cose.sig sbom.json
         ```
+
+    > [!NOITE]
+    > In case users use `notation verify` command to verify blob sigantures, the verification should fail, but should indicate to users that they should try download the blob and signatures and using `notation blob verify` command.
 
 ### Manage blob policies
 
@@ -240,13 +280,13 @@ The following commands are available for managing blob policies:
 - Overwrite an existing policy with a prompt:
 
     ```shell
-    notation blob policy init --trust-store "ca:myCACerts" --trust-identity "x509.subject:C=US,ST=WA,O=wabbit-network.io"
+    notation blob policy init --name "myBlobPolicy" --trust-store "ca:myCACerts" --trust-identity "x509.subject:C=US,ST=WA,O=wabbit-network.io"
     ```
 
 - Overwrite an existing policy with a prompt using the flag `--force`:
 
     ```shell
-    notation blob policy init --force --trust-store "ca:myCACerts" --trust-identity "x509.subject:C=US,ST=WA,O=wabbit-network.io"
+    notation blob policy init --force --name "myBlobPolicy" --trust-store "ca:myCACerts" --trust-identity "x509.subject:C=US,ST=WA,O=wabbit-network.io"
     ```
 
 - Show the blob policy:
--- a/specs/proposals/blob-signing.md
+++ b/specs/proposals/blob-signing.md
@@ -1,0 +1,282 @@
+# Proposal for Supporting Blob Signing and Verification in Notation
+
+## Overview 
+
+A **blob**, short for [Binary Large Object](https://wikipedia.org/wiki/Binary_large_object) is a collection of binary data stored as a single entity. In this proposal, a blob refers to any binary data or file, including SBOM (Software Bill of Materials) files, release assets, AI model files, WebAssembly files or other forms of unstructured data. While Notation currently supports signing and verifying OCI (Open Container Initiative) artifacts such as container images, this document outlines a proposal to extend the Notation tool's capabilities to include blob signing and verification.
+
+## Problem Statement & Motivation 
+
+Signing container images alone is not enough to secure the software supply chain for containers. In addtion to images, SBOMs, configuration files, release binaries or other artifacts must also be signed and verified to ensure their integrity and authenticity. With the increasing use of AI model files and WebAssembly files, securing these artifacts is equally critical. By adding blob signing, Notation can extend its capabilities to handle a broader range of artifacts across various scenarios, enhancing its versatility while leveraging its existing key management and trust policy framework.
+
+## Scenarios 
+
+The following describes how blob signing and verification can be used for various scenarios:
+
+### Scenario 1: Blob signing and verification with file-based distribution
+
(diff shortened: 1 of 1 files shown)


specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,282 @@
...
+        notation cert ls
+        ```
+
+        > NOTE
+        >
+        > Learn more options using `notation cert ls --help`.
+    - Set up trust policy for blobs by adding a new command `notation blob policy init`. This command streamlines the process, eliminating the need for users to consult documentation for the correct trust policy format and preventing the accidental use of policies intended for other verification purposes.
+
+        ```shell
+        notation blob policy init --name "myBlobPolicy" --trust-store "ca:myCACerts" --trust-identity "x509.subject:C=US,ST=WA,O=wabbit-network.io"
```
JeyJeyGao, line 76: I think the verification level is `strict` and GlobalPolicy is true when using blob policy init command, right?

specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,282 @@
...
+
+     Confirm the certificate is added.
+
+        ```shell
+        notation cert ls
+        ```
+
+        > NOTE
+        >
+        > Learn more options using `notation cert ls --help`.
```
Two-Hearts, line 72: nit:
```suggestion
        > [!NOTE]
        > Learn more options using `notation cert ls --help`.
```

specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,282 @@
...
+        },
+        "annotations": {
+            "io.cncf.notary.blob.signature": "true",
+            "io.cncf.notary.x509chain.thumbprint#S256": "[\"aaaa\", \"bbbb\"]",
+            "org.opencontainers.image.created": "2025-02-18T06:08:45Z"
+        }
+    }
+    ```
+
+    Alternatively, you can also use ORAS tool to push the blob file first, and then use Notation CLI command to push only the signature:
```
Two-Hearts, line 182:
```suggestion
    Alternatively, you can also use other tools such as ORAS to push the blob file first, and then use Notation CLI command to push the signature:
```

specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,282 @@
...
+
+    ```shell
+    oras push myregistry/mypath/mysboms:v1 sbom.json:application/octet-stream 
+    notation blob push --signature sbom.json.cose.sig --reference myregistry/mypath/mysboms:v1
+    ```
+
+    You can also create an oci image layout and then use ORAS tool to copy it to the target registry.
```
Two-Hearts, line 189: Since oci image layout is an experimental feature in Notation, I think we should remove it from this doc for now.

specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,282 @@
...
+
+**Steps:**
+
+1. Sign a file on filesystem with a specified signature envelope:
+
+    ```shell
+    notation blob sign --signature-format cose sbom.json
+    ```
+
+    This generates a signature named `sbom.json.cose.sig` in the same directory as `sbom.json`. The signature file name follows the [Notary Project specification](https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#blob-signatures).
```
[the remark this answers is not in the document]
yizha1: No, we will not change the current behavior. I will update the text.

specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,282 @@
...
+
+        ```shell
+        notation blob policy init --name "myBlobPolicy" --trust-store "ca:myCACerts" --trust-identity "x509.subject:C=US,ST=WA,O=wabbit-network.io"
+        ```
+
+        > NOTE
+        >
+        > See the [section](#manage-blob-policies) for more commands for `notation blob policy`.
+        >
+        > This `init` command can be extended to policy commands for OCI artifacts.
```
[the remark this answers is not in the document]
yizha1: I don't have strong opinion. I can remove it.

specs/proposals/blob-signing.md
yizha1: I would like to keep it to avoid missing this case in the future.

specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,282 @@
...
+
+    ```shell
+    notation blob push --artifact-type "application/spdx+json" sbom.json:"application/json" --signature sbom.json.cose.sig --reference myregistry/mypath/mysboms:v1
+    ```
+
+    This command stores `sbom.json` as an OCI image referenced by `myregistry/mypath/mysboms:v1` in the registry, which has the artifact type `application/spdx+json` and `sbom.json` data stored as a layer with the media type `application/json`. Additionally, this command stores the signature file `sbom.json.cose.sig` as another OCI image with the artifact type `application/vnd.cncf.notary.signature` and the signature data as a layer with media type `application/cose`. The artifact type should be `application/vnd.cncf.notary.signature` as it is Notary Project signature. The layer media type can be detected according to the file name, which is either `application/cose` or `application/jws`. The signature manifest's `subject` property refers to the `myregistry/mypath/mysboms:v1`. To distinguish these signatures from those generated during OCI artifact signing, the specific annotation `io.cncf.notary.blob.signature=true` is added for the signature manifest.
```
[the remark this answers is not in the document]
yizha1: We can discuss it. From users point of view, this will provide better experience. If users only want to push the blob, we should not support, but users want to push both blob and signature, this will make it easy for users.

specs/proposals/blob-signing.md
yizha1: Yes, other parameters will use the default values. Users can update them later according to the update guideline.

specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,282 @@
...
+        ```
+
+### Manage blob policies
+
+The following commands are available for managing blob policies:
+
+- Initialize blob policies:
+
+    ```shell
+    notation blob policy init --name "myBlobPolicy" --trust-store "ca:myCACerts" --trust-identity "x509.subject:C=US,ST=WA,O=wabbit-network.io"
```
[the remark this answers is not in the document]
yizha1: Yes, but users need to specify it clearly, we will not by defaut use `*`, which could be security risk. I will add a note in the document

specs/proposals/blob-signing.md
toddysm: How is the blob policy different from the standard Notation policy? Why do we need to have commands in `notation blob policy` and `notation policy`?

specs/proposals/blob-signing.md
toddysm, line 115: > This command stores `sbom.json` as an OCI image...

Do you mean _OCI image_ or _OCI artifact_?

specs/proposals/blob-signing.md
toddysm, line 182: As this only requires pushing artifacts to the registry, can we have both, the SBOM and the signature, pushed by ORAS? From experience point of view, mixing the tools degrades the experience. So, thinking about the experience, Notation is for "signing", so, it is used to sign the artifact (SBOM). ORAS is for pushing and pulling artifacts, so, it is used to push both, the SBOM and the signaturel.

specs/proposals/blob-signing.md
toddysm, line 189: Why does it need to be OCI layout?

specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,282 @@
...
+        Confirm that the cert is added.
+
+        ```shell
+        notation cert ls
+        ```
+
+    - Set up the trust policy:
+
+        ```shell
+        notation blob policy init --name "myBlobPolicy" --trust-store "ca:myCACerts" --trust-identity "x509.subject:C=US,ST=WA,O=wabbit-network.io"
```
toddysm, line 217: Again, why do we need to have `policy` command and `blob policy` subcommand? This is suboptimal experience IMHO

specs/proposals/blob-signing.md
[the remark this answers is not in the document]
Two-Hearts: > Can we have default option and overwrite? For example, somebody may want to store in specific folder.

Yes, we do have this overwrite option. It's just the default location that needs to be discussed here.

specs/proposals/blob-signing.md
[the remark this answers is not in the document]
Two-Hearts: @shizhMSFT @FeynmanZhou @JeyJeyGao Looking for your insights here.

specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,282 @@
+# Proposal for Supporting Blob Signing and Verification in Notation
+
+## Overview 
+
+A **blob**, short for [Binary Large Object](https://wikipedia.org/wiki/Binary_large_object) is a collection of binary data stored as a single entity. In this proposal, a blob refers to any binary data or file, including SBOM (Software Bill of Materials) files, release assets, AI model files, WebAssembly files or other forms of unstructured data. While Notation currently supports signing and verifying OCI (Open Container Initiative) artifacts such as container images, this document outlines a proposal to extend the Notation tool's capabilities to include blob signing and verification.
```
shizhMSFT, line 5: Just some notes for AI models:
- **Ollama** --> OCI-like Artifacts
- **LM Studio** --> Hugging Face (individual files)

specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,282 @@
...
+
+## Scenarios 
+
+The following describes how blob signing and verification can be used for various scenarios:
+
+### Scenario 1: Blob signing and verification with file-based distribution
+
+Sarah, a software engineer, creates binaries for applications that are distributed via her company's website. In CI/CD pipelines, she creates tasks that generates an SBOM for each binary and signs the SBOM to ensure the authenticity and integrity of the SBOM. The pipeline automatically generates a digital signature for each SBOM, which is stored alongside the SBOM on the filesystem. With SBOM and signatures created, the pipeline tasks automatically upload the binaries, SBOMs, and SBOM signatures to the company website as separate downloadable artifacts. The compliance team downloads the SBOM and signature to check whether the corresponding binary is compliant with both company and governance rules. Before analyzing the SBOM, they verify its integrity against the corresponding signature, ensuring the SBOM hasn't been tampered with since being signed in the pipeline and that it originates from the Sarah's company.
```
shizhMSFT, line 17: nit: just curious why Sarah :D

Usually,
- A --> Alice
- B --> Bob
- C --> Charlie
- D --> David

Basically, those names are used to replace User A, User B, User C, bla bla...

specs/proposals/blob-signing.md
shizhMSFT, line 50: Please use a versioned link or a link with commit hash.

specs/proposals/blob-signing.md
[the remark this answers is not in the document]
shizhMSFT: I agree with @Two-Hearts

specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,282 @@
...
+        > This `init` command can be extended to policy commands for OCI artifacts.
+    - Verify the signature:
+
+        ```shell
+        notation blob verify --policy-name myBlobPolicy --signature sbom.json.cose.sig sbom.json
+        ```
+
+        If a [global policy](https://github.com/notaryproject/specifications/blob/main/specs/trust-store-trust-policy.md#blob-trust-policy) is set, you can skip the `--policy-name` flag.
+
+### Blob Sign and verify with registry-based distribution
```
shizhMSFT, line 92: I have a fundamental question: What the difference and advantage of "Blob Sign and verify with registry-based distribution", comparing with the following workflow?
```sh
# sign
notation blob sign --signature-format cose sbom.json
oras push myregistry/mypath/mysboms:v1 sbom.json sbom.json.cose.sig

# verify
oras pull myregistry/mypath/mysboms:v1
notation blob verify --policy-name myBlobPolicy --signature sbom.json.cose.sig sbom.json
```
Basically, pack the sbom with its signature as a single OCI artifact

Why I'm asking since there will only be one signature for a blob for the most cases. Registries are one of mediums just like the file system.

shizhMSFT: The only advantage of the registries is that the registry may have the digest pre-computed in the manifest. However, the digest algorithm may not match the digest algorithm in the signature payload. /cc @Two-Hearts

specs/proposals/blob-signing.md
yizha1: It was generated by AI :)

specs/proposals/blob-signing.md
yizha1: @toddysm It is according to the spec https://github.com/notaryproject/specifications/blob/main/specs/trust-store-trust-policy.md#trust-policy-properties, which has been discussed and aligned within the community when AWS folks worked on this feature. Blob will use a slight different trust policy format in a different trust policy config file. So to make the experience simple for users, the new command `init` will help users to intialize a trust policy easily with default secure settings. Later on, it can be expanded to `notation policy init` command for OCI artifact, so that users do not need to do any manual edit for default config.

specs/proposals/blob-signing.md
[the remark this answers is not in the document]
yizha1: @toddysm This document provides an overview of commands, while additional flags will be detailed in the `notation blob sign` command specification. Each command's specification will include complete flags and usages.

specs/proposals/blob-signing.md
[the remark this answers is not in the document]
yizha1: Agree. I will keep only pushing signatures

specs/proposals/blob-signing.md
yizha1: It seems the blob spec was not released, so it cannot use versioned link.

specs/proposals/blob-signing.md
Two-Hearts: This is a question on the use case.

@yizha1 Do we want to support the following scenario:
Suppose I have a file signed in my local filesystem, and I used oras to push both the file and signature to the registry. For some reason, I want to re-sign the file. What I need to do is `notation blob sign` -> `oras push` (since the file and the first signature was packed as a OCI image instead of a referrer, I need a second image to verify the second signature). The verifier would then need to do `oras pull` all applicable signatures-> `notation blob verify` for each pulled signature.
If we have the design suggested by @yizha1, then the workflow becomes:
`notation blob sign` again, and then `notation blob push`. On verification side, `notation blob verify`, Notation would verify every applicable signature on the fly.

If we only want to support one signature per blob file at any time point, then yes, both designs would do the job. But if as the time goes by, the number of signatures could increase, then the registry way is better. /cc: @shizhMSFT

specs/proposals/blob-signing.md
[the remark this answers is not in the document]
FeynmanZhou: Agree with @Two-Hearts  . Keeping the scope of notation focus on signature generation and managment makes more sense.

If there is a blob file has a signature and other attestations,  should notation supports pushing blob, its signature, and other attestations together? That may beyond the scope.

yizha1: It is an OCI artifact. I want to say they are stored per the OCI image spec. I will rephase it.

specs/proposals/blob-signing.md
yizha1: Manage signatures is with Notation's scope. So, considering the comments from Patrick and Shiwei, I will remove the function of using Notation to push both blob and signature, but keep `notation blob push` for pushing signatures. If users want to use oras to push signature as referrers, then the command looks like below:

```
oras attach <my_registry>/<my_repo>:v1 --artifact-type 'application/vnd.cncf.notary.signature' myFile.cose.sig:application/cose -a "io.cncf.notary.blob.signature=true" -a "io.cncf.notary.x509chain.thumbprint#S256=[\"aaaa\", \"bbbb\"]"
```

As you can see, it is not a good experience for users, epecially the thumbprint parameter. 

Shiwei's has a good comment, https://github.com/notaryproject/notation/pull/1180#discussion_r1961950312, if we agree with this, then we probably do not need `notation blob push`.

specs/proposals/blob-signing.md
yizha1: This is another opition. That users can create OCI layout for signature first and then copy to the registry.

specs/proposals/blob-signing.md
yizha1: See this comment https://github.com/notaryproject/notation/pull/1180#discussion_r1962855581

specs/proposals/blob-signing.md
yizha1: Thanks.

specs/proposals/blob-signing.md
yizha1: The blob spec was not released yet, so I cannot use a versioned link, will use a commit hash

specs/proposals/blob-signing.md
yizha1: Toddy and I discussed offline and agreed that the current proposal for managing trust policies for OCI artifacts and blobs is not optimal.  We should explore a better solution due to potential user confusion with the two policy commands and configuration files. /cc @shizhMSFT @Two-Hearts @JeyJeyGao

We also need to consider the release version, if we starts with v2, we may break the experience later on and require a major version stepping up. One option is that we could do v1.4.0 release first, and improve the experience and other command experience in v2 including trust policy format.

specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,282 @@
...
+
+## Proposal 
+
+This section outlines the proposed solution for using Notation CLI commands to sign and verify blobs. The following points are not covered in this document:
+- Detailed command usage, which will be covered in the individual command specifications at [Notation CLI Docs](https://github.com/notaryproject/notation/tree/v1.3.0/specs/commandline).
+- Distribution of certificates for verification across different scenarios.
+- Compatibility of the payload format. While some users are utilizing Notary Project-compliant signature envelopes, the payload format for blobs differs from the [Notary Project specification](https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#blob-payload).
```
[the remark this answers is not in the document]
yizha1: Discussed with Toddy, it is mainly about payload format of blob signature is different from other tools which generates a different payload format, for example, only digest is included as the payload format. I will start a new discusssion thread about the payload format, the goal is no matter which tool produces Notary Project signatures, it can be processed by Notation and other tools per the Notary Project signature specification. /cc: @shizhMSFT @Two-Hearts @JeyJeyGao

specs/proposals/blob-signing.md
yizha1: See the updated comment at https://github.com/notaryproject/notation/pull/1180#discussion_r1966433336

specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,282 @@
...
+    - Set up the trust policy:
+
+        ```shell
+        notation blob policy init --name "myBlobPolicy" --trust-store "ca:myCACerts" --trust-identity "x509.subject:C=US,ST=WA,O=wabbit-network.io"
+        ```
+
+    - Verify the signature:
+
+        ```shell
+        notation blob verify --reference myregistry/mypath/mysboms:v1 --policy-name myBlobPolicy
```
FeynmanZhou, line 223: If the blob and its signature have been pushed to the registry, how can an end-user differentiate between a blob and a normal OCI image in the registry? It determines which `notation` command to verify.

specs/proposals/blob-signing.md
Two-Hearts: And a potential security question regarding `oras pull` both the blob and the signature to local file system for verification:
Based on Notation's [threat model](https://github.com/notaryproject/specifications/blob/main/threatmodels/notation-threatmodel.md#notation-verify-artifacts-against-signatures), the local file system is within the trust boundary, but remote registry is not. Downloading the blob from a registry to local file system before doing any Notation verification can be risky?

specs/proposals/blob-signing.md
yizha1: Aligend after offline discussion. We will follow the existing [blob CLI spec](https://github.com/notaryproject/notation/blob/main/specs/commandline/blob.md) for implementing the commands.

specs/proposals/blob-signing.md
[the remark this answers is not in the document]
yizha1: Aligned offline that users will use oras to push blob file and blob signatures, and berfore verification, users need to download the blob file and signatures to filesystem and then use `notation blob verify` command for verification.

specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,282 @@
...
+            "org.opencontainers.image.created": "2025-02-18T06:08:45Z"
+        }
+    }
+    ```
+
+    Alternatively, you can also use ORAS tool to push the blob file first, and then use Notation CLI command to push only the signature:
+
+    ```shell
+    oras push myregistry/mypath/mysboms:v1 sbom.json:application/octet-stream 
+    notation blob push --signature sbom.json.cose.sig --reference myregistry/mypath/mysboms:v1
```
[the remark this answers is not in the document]
yizha1: `notation blob push` command will not be introduced based on alignment. See https://github.com/notaryproject/notation/pull/1180#discussion_r1976467426

specs/proposals/blob-signing.md
yizha1: Remove OCI image layout from the proposal. It will be brought back if there is a clear scenario.

specs/proposals/blob-signing.md
yizha1: Discussed with @Two-Hearts, the threat model will be tracked by the existing issue https://github.com/notaryproject/specifications/issues/296

specs/proposals/blob-signing.md
yizha1: Aligend on the comment https://github.com/notaryproject/notation/pull/1180#discussion_r1976467426

specs/proposals/blob-signing.md
yizha1: The blob signature manifest will have an annotation `io.cncf.notary.blob.signature: "true"`.

yizha1:
@toddysm @shizhMSFT @JeyJeyGao @Two-Hearts @FeynmanZhou I resolved comments based on our discussion. It is ready for another round of review. Thanks.

specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,335 @@
+# Proposal for Supporting Blob Signing and Verification in Notation
+
+## Overview 
+
+A **blob**, short for [Binary Large Object](https://wikipedia.org/wiki/Binary_large_object) is a collection of binary data stored as a single entity. In this proposal, a blob refers to any binary data or file, including SBOM (Software Bill of Materials) files, release assets, AI model files (such as model files on [Hugging Face](https://huggingface.co/)), WebAssembly files or other forms of unstructured data. While Notation currently supports signing and verifying OCI (Open Container Initiative) artifacts such as container images, this document outlines a proposal to extend the Notation tool's capabilities to include blob signing and verification.
```
shizhMSFT, line 5: Note: Technically, a file and a blob are different. A file has a header with create time, mod time, user, group, permission, etc. while a blob has only the content data.

specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,335 @@
...
+This section outlines the proposed solution for signing and verifying blobs using Notation CLI commands. The following topics are outside the scope of this document:  
+
+- **Detailed command usage**, which is covered in the individual command specifications at [Notation CLI specs](https://github.com/notaryproject/notation/tree/main/specs/commandline).  
+- **Distribution of certificates** for verification in different scenarios.  
+- **Blob signature definitions**, as outlined in the [Notary Project specification](https://github.com/notaryproject/specifications/blob/main/specs/signature-specification.md#blob-payload).  
+- **CI/CD integration**, as this document specifically focuses on using Notation CLI commands.  
+
+In general, all blob-related commands are grouped under the `notation blob` command per the [blob command spec](https://github.com/notaryproject/notation/blob/main/specs/commandline/blob.md).
+
+### Blob Sign and verify with file-based distribution
```
shizhMSFT, line 36: nit:
```suggestion
### Blob sign and verify with file-based distribution
```

specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,335 @@
...
+        ```
+
+     Confirm the certificate is added.
+
+        ```shell
+        notation cert ls
+        ```
+
+        > [!NOTE]
+        > Learn more options using `notation cert ls --help`.
```
shizhMSFT, line 71: nit: It seems those two lines are not rendered well on GitHub in a list. We probably need to reformat it for better rendering.

specs/proposals/blob-signing.md
```diff
@@ -0,0 +1,335 @@
...
+    ```
+
+    This generates a signature named `sbom.json.cose.sig` in the current working directory.
+
+2. Publish the file and signature to the OCI-compliant registry:
+
+    Use [ORAS tool](https://oras.land/) to push the blob file:
+
+    ```shell
+    oras push --artifact-type "application/spdx+json" sbom.json:application/json myregistry/myapp/sbom:v1
```
shizhMSFT, line 118: Can I also use `oras attach`?

yizha1: This command pushes the SBOM file, while `oras attach` is used later to attach the signature.

In this scenario, the SBOM is generated for a software application, not for a container image. Therefore, the example uses `oras push`, not to attach the SBOM to a container image in the registry.

yizha1:
@shizhMSFT @Two-Hearts @JeyJeyGao would you mind review it again?

shizhMSFT (approved):
LGTM

Two-Hearts (approved):
LGTM

PR #1181, opened 2025-02-18
bump: update go v1.24

JeyJeyGao:
Fix:
- fixed the [testable example](https://go.dev/blog/examples) function name, which is enforced by go v1.24

Bump:
- updated go 1.24

### DIFF
--- a/cmd/notation/internal/display/metadata/tree/tree_test.go
+++ b/cmd/notation/internal/display/metadata/tree/tree_test.go
@@ -48,7 +48,8 @@ func TestNodeAddPair(t *testing.T) {
 	}
 }
 
-func ExampleRootPrint() {
+// example for Print emthod of node type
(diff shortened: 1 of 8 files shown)


shizhMSFT (approved):
LGTM

cmd/notation/internal/display/metadata/tree/tree_test.go
Two-Hearts, line 1: This change should not be appeared in this PR.

Two-Hearts (changes requested):

Two-Hearts (approved):
LGTM

cmd/notation/internal/display/metadata/tree/tree_test.go
JeyJeyGao: The fix is required for Go v1.24; otherwise, the go test will fail.

Issue #1151, opened 2025-01-17
housekeeping: refactoring for sustainable coding


JeyJeyGao:
## command output refactoring
- [x] refactor inspect command output handler #1150 
- [x] refactor verify command output handler #1167 
- [x] refactor list command output handler #1176 
- [ ] refactor other commands handler

## Input flags code refactoring
- [ ] refactor common flags: #1195

JeyJeyGao:
As suggested by @shizhMSFT . We need to enhance the code https://github.com/notaryproject/notation/pull/1150/files#r1922159179

PR #1176, opened 2025-02-17
refactor: list output handler

JeyJeyGao:
Refactor:
- moved the output logic of `list` command to `ListHandler` to simplify list command code logic
- added `streamPrinter` to print tree node in streaming fashion

Test:
- improved E2E

Resolve part of #1151

PR #1183, opened 2025-02-19
refactor: output handler in table format

JeyJeyGao:
Resolve part of #1151

### DIFF
--- a/cmd/notation/cert/list.go
+++ b/cmd/notation/cert/list.go
@@ -16,18 +16,19 @@ package cert
 import (
 	"context"
 	"fmt"
-	"os"
 
 	"github.com/notaryproject/notation-go/dir"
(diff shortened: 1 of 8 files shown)


Two-Hearts (changes requested):
I feel like we are having several similar refactor PR's regarding the same `internal/display` dir such as #1176 . 
Can you group all such refactor works regarding `internal/display` into one PR so that I could review them systematically at once?

JeyJeyGao:
> I feel like we are having several similar refactor PR's regarding the same `internal/display` dir such as #1176 . Can you group all such refactor works regarding `internal/display` into one PR so that I could review them systematically at once?

#1176 is for tree output, but this PR is for table output, so they don't have a logical dependency. Also, #1176 is complex enough to be a single PR.

PR #1194, opened 2025-02-27
fix: output in streaming fashion for `inspect` command in `tree` format

JeyJeyGao:
The `inspect` command can now output in streaming fashion in `tree` format. Previously, the `inspect` command in `tree` format could only output when all signatures were read, which was not user-friendly when there were many signatures, and users had to wait for a while for the output.


shizhMSFT (approved):
LGTM

Two-Hearts (approved):
LGTM

PR #1195, opened 2025-02-27
chore: move common flags to `cmd` package

JeyJeyGao:
Moved `cmd/notation/common.go` to `cmd` package to concentrate all the flags and options related code in one place.


cmd/notation/internal/constant/env.go
shizhMSFT, line 1: Please leave it in the `cmd/notation/internal/option/options.go` file.

cmd/notation/internal/constant/env.go
```diff
@@ -0,0 +1,20 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package constant defines constants used in the notation CLI.
+package constant
+
+const (
+	UsernameEnv = "NOTATION_USERNAME"
+	PasswordEnv = "NOTATION_PASSWORD"
```
shizhMSFT, line 19: Better names for them should be `EnvironmentUsername` and `EnvironmentPassword`.

cmd/notation/internal/constant/env.go
JeyJeyGao: Updated.

cmd/notation/internal/constant/env.go
JeyJeyGao: Updated.

shizhMSFT (dismissed):
LGTM

cmd/notation/internal/option/flags.go
```diff
@@ -12,7 +12,7 @@
 // limitations under the License.
 
 // Package cmd contains common flags and routines for all CLIs.
-package cmd
+package option
```
Two-Hearts, line 15: Well... After reading through the PR, I personally prefer the previous structure, where `internal/cmd/flags.go` is clear enough for any developer to add a new flag in it.

From description at line 14: `// Package cmd contains common flags and routines for all CLIs.`
This description on package `option` leaves a deep concern, and my question is: Why do we have to keep creating new PRs to just move around internal files, renaming the package names with something else, while reducing readability of the code base and increasing developer mistakes like above? Is this really a MUST to do item?

Two-Hearts (changes requested):

cmd/notation/internal/option/flags.go
JeyJeyGao: There are two issues in the repo:
1. Flags and options are located in multiple places: the `internal/cmd` package and `cmd/notation/common.go`.
2. The `internal/cmd` package name is too broad. What does `cmd` mean? Everything we are working on is command-related.

The `option` package only includes flags and options that are used when defining a command opts struct, so I think this name is a better choice.

Two-Hearts: As I mentioned in the above comment, your description to package `option` is already wrong, I'm pretty sure this is a reduction of readability and increases developing difficulty for others.

If you really want this chore PR, my suggestion is:
Move flag related code from `common.go` to `internal/cmd/flags.go` and that's it.

JeyJeyGao: Updated.

cmd/notation/list_test.go
```diff
@@ -15,14 +15,17 @@ package main
 
 import (
 	"testing"
+
+	"github.com/notaryproject/notation/internal/cmd"
+	cmdopts "github.com/notaryproject/notation/internal/cmd"
```
shizhMSFT, line 20: They are the same?

shizhMSFT (approved):
LGTM with suggestions

cmd/notation/list_test.go
JeyJeyGao: Nice catch! Updated.

shizhMSFT (approved):
LGTM

Two-Hearts (approved):
LGTM

PR #1204, opened 2025-03-07
feat: output signature manifest reference after sign

### LINKED ISSUE #511: Allow user to specify that digest of signature should be returned from signing operations

Notation CLI

I'm signing container images and want to be able to refer to the signature artifact in the immediate next step of my workflow. The output of 'notation sign' today does not include this, and the only workaround I've found so far is to do some likely fragile parsing of the output of an 'oras discover' command. 

Flag that specifies only to return the location of the new artifact, verbosity level that only returns it, or some other method of assigning an environment variable. 

Parsing output of 'oras discover' 

### Any additional context?

Wish I had the output here.
image

Manage to parse it from here. Assuming edge cases exist to make this really more complicated and fragile though. 
image

### LINKED ISSUE #695: feat: Print out the signature digest when sign an artifact

Notation CLI

Notation does not print out the digest of the generated signature artifact when signing an OCI artifact using `notation sign` . In this case, users are not able to find the signature generated from the last step if there are multiple signatures associated with the signed OCI artifact.

```
$ notation sign $IMAGE
Successfully signed localhost:5001/net-monitor@sha256:073b75987e95b89f187a89809f08a32033972bb63cda279db8a9ca16b7ff555a
```

```
$ notation ls $IMAGE
localhost:5001/net-monitor@sha256:073b75987e95b89f187a89809f08a32033972bb63cda279db8a9ca16b7ff555a
└── application/vnd.cncf.notary.v2.signature
    └── sha256:ba3a68a28648ba18c51a479145fca60d96b43dc96c6ab22f412c89ac56a9038b
    └── sha256:22f412c89ac56a90385fca60d96b43dc96c6abbba3a68a28648ba18c51a47914
```

I think we could only print out the signature digest and remove the signed artifact digest from the output as below. The artifact digest is known and immutable when using `notation sign` so we may not need to print out it.

```
$ notation sign $IMAGE
Successfully signed localhost:5001/net-monitor@sha256:073b75987e95b89f187a89809f08a32033972bb63cda279db8a9ca16b7ff555a
Pushed the signature to localhost:5001/net-monitor@sha256:22f412c89ac56a90385fca60d96b43dc96c6abbba3a68a28648ba18c51a47914
```

N/A

### Any additional context?

_No response_

JeyJeyGao:
Added signature manifest reference to sign command output.

Example
```bash
$ notation sign notationreg.azurecr.io/hello-world:s1
Warning: Always sign the artifact using digest(@sha256:...) rather than a tag(:s1) because tags are mutable and a tag reference can point to a different artifact than the one signed.
Successfully signed notationreg.azurecr.io/hello-world@sha256:5790488da2913706b59e5d7c505f3168f5a7d7c20cca1e103d0888147060e89f
Pushed the signature to notationreg.azurecr.io/hello-world@sha256:b8f3ce68525392f19a05fec4670f78445d9b55149181e475eebe491364968122

$ notation ls notationreg.azurecr.io/hello-world:s1  
Warning: Always list the artifact using digest(@sha256:...) rather than a tag(:s1) because resolved digest may not point to the same signed artifact, as tags are mutable.
notationreg.azurecr.io/hello-world@sha256:5790488da2913706b59e5d7c505f3168f5a7d7c20cca1e103d0888147060e89f
    ├── sha256:b8f3ce68525392f19a05fec4670f78445d9b55149181e475eebe491364968122
    └── sha256:50286e9a3c16a161cca77fe2afb1e9db7df54b24030bbc3ddce506138366dbf0
```

Resolves #695 
Resolves #511


cmd/notation/manifest.go
```diff
@@ -38,59 +38,66 @@ func resolveReferenceWithWarning(ctx context.Context, inputType inputType, refer
...
+	// tagOrDigestRef is a tag reference
+	if fn != nil {
+		fn(tagOrDigestRef, manifestDesc)
+	}
+	return manifestDesc, resolvedRef, nil
+}
+
+// parseReference parses the user input reference based on the input type.
+// Returns the repository reference and tag or digest reference.
+func parseReference(reference string, inputType inputType) (repositoryRef, tagOrDigestRef string, err error) {
```
shizhMSFT, line 67: nit: code style: try to avoid named return values. document them in the documentation instead.
```suggestion
func parseReference(reference string, inputType inputType) (string, string, error) {
```

shizhMSFT (approved):
LGTM with suggestions

yizha1:
Just share some information. This is output in text format. We can consider structured output such as JSON in the future, and we can consider more metadata to be outputed.

cmd/notation/manifest.go
JeyJeyGao: Updated.

cmd/notation/sign.go
```diff
@@ -164,7 +162,11 @@ func runSign(command *cobra.Command, cmdOpts *signOpts) error {
 	if err != nil {
 		return err
 	}
-	manifestDesc, resolvedRef, err := resolveReference(ctx, cmdOpts.inputType, cmdOpts.reference, sigRepo, func(ref string, manifestDesc ocispec.Descriptor) {
+	repositoryRef, _, err := parseReference(cmdOpts.reference, cmdOpts.inputType)
```
Two-Hearts, line 165: I don't think we need a refactoring just for an extra print out. The original `resolveReference` does provide all the information needed here.

Two-Hearts (changes requested):

cmd/notation/sign.go
JeyJeyGao: Updated.

cmd/notation/sign.go
```diff
@@ -230,3 +233,13 @@ func prepareSigningOpts(ctx context.Context, opts *signOpts) (notation.SignOptio
 	}
 	return signOpts, nil
 }
+
+// parseRepositoryReference parses the reference with digest and returns the
+// repository reference
+func parseRepositoryReference(reference string) (string, error) {
+	if idx := strings.LastIndex(reference, "@"); idx != -1 {
+		// `digest` found
+		return reference[:idx], nil
```
Two-Hearts, line 242: Consider `strings.Cut(reference, "@")`. 
We could do this because oras-go already does the validation while doing `ParseReference`.

JeyJeyGao: Updated.

Two-Hearts (approved):

Two-Hearts (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #1223, opened 2025-03-14
fix: fix main function

Two-Hearts:
Go won't run defers if `os.Exit()` is called, so in the case of an error the context cancelation will not run.


JeyJeyGao (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #1242, opened 2025-03-26
fix: fix blob command based on spec

### LINKED ISSUE #1171: Blob commands UX revisit
In the process of developing the blob commands, there are comments that the UX can be improved. 

For example: https://github.com/notaryproject/notation/pull/1137#discussion_r1954016455

Hence, creating this issue to track all such discussions. Any related change would need to start with the CLI spec.

Two-Hearts:
This PR fixes blob command based on updated spec.

Resolves #1171


shizhMSFT (approved):
LGTM

Issue #973, opened 2024-06-18
Sign Notation CLI release assets with Notation GitHub Actions


FeynmanZhou:
### Is your feature request related to a problem?

CNCF provides a tool Clomonitor to run OpenSSF Security Best Practice checks on CNCF projects. There is a signing related check that has not been passed in the security check items. See https://clomonitor.io/projects/cncf/notary#notation_security.

According to [OpenSSF Scorecard signing check criteria](https://github.com/ossf/scorecard/blob/main/docs/checks.md#signed-releases), a `.sig` file should be generated in the Notation release assets so the signature could be detected by OpenSSF Scorecard check tool. To meet the OpenSSF security best practice, we need to sign the Notation CLI release assets.

![image](https://github.com/notaryproject/notation/assets/40452856/63822e18-20f6-4a26-9c0e-b89faf9ee451)

### What solution do you propose?

Notation provides [GitHub Actions](https://github.com/notaryproject/notation-action) for signing and verification that can be easily used in GitHub Actions Workflow and release process. 

There are a few dependencies before implementing the signing flow in the release process, such as blob signing and timestamping which are working in progress. Retrieve signing key from GitHub infra as articulated in #905 might another dependency. It's recommended to sign Notation CLI release assets with Notation GitHub Actions after these dependencies are supported in upcoming releases.

### What alternatives have you considered?

N/A

### Any additional context?

N/A

Issue #905, opened 2023-09-20
Support using signing key from GitHub Encrypted Secret


FeynmanZhou:
Signing an image with Notation GitHub Actions relies on a vendor-specific plugin such as [AWS Signer plugin for Notation](https://docs.aws.amazon.com/signer/latest/developerguide/Welcome.html), [Azure Key Vault for Notation](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-tutorial-sign-build-push), [HashiCorp Vault plugin](https://github.com/notaryproject/notation-hashicorp-vault/pulls).. This is a production solution for users to keep the private signing keys secure enough.

However, OSS project maintainers come from different vendors and might not be reluctant to use a vendor-specific plugin in their GHA workflow and release process.

In addition, when users get started with Notation GitHub Actions, they might have some limited access or are unavailable to use a KMS. Instead, they may want to load a private key from the GitHub Encrypted Secret for convenience. 

There was a discussion and related feature request in the [Slack channel](https://cloud-native.slack.com/archives/CQUH8U287/p1689187088295789). 

Signing with a private key loaded from the GitHub secret could be helpful in this scenario. But we need to investigate whether the GitHub encrypted secret is secure enough to store private keys.

To make the Notation GHA signing experience easier, I would suggest supporting signing with a private key loaded from GitHub Encrypted Secret. 

This will make Notation be a good solution to secure open-source project. Open-source project maintainers can easily sign their release binary assets in their release workflow with Notation GitHub Actions.

akashsinghal:
@FeynmanZhou is there any update on prioritizing this. Ratify project is looking to enable signing on image assets published to ghcr. It looks like current support will require us to maintain private key in Azure Key Vault. It's ok for Ratify since we already have existing Azure pipeline support however, I think it's barrier for adoption for smaller projects. Getting GH secret support would be best.

FeynmanZhou:
@akashsinghal I will bring this topic to Notary Project community meeting on Jun 17. Feel free to join if you are convenient.

Issue #539, opened 2023-02-06
Signing with local private keys


ningziwen:
### What is the areas you would like to add the new feature to?

Notation CLI

### Is your feature request related to a problem?

Based on the [discussion](https://github.com/notaryproject/notation/discussions/533), Notation is not supporting signing with local private keys except the testing only keys, and will support it in the future.

I understand current plugin approach is more secure which can manage the private keys for users, but I'd like to see local private key feature as a less advanced feature than current plugin approach, which gives users more flexibility to onboard to existing key management workflow.

### What solution do you propose?

* Not exact a completed solution. I'd like to discuss the external interface first before the detailed design.
* I also like alternative 2. Want to get community's input.

The current workflow is Notation is to manage the key in a separated store, and let `sign` use the key from the store. To be compatible with current workflow, the intuitive way is to add local private key to the key store.
1. Add a special plugin value "none" (or other better name) to state the key is from local.
2. Add a new flag "key-file" to pass the local key file. 
```
notation key add --plugin none --key-file <key_file_path> <key_name>
notation key add --plugin none --key-file key.key my_key
```

### What alternatives have you considered?

1. Bypass the key store workflow. Directly make --key-file part of `sign`

```
notation sign --key-file <key_file_path> <reference>
```
Even we don't choose this alternative, the alternative could be an improvement of proposed solution.

2. Add a very simple build-in plugin for local key file.

The good thing is no special value is needed. The plugin can be called "local" and has only one args --key-file with only one feature of providing local key file.

### Any additional context?

I'm happy to contribute this feature when we reach a concensus.

yizha1:
@ningziwen Thank you for this initiative. Contribution to this project is always welcome! I have some questions about `local private key`. Could you provide more details about your scenario, for example, where does this `local private key` come from? Does notation generate it? How can we be sure the right local key is used for signing? /cc @priteshbandi @shizhMSFT for further discussion.

ningziwen:
The "local private key" is managed by the users with any customized workflow as they have. Notation doesn't generate it. If the users choose to manage the private key by themselves with choosing "local private key", they will have the responsibilty to make sure they are using the right key. Again, this is not the most secure/best solution, but will just make it more flexible by providing an alternative for users to choose.

(If this doesn't align notation's goal or tenets --- for example, if notation only wants to provide the solutions that can help users manage the keys securely, please let me know. That can also help me understand notation better, and we can see if there is anything we may add to the doc to clarify the goals/tenets.)

[...]

Issue #430, opened 2022-11-03
ux: broken default `notation verify` experience


shizhMSFT:
The default experience is broken in `beta.1`.

Expected:
1. Download and install the `notation` binary
2. Run `notation verify` against a public registry
3. `notation` should verify the specified artifact using a default policy with system trusted root CAs.

Actual:
1. Download and install the `notation` binary
2. Run `notation verify` against a public registry fails with the following log.

```console
$ notation verify someregistry.io/somerepo:sometag
Error: open /home/exampleuser/.config/notation/trustpolicy.json: no such file or directory
```

priteshbandi:
The user needs to configure trust-store and trust-policy to start using verify operation.
1. Default policy would differ from usecase to usecase. E.g. during build time, customers might want to have strong validations than deployment time.
2. We took a conscious call to not trust the system's roots

iamsamirzon:
@shizhMSFT  - Was this working in prior version of Notation?

shizhMSFT:
The previous version of `notation` requires a cert but it is relatively easy to do so with a `--cert` option or `notation cert add`. Now, we need to configure the trust policy, which is complex for most new users.

Basically, I'm proposing that we should have an [out of box experience](https://en.wikipedia.org/wiki/Out-of-box_experience) for `notation`.

FeynmanZhou:
Same feeling here. The prerequisites of `notation verify` require configuring the Trust Store and Trust Store since Beta.1. It provides fine-grained authenticity and integrity validation which enhance the security level indeed, but it is too complicated for dummy users to learn and understand the TS/TP concepts. 

As we are developing an open-source project, the users are not limited to security engineers but also include application/platform developers and DevOps engineers, we really need to think about how to [balance security and usability](https://github.com/notaryproject/notaryproject/discussions/159) proposed by @SteveLasker .

Can we abstract two types of users in two different environments to relax the `notation verify` cumbersome user experience and provide different levels of security? 

**Two types of users:**
- Normal users (application/platform developers and DevOps engineers) 
- Professional users (System administrator/Security engineers) 

**Two major environments (scenarios)** 
- Production 
- Development and testing / PoC

  | Testing and Development | Production
-- | -- | --
Normal   users | No need to  configure TS/TP | SHOULD use a default trust         policy with system trusted root CA     Support configure TP/TS         and certificate issue by CA
[...]

PR #1265, opened 2025-04-22
docs: proposal for Signing and Verifying Github Release Assets

JeyJeyGao:


specs/proposals/oss-release-signing.md
```diff
@@ -0,0 +1,89 @@
+## Proposal for Signing and Verifying OSS Release Assets
+
+### Overview
+
+This document proposes enhancing Notation's capabilities for signing and verifying Open Source Software (OSS) community released assets. It outlines the design for a new vendor-neutral `notation-signer` plugin and introduces a simplified, stateless verification mode for the `notation blob verify` command.
+
+### Problem Statement & Motivation
+
+Notation v2.0.0-alpha.1 introduced blob signing, enabling the signing of assets like GitHub releases. However, several issues remain:
```
FeynmanZhou, line 9: Suggest listing the existing GitHub issues for justification. It sets the background for reviewers and lets people understand that this proposal aims to resolve existing gaps.

- Sign Notation CLI release assets with Notation GitHub Actions https://github.com/notaryproject/notation/issues/973
- Support using signing key from GitHub Encrypted Secret https://github.com/notaryproject/notation/issues/905
- Signing with local private keys https://github.com/notaryproject/notation/issues/539
- ux: broken default `notation verify` experience: https://github.com/notaryproject/notation/issues/430

specs/proposals/oss-release-signing.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Existing Solutions or Expectations
+
+Currently, verifying a detached blob signature with Notation requires users to manually download the root certificate, add it to their local trust store, define a trust policy referencing the trust store, and then execute the `notation verify` command. This multi-step process, involving at least 4-5 commands, is cumbersome for one-time verifications. Users often expect a simpler mechanism, perhaps a single command, to check the signature of a downloaded file against a publicly available certificate and identity claim.
+
+### Proposal
+
+This proposal suggests two key enhancements:
+
+1.  **Notation Signer Plugin (`notation-signer`)**: A new plugin implementation allowing users to sign blobs or artifacts using a locally stored key and certificate bundle. The key would ideally be encrypted, with the password manageable via environment variables or secure inputs (e.g., GitHub Actions secrets). The certificate could be stored alongside the signed assets or in the code repository for easy public access. This provides a vendor-neutral signing option.
```
FeynmanZhou, line 48: This proposal doc seems lack of description on the `notation signer plugin` and its signing flow with `notation`. Do you want to elaborate on the signing flow with the new plugin?

specs/proposals/oss-release-signing.md
JeyJeyGao: Nice idea. Updated.

specs/proposals/oss-release-signing.md
JeyJeyGao: Great idea. Updated.

specs/proposals/oss-release-signing.md
```diff
@@ -0,0 +1,113 @@
...
+    3. The publisher adds the content of the encrypted key `notation.key.enc` and the password to the CI workflow secret (GitHub Action Secret), then exports them as environment variables `NOTATION_SIGNER_KEY` and `NOTATION_SIGNER_KEY_PASSWORD` later for the `notation` binary.
+    4. Configure the release workflow to set up `notation` with the `notation-signer` plugin.
+    5. Export the secrets as environment variables for notation and call the `notation blob sign` command to sign the release asset:
+        ```sh
+        # on CI workflow
+        notation blob sign --id signer --plugin signer \
+          --plugin-config certificate_bundle_path=./notation.pem \
+          <release-asset-path>
+        ```
+    6. The publisher attaches the assets, along with their signatures, to the release page.
```
Two-Hearts, line 72: I do have a concern regarding step 1, where user needs to bring their own private key. A leak of this private key could be a severe security issue. I'm thinking if we can give `notation-signer` the capability to generate a key pair, so that the signer can generate their key pair in the CI release pipeline on the fly. This private key is short-lived and won't be stored anywhere in the whole process. 
Take GitHub release as an example:
1. The publisher would like to release their assets through a GitHub workflow. In this workflow, they will need to install Notation CLI and notation-signer plugin. 
2. The next step would be using the notation-signer plugin to generate a key pair on the fly, sign their assets with the private key, throw away the private key once signing is completed, i.e., do not store the private key anywhere since it's no longer needed in future verification. 
3. In the end, the assets, the signature, and the certificate containing the public key is released.

JeyJeyGao: I understand your concern, so we have step 2 to encrypt the key. After step 2, the original key can be removed. The risk of leaking the key will be low as we have password encryption.

Regarding your suggestion, I think the critical issue is ensuring authenticity. Every release has a new certificate, which serves as the trust anchor and proves authenticity. Publishing the signature with the certificate is very risky. As you know, our signature includes the certificate chain, so using the certificate published with the signature is similar to using the certificate in the signature directly. We should put the signature and certificate in different trust sources.

If we prefer a short-lived certificate, we should have a way to ensure authenticity. OIDC is a solution since the signing step needs to authenticate the signer's identity, but that is another story."

Two-Hearts: > I understand your concern, so we have step 2 to encrypt the key. After step 2, the original key can be removed. The risk of leaking the key will be low as we have password encryption.

The nature of letting the user deal with their own private key is not secure. This is the fundamental reason that we use key vaults in other Notation plugins. In addition, it is more convenient to generate the key pair on the fly. User won't need to worry about how to store/remove their private keys. 

> 
> Regarding your suggestion, I think the critical issue is ensuring authenticity. Every release has a new certificate, which serves as the trust anchor and proves authenticity. Publishing the signature with the certificate is very risky. As you know, our signature includes the certificate chain, so using the certificate published with the signature is similar to using the certificate in the signature directly. We should put the signature and certificate in different trust sources.

What do you mean by `Publishing the signature with the certificate is very risky`? The verifier needs the public key to verify the signature, the certificate chain contains the signing certificate and is published along with the signature right (as an unprotected header)?

Two-Hearts (changes requested):
We need further discussion on the proposal.

yizha1:
Let me know when it is ready for review. Thanks.

PR #1270, opened 2025-04-27
fix: fix e2e test urls

Two-Hearts:
The current invalid urls in e2e tests may cause the test to fail occasionally.


JeyJeyGao (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #1335, opened 2025-12-15
docs: add dm-verity image layer signing

dallasd1:
This proposal discusses adding per-layer container image signing using the PKCS#7 format. This will enable signing individual container image layers that are later verified by the kernel at runtime. 

Runtime verification also depends on milestone 1 of this [RFC for code integrity](https://github.com/containerd/containerd/issues/12081) in containerd. At the time of writing, milestone 1.2 is in PR review and milestone 1.3 remains.


NiazFK (approved):
LGTM.

PR #1349, opened 2026-06-10
fix: use %w instead of %v for error wrapping in login and logout

ogulcanaydogan:
Switch `return fmt.Errorf(...)` calls in `cmd/notation/login.go` and `cmd/notation/logout.go` from `%v` to `%w` so callers can inspect the underlying error with `errors.Is`/`errors.As`. The house style across the rest of the codebase already uses `%w` for these patterns (22 non-test files).

**Files changed:**
- `cmd/notation/login.go` (3 sites: credentials store, registry client, login failure)
- `cmd/notation/logout.go` (2 sites: credentials store, logout failure)

### DIFF
--- a/cmd/notation/login.go
+++ b/cmd/notation/login.go
@@ -112,16 +112,16 @@ func runLogin(ctx context.Context, opts *loginOpts) error {
 	cred := opts.Credential()
 	credsStore, err := auth.NewCredentialsStore()
 	if err != nil {
-		return fmt.Errorf("failed to get credentials store: %v", err)
+		return fmt.Errorf("failed to get credentials store: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


kunalworldwide (commented):
Straightforward fix — switching `%v` to `%w` in `fmt.Errorf` wrapping enables callers to use `errors.Is()` and `errors.As()` to inspect the underlying error. This is particularly relevant for the `ErrPlaintextPutDisabled` check on line 123 which already uses `errors.Is` — if any of the earlier wrapping used `%v`, that check would silently break.

All five changes are in login/logout paths where callers would benefit from programmatic error inspection. LGTM.

ogulcanaydogan:
Still valid if anyone has time to take a look. Small mechanical change, build/lint/CodeQL all pass, only the codecov patch-coverage check is flagging the new lines (the error paths weren't covered before either)

PR #1350, opened 2026-06-11
docs: add Long description and example to version command

ogulcanaydogan:
The `version` command had a `Short:` description but no `Long:` or example, unlike most other commands. Add a `Long:` with a usage example to match the pattern used by other notation commands.


ogulcanaydogan:
Bumping this, still a small doc-only change and CI is green. Happy to adjust if the format doesn't match what you'd prefer.

